PR #1, opened 2022-04-20
go bindings for substrait

zeroshade:
Adding go bindings into here as discussed in https://github.com/substrait-io/substrait/pull/168


cpcloud (approved):
Is there a way we can do this entirely with buf and avoid submodules altogether? Can be done as a follow up for sure.

zeroshade:
@cpcloud I used the submodule rather than directly copy the .proto files into this repo so that it ensures we can easily keep the proto files in sync between the two repos. I figured that was easier than using buf and then manually copying the resulting files.

zeroshade:
@cpcloud it looks like it's possible to have buf point at remote .proto files (https://docs.buf.build/tour/list-all-protobuf-files) So i can definitely look into a follow-up to this that can put a buf.gen.yaml here which uses this remote functionality to remove the need for the submodule

cpcloud:
Thanks @zeroshade !

PR #2, opened 2022-05-01
Remove submodule, simplify buf gen

zeroshade:
@cpcloud here we go. Removing the need for the submodule because buf is able to reference the `.proto` files remotely.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// Package substraitgo contains the experimental go bindings for substrait
+// (https://substrait.io).
+package substraitgo
+
+//go:generate buf generate https://github.com/substrait-io/substrait.git#branch=main
--- a/doc.go
+++ b/doc.go
@@ -4,4 +4,4 @@
 // (https://substrait.io).
 package substraitgo
 
-//go:generate buf generate https://github.com/substrait-io/substrait.git#branch=main
+//go:generate buf generate https://github.com/substrait-io/substrait.git#branch=v0.1.2
--- a/doc.go
+++ b/doc.go
@@ -4,4 +4,4 @@
 // (https://substrait.io).
 package substraitgo
 
-//go:generate buf generate https://github.com/substrait-io/substrait.git#branch=master
+//go:generate buf generate https://github.com/substrait-io/substrait.git#branch=main
--- a/doc.go
+++ b/doc.go
@@ -4,4 +4,4 @@
 // (https://substrait.io).
 package substraitgo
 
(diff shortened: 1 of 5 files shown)


zeroshade:
@jacques-n I've updated `doc.go` so that it's now pinned to the version tag `v0.1.2` of the substrait repo. That can be updated and then re-run `go generate` to generate the files using the desired version tag rather than just pinning it to master.

That work?

zeroshade:
@jacques-n also notice that the checked in `.pb.go` files are still checked in, so even having `doc.go` referencing master wouldn't cause any issues building until someone attempts to re-generate the files. The standard practice with go is that the generated protobuf files are checked into the repo so that `go get` works properly with the repository because `go get` will not run any extra code, it only fetches a repository to ensure safety by preventing running arbitrary code when you pull a module. 

That's why originally I had it pointing at master since it only matters what's in `doc.go` when someone runs `go generate` and attempts to check the code in.

cpcloud:
I think the way you've done it makes sense @zeroshade. Ultimately we should publish go bindings in the main substrait repo, and then this repo can use the technique [here](https://docs.buf.build/tour/generate-go-code#install-plugins) to generate code. That can be done later though.

cpcloud (approved):

zeroshade:
@cpcloud your link is precisely the technique I'm using for generating the code if you look at the `go generate` line of doc.go. Should i leave it as pinned to the version v0.1.2 or put it back to pointing a master for easy re-generation without having to also update the doc.go ?

zeroshade:
@cpcloud @jacques-n any thought on leaving the `doc.go` pointing at a specific tag vs pointing at master for easy future codegen before you merge this?

cpcloud:
How about we leave it at `master` for now for easy regen, and then we build some automation in the main substrait repo to submit PRs to update the tag.

@zeroshade Can that line be programmatically updated without sed hacking?

zeroshade:
@jacques-n not really, no. It's basically just a line indicating a command to run. sed is probably the best way to update that line. I'll shift it back to be master

doc.go
```diff
@@ -4,4 +4,4 @@
 // (https://substrait.io).
 package substraitgo
 
-//go:generate buf generate https://github.com/substrait-io/substrait.git#branch=v0.1.2
+//go:generate buf generate https://github.com/substrait-io/substrait.git#branch=master
```
jacques-n, line 7: Should be main, not master.

zeroshade: Bah, I had it right the first time and then confused myself lol

Fixed it now. It correctly says main

PR #4, opened 2022-12-11
feat(expressions): Initial implementation of Expressions

zeroshade:
In the interests of keeping the PRs from getting too large this only sketches out the infrastructure for expressions and implements Literals (and sketches some structs for the other expression types). Subsequent PRs will flesh out and further implement the other Expression types.


JamesRTaylor:
Thanks for putting this together, @zeroshade. Would be great to see some example usage through some tests.

zeroshade:
@JamesRTaylor Updated with a whole mess of stuff plus a testable example and some tests for pretty-printing types and literals to strings. I'll add a few more tests later today and tomorrow, but this isn't implementing *everything* so there's still areas that aren't easily usable yet as there's more to implement, and so it's not easy to show sample usage yet.

JamesRTaylor:
Wanted to connect on something I've been working on to see if we could converge on a solution. I've been working on an analyzer derived from an earlier Apache-licensed version of CockroachDB. They have a language called [OptGen](https://www.cockroachlabs.com/blog/building-cost-based-sql-optimizer/) that supports traversals and rewrites of logical plans through rules (see examples in that link). I'm aiming to open source this work to enable OptGen to be used used over substrait plans. Would be great if we could model expressions similar to [this](https://github.com/cockroachdb/cockroach/blob/v19.1.11/pkg/sql/opt/optgen/lang/expr.og.go). The base expression interface is [here](https://github.com/cockroachdb/cockroach/blob/v19.1.11/pkg/sql/opt/operator.go).

Thoughts?

zeroshade:
I really like the idea of being able to leverage something to perform traversals and rewrites, the only issue I see is that there isn't an easy or obvious way to map those interfaces to the way substrait is designed (that I can tell at least). 

But if we connect and discuss, I think between us we could probably come up with something. My design here was specifically trying to hew close to substrait's design to make conversion between the proto objects and the interfaces low cost, but I think it's worthwhile to see if I can get something closer to what you linked. I'm not going to do much work this week, but I'll definitely take a crack at it early next week. Though if you already have some ideas on how to fit the substrait design closer to those interfaces, I'd happily look at some code!

JamesRTaylor:
Happy New Years! I was out last week as well, but back now. My thinking on the interfaces/impls I pointed you to is that they specify more than substrait needs (mostly around traversing the objects), but that they could be compatible. They have the same set of objects for expressions and relations. For example, the AndExpr implementation:
```
type AndExpr struct {
	Left  Expr
	Right Expr
	Src   *SourceLoc
	Typ   DataType
}

func (e *AndExpr) Op() Operator {
	return AndOp
}

func (e *AndExpr) ChildCount() int {
	return 2
}

func (e *AndExpr) Child(nth int) Expr {
	switch nth {
	case 0:
		return e.Left
	case 1:
		return e.Right
	}
	panic(fmt.Sprintf("child index %d is out of range", nth))
}

func (e *AndExpr) ChildName(nth int) string {
	switch nth {
	case 0:
		return "Left"
	case 1:
		return "Right"
	}
	return ""
}

func (e *AndExpr) Value() interface{} {
	return nil
}

func (e *AndExpr) Visit(visit VisitFunc) Expr {
	children := visitChildren(e, visit)
	if children != nil {
		return &AndExpr{Left: children[0], Right: children[1], Src: e.Source()}
	}
	return e
}

func (e *AndExpr) Source() *SourceLoc {
	return e.Src
}

func (e *AndExpr) InferredType() DataType {
	return e.Typ
}

func (e *AndExpr) String() string {
	var buf bytes.Buffer
	e.Format(&buf, 0)
	return buf.String()
}

func (e *AndExpr) Format(buf *bytes.Buffer, level int) {
	formatExpr(e, buf, level)
}
```
Let me try to pull this out of our codebase with a minimal set of dependencies. One useful interface in these objects (not part of substrait) is a way to print them out (which is obviously useful :-) ). I'm just finishing up something else, but should be able to have something soon.

zeroshade:
@JamesRTaylor could you take a look at https://github.com/apache/arrow/blob/master/go/arrow/compute/expression.go#L48 and tell me what you think of the expression and field reference interfaces I wrote for the Arrow module (before I started trying to adapt substrait). Personally I like the simplicity of that interface in that there's only three implementations of the interface: an expression is either a literal, a parameter reference or a function call with a slice of expressions as arguments.

But there's definitely a benefit to fitting the interface that OptGen uses for inspiration interoperability. While I wait to see what you come up with, I'll take a stab myself and putting something together and let's see if we can catch up and find something that works.

types.go
```diff
@@ -0,0 +1,921 @@
...
+		case t.Nullability != other.Nullability:
+			return false
+		case t.TypeVariationRef != other.TypeVariationRef:
+			return false
+		case t.TypeReference != other.TypeReference:
+			return false
+		case len(t.TypeParameters) != len(other.TypeParameters):
+			return false
+		}
+
```
JamesRTaylor, line 819: Do we need a check here comparing len(t.TypeParameters) and len(other.TypeParameters)?

expression.go
```diff
@@ -0,0 +1,195 @@
...
+			Options:    et.WindowFunction.Options,
+			OutputType: TypeFromProto(et.WindowFunction.OutputType),
+			Phase:      et.WindowFunction.Phase,
+			Invocation: et.WindowFunction.Invocation,
+			Partitions: parts,
+			Sorts:      sorts,
+			LowerBound: BoundFromProto(et.WindowFunction.LowerBound),
+			UpperBound: BoundFromProto(et.WindowFunction.UpperBound),
+		}, nil
+	case *proto.Expression_IfThen_:
```
JamesRTaylor, line 91: Any reason not to implement these?

expression.go
```diff
@@ -0,0 +1,195 @@
...
+//  - A Cast expression
+//  - A Subquery
+//  - A Nested expression
+type Expression interface {
+	FuncArg
+	RootRefType
+
+	GetType() Type
+	ToProto() *proto.Expression
+	Equals(Expression) bool
```
JamesRTaylor, line 125: I think we'll need a general way of walking through children, no? Like `Walk(Visitor) Expression`

expressions_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"fmt"
+
+	substraitgo "github.com/substrait-io/substrait-go"
+	"github.com/substrait-io/substrait-go/proto"
+	"google.golang.org/protobuf/encoding/protojson"
+	pb "google.golang.org/protobuf/proto"
+)
+
```
JamesRTaylor, line 13: I think we need roundtrip tests (like your type tests) for various nested expressions.

types.go
zeroshade: we check it on line 816, if they aren't equal, we'll have returned false already.

expression.go
zeroshade: I was just trying to keep the size of this PR down by not implementing these yet, and instead implementing them in the next PR. If you'd prefer I can implement them now and add them to this PR.

expressions_test.go
zeroshade: sure, i can add some more roundtrip tests for some nested expressions. I'll do that this week.

expression.go
zeroshade: I'll follow the pattern from optgen that you linked earlier where the visiting could be used to generate a replacement tree if desired etc.

zeroshade: added

expressions_test.go
zeroshade: added a roundtrip test, and a comment that I'll add more roundtrip tests after parsing is implemented since that'll make it easier to generate the nested expressions to test with.

JamesRTaylor:
This LGTM. Any feedback, @jacques-n before I merge?

zeroshade:
@JamesRTaylor Added a note for the README on this. Let me know if there's anything you'd like me to change or include in it.

JamesRTaylor (approved):
LGTM

PR #8, opened 2023-02-01
ci(release): update substrait version and setup release workflow

zeroshade:
As per the conversation at the bi-weekly meetup we'll pin the protobuf to a specific generated version which will show up in the package doc and also add a release workflow to automate releases based on the conventional semantic commits.


JamesRTaylor (approved):
LGTM. Thanks for setting this up.

PR #9, opened 2023-02-15
feat(expressions): Implement more expression types

zeroshade:
This implements handling for all expression types *except* for Subquery, and sets up the framework for plan handling.

Including a unit test framework for defining expressions in a test yaml file which get used as test cases.

### DIFF
--- a/expr/expression.go
+++ b/expr/expression.go
@@ -53,12 +53,18 @@ func ExprFromProto(e *proto.Expression, baseSchema types.Type, ext ExtensionLook
 			}
 		}
 
-		id, ok := ext.DecodeFunc(et.ScalarFunction.FunctionReference)
-		if !ok {
-			return nil, substraitgo.ErrNotFound
-		}
+		var (
+			id   extensions.ID
+			decl *extensions.ScalarFunctionVariant
+		)
+		if ext != nil {
+			var ok bool
+			if id, ok = ext.DecodeFunc(et.ScalarFunction.FunctionReference); !ok {
+				return nil, substraitgo.ErrNotFound
+			}
 
-		decl, _ := ext.LookupScalarFunction(et.ScalarFunction.FunctionReference, c)
+			decl, _ = ext.LookupScalarFunction(et.ScalarFunction.FunctionReference, c)
+		}
 		return &ScalarFunction{
 			FuncRef:     et.ScalarFunction.FunctionReference,
 			Declaration: decl,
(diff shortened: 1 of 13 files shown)


expr/expression.go
```diff
@@ -33,6 +35,10 @@ func FuncArgFromProto(e *proto.FunctionArgument, baseSchema types.Type, ext Exte
 }
 
 func ExprFromProto(e *proto.Expression, baseSchema types.Type, ext ExtensionLookup, c *extensions.Collection) (Expression, error) {
+	if e == nil {
+		return nil, nil
+	}
```
JamesRTaylor, line 40: Does this really happen? Not sure as a caller what I'd do if I got back (nil, nil).

expr/expression.go
```diff
@@ -47,12 +53,18 @@ func ExprFromProto(e *proto.Expression, baseSchema types.Type, ext ExtensionLook
...
+		)
+		if ext != nil {
+			var ok bool
+			if id, ok = ext.DecodeFunc(et.ScalarFunction.FunctionReference); !ok {
+				return nil, substraitgo.ErrNotFound
+			}
 
-		decl, _ := ext.LookupScalarFunction(et.ScalarFunction.FunctionReference, c)
+			decl, _ = ext.LookupScalarFunction(et.ScalarFunction.FunctionReference, c)
+		}
```
JamesRTaylor, line 67: Can't this return an error if the scalar function isn't found? Should that be checked here and then returned on error?

expr/expression.go
```diff
@@ -84,12 +96,18 @@ func ExprFromProto(e *proto.Expression, baseSchema types.Type, ext ExtensionLook
...
+		)
+		if ext != nil {
+			var ok bool
+			if id, ok = ext.DecodeFunc(et.WindowFunction.FunctionReference); !ok {
+				return nil, substraitgo.ErrNotFound
+			}
 
-		decl, _ := ext.LookupWindowFunction(et.WindowFunction.FunctionReference, c)
+			decl, _ = ext.LookupWindowFunction(et.WindowFunction.FunctionReference, c)
+		}
```
JamesRTaylor, line 110: Same question here

JamesRTaylor:
Couple of minor comments, but it's great to see this get rounded out!

expr/expression.go
zeroshade: So I put this for the off-chance where a user passes a `nil` expression or something in the nested protobuf has a `nil` value so that we don't panic with a `dereferenced nil` exception etc. In theory this should only ever happen from either a coding error or a malformed protobuf (I think), but I'd rather avoid the panic if possible by accidentally dereferencing a nil pointer.

JamesRTaylor: IMHO, I think raising our own panic here would be preferable, otherwise the caller has to be aware of this case of getting (nil, nil) back (and it's not clear what the right thing to do would be). We could throw our own specific error if we think callers would want to potentially recover from and ignore this situation.

expr/expression.go
JamesRTaylor: What happens down-the-road if decl is nil for ScalarFunction? In Java-land, we load the default extension collections statically up front, so they'd always be there - so in this case we'd error if we don't find it.

I think, in general, it's better to error in these cases early. Otherwise, strange things will happen down-the-road and it'll be difficult to trace it back to this.

expr/expression.go
zeroshade: Fair point, i'll update this shortly

expr/expression.go
zeroshade: Make a lot of sense. I'll have it skip this check if the collection passed in is nil, but if the collection is non-nil it'll error if it can't find it.

I do plan on having the default extension collections eventually statically loaded, I just haven't implemented that yet.

zeroshade: updated to check  if collection is nil and if it isn't, we error if we can't find the function

expr/expression.go
zeroshade: updated

zeroshade:
@JamesRTaylor Updated for your feedback. Let me know if there's anything else needed here! :smile:

JamesRTaylor (approved):
LGTM

PR #12, opened 2023-03-13
ci(release): fix release yaml docker syntax to equals sign

zeroshade:


JamesRTaylor (approved):
Thanks for following up on this

PR #13, opened 2023-03-14
feat(extensions): add embedded default extension collection

zeroshade:
Along with some small fixes I encountered while testing expression evaluation.

### DIFF
--- a/extensions/extension_mgr.go
+++ b/extensions/extension_mgr.go
@@ -3,15 +3,47 @@
 package extensions
 
 import (
+	"embed"
 	"fmt"
 	"io"
+	"path"
 
+	"github.com/goccy/go-yaml"
 	substraitgo "github.com/substrait-io/substrait-go"
 	"github.com/substrait-io/substrait-go/proto/extensions"
-	"gopkg.in/yaml.v3"
 )
 
 type AdvancedExtension = extensions.AdvancedExtension
+
+const substraitDefaultURIPrefix = "https://github.com/substrait-io/substrait/blob/main/extensions/"
+
+// DefaultCollection is loaded with the default Substrait extension
+// definitions with the exception of decimal arithemtic. Decimal arithmetic
+// functions are missing as the complex return type expressions are not
+// yet implemented.
+var DefaultCollection Collection
+
+//go:embed vendor/*
+var vendored embed.FS
+
+func init() {
+	entries, err := vendored.ReadDir("vendor")
+	if err != nil {
+		return
+	}
+
+	for _, ent := range entries {
+		f, err := vendored.Open(path.Join("vendor/", ent.Name()))
+		if err != nil {
+			panic(err)
+		}
+		defer f.Close()
+		err = DefaultCollection.Load(substraitDefaultURIPrefix+ent.Name(), f)
+		if err != nil {
+			panic(err)
+		}
+	}
+}
 
 type ID struct {
 	URI, Name string
--- a/extensions/vendor/functions_arithmetic.yaml
+++ b/extensions/vendor/functions_arithmetic.yaml
@@ -1,0 +1,1497 @@
+%YAML 1.2
+---
+scalar_functions:
+  -
+    name: "add"
+    description: "Add two values."
+    impls:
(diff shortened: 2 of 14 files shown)


extensions/extension_mgr.go
```diff
@@ -3,16 +3,48 @@
...
 )
 
 type AdvancedExtension = extensions.AdvancedExtension
 
+const substraitDefaultURIPrefix = "https://github.com/substrait-io/substrait/blob/main/extensions/"
+
+// DefaultCollection is loaded with the default Substrait extension
+// definitions with the exception of decimal arithemtic. Decimal arithmetic
+// functions are missing as the complex return type expressions are not
+// yet implemented.
```
JamesRTaylor, line 23: Not sure if it'd help you, but CockroachDB has a Decimal implementation [here](https://github.com/cockroachdb/apd).

extensions/vendor/functions_arithmetic.yaml
```diff
@@ -0,0 +1,1497 @@
...
+            value: i32
+        return: i32
+      - args:
+          - name: x
+            value: i64
+          - name: y
+            value: i64
+        return: i64
+
+aggregate_functions:
```
JamesRTaylor, line 852: Are `count` and `approx_count_distinct` implemented already?

extensions/extension_mgr.go
zeroshade: Arrow has a full decimal implementation for decimal128 and decimal256 (that I wrote portions of haha) the issue isn't the decimal implementation but the [return type expression](https://substrait.io/expressions/scalar_functions/#return-type-expressions) parsing that I haven't implemented yet here. That file is the only one which uses the return type expressions, used to dynamically determine the precision and scale for the returned type.

extensions/vendor/functions_arithmetic.yaml
zeroshade: All I did was copy the yaml files from `substrait-io/substrait` here because `go get` doesn't resolve submodules. 

So if those are expected to exist but aren't here, that would mean they are missing from the yaml files in the substrait main repo.

JamesRTaylor: Maybe some files didn't get copied? Like these:
https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_approx.yaml
https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml

zeroshade: ah must have missed those! thanks. i'll add those and update

zeroshade: just pushed the files

zeroshade:
@JamesRTaylor the existing tests will by virtue of it being in the `init` function, test that there's no errors when reading in the files. 

While the current tests don't explicitly check the default collection to ensure that everything is there (because I didn't want to tie the tests the content in the vendored yaml files) the *functionality* of loading the yaml and looking up methods is tested.

If you think it's worth it, I can add some tests to check the default collection for the existence of a few specific functions.

JamesRTaylor:
> If you think it's worth it, I can add some tests to check the default collection for the existence of a few specific functions.

Yep, I think that's a good idea just to make sure the default collections are consumable like we expect.

expr/field_reference.go
```diff
@@ -401,6 +401,34 @@ func (f *FieldReference) Visit(VisitFunc) Expression {
...
+			rootType = rootExpr.GetType()
+		}
+
+		typ, err := rt.GetType(rootType)
+		if err != nil {
+			return err
+		}
+		f.knownType = typ
+		return nil
+	case *MaskExpression:
```
JamesRTaylor, line 426: Not a big deal, but any reason for this case *MaskExpression

zeroshade: it'll end up falling through to the `NotImplemented` error for now. I just wanted to make it obvious that the `MaskExpression` case wasn't implemented yet here so it doesn't get lost or forgotten about :smile:

PR #14, opened 2023-03-16
feat(builders): add helper functions for creating and binding expressions

zeroshade:


expr/expression.go
```diff
@@ -318,23 +323,34 @@ type VisitFunc func(Expression) Expression
...
+	// IsBound means that this expression has a defined output
+	// type and, in the case of Scalar/Window functions, has
+	// properly set a Function Anchor and a Declaration defining
+	// the function variant being called as registered in an
+	// extension collection.
+	//
+	// To bind an expression, use BindExpression which will
+	// recursively resolve the output types for all references and
+	// functions that are not yet bound in it.
+	IsBound() bool
```
JamesRTaylor, line 352: We don't expose this idea of "partially bound" references in substrait-java. If we can't find a function declaration, then we'd just error out. Is there a reason we'd want to have the plan be in this state temporarily? I think it'll make it harder to interact with the plan if the caller has to know to check this.

zeroshade: This decision came out of the discussion at the last bi-weekly meeting when @westonpace made the observation that we don't want to error out if we can't find the function declaration because of potential usage of custom URNs that aren't actually a URL but specify other types of in-built plugins or operations. He explicitly requested that I not make it error out if it can't find the function declaration. (plz chime in here weston if I didn't explain that correctly / well enough).

In addition, field references that are root references (as opposed to an expression reference) need to know their base schema in order to know what their output type is. By having this concept of a "partially bound" expression, the same expression could get utilized for different input shapes by simply binding it to different base schemas which could then change which variant of a function declaration gets used or bound.

It makes the objects more reusable when programmatically constructing expressions from the ground up while preserving the ability to customize extension usage.

JamesRTaylor: Is being unbound an intermediate state? In substrait-java we have builders that captures any intermediate state, but once built, the plan is known to be valid. Can a full plan be read and then contain unbound expressions?

zeroshade: > To create a Substrait plan the only piece that is needed from the YAML in this scenario is the return type correct? Would it be possible to require that either the function exists in YAML somewhere or the user provides the return type themselves?

I guess the question becomes whether or not it's something desireable to be able to re-use an expression with a different input base schema (to allow re-resolving the extension function to a potentially different variant if the arguments change types due to field references). I could likely refactor the builders here to enforce *either* having the output type provided *or* successfully getting the declaration from the extension lookup / collection. But that still leaves the case of field references. If we don't want to allow "partially bound" expressions, it would require having to know the input base schema at the time any field reference (that uses root) is constructed so that we can know what the output type of the field reference is at the moment of construction. Which seems a bit constrictive but is definitely possible.

In this way we'd probably want to also refactor the objects to not expose the members publicly but instead convert everything to Getters so that modifying or manipulating an expression would require creating a new object in a setter where we can validate it. yes?

JamesRTaylor: > In this way we'd probably want to also refactor the objects to not expose the members publicly but instead convert everything to Getters so that modifying or manipulating an expression would require creating a new object in a setter where we can validate it. yes?

I think this is a good idea. It'll keep the read plan API lean and mean. Otherwise the producer API will probably evolve to eclipse the read API.

zeroshade: I'll update this with the refactor in a bit. Hopefully the result isn't too big for a review.... haha

zeroshade:
@JamesRTaylor Here's the changes I made:

* Instead of passing an `extensions.Set` and `extensions.Collection` everywhere, I combined them into a single `ExtensionRegistry` that can be used for expression building and extracting from Proto. You can either create an empty registry by providing a `Collection` or utilize the set that was extracted by reading in a Plan/ExtendedExpression (both situations have examples in the tests)
* Removed the `BindExpression` function and `IsBound` methods
* Made large portions of the members of the objects no longer exported and only accessible via getter methods now
* Added `NewScalarFunc`/`NewCustomScalarFunc` style methods for all three function types. the `Custom` variant is for the use case that @westonpace mentioned, allowing someone to provide the variant + an output type for customization without resulting in an invalid expression.
* Added other `New....` functions for constructing expression types + error checking them based on the substrait docs
* a few other fixes here and there to conform to the updated APIs

Hope this isn't *too* big now lol

extensions/variants.go
```diff
@@ -82,6 +240,74 @@ type WindowFunctionVariant struct {
...
+	Decomposable DecomposeType
+	// should be a type expression
+	// must not be empty if decomposable is not DecomposeNone
+	IntermediateOutputType string
+	WindowType             WindowType
+}
+
+func NewWindowFuncVariantOpts(id ID, opts WindowVariantOpts) *WindowFunctionVariant {
+	var aggIntermediate parser.TypeExpression
+	if opts.Decomposable == "" {
```
JamesRTaylor, line 272: I don't have a strong preference, but just curious on plus/minus of using const strings versus enum with iota. It'd prevent this kind of translation, but then you'd need a String method. For example:
```
type DecomposeType int

const (
	DecomposeNone DecomposeType = iota
	DecomposeOne
	DecomposeMany
)
```
What are your thoughts?

JamesRTaylor (approved):
This looks great. Thanks for all the refactoring. Just had one question.

extensions/variants.go
zeroshade: So I had originally done this using the const strings so that it could easily and freely get the conversion from YAML. but i think you're right that it would be better to use `iota` and add a `String` method + `UnmarshalYAML` etc. methods

PR #15, opened 2023-03-23
fix(deps): rename go embed dir

zeroshade:
Turns out that directories named `vendor` are excluded from modules and thus are excluded from `go embed` being able to auto include them with the module. So renaming the directory in order to ensure that the extension definitions get pulled with the module when using `go get`


JamesRTaylor (approved):
Thanks for the quick turnaround!

PR #16, opened 2023-04-03
ci(release): use docker run -u root to fix release workflow

zeroshade:
`go-semantic-release` tries to write a `.version` file and then fails with a permission denied error, probably because the `docker run` runs as a different user. So let's use `-u root` so the release workflow can succeed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
         # if event_name == pull_request it will not perform a release          
         run: |
           echo $USER    
-          docker run -u $USER -e GITHUB_ACTION=${{ github.action}} \
+          docker run -u 1000 -e GITHUB_ACTION=${{ github.action}} \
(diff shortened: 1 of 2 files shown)


jacques-n:
FYI, I'm not trying to block this commit, just felt icky/suboptimal.

zeroshade:
I agree that it feels icky, I'll poke around using job.container and see if I can make that work

zeroshade:
@mbrobbel Okay so i can't use `[jobs.<job_id>.container](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container)` because there's no node install in the container, it's basically just the binary itself that i'm trying to run on its own. So the checkout step can't run in the container. 

But I think I got this to work as the dockerfile for the binary sets USER=1000 and I think the GHA docker settings had it running as a different user? I don't now, but I think the current iteration here will work, hopefully.

PR #18, opened 2023-04-17
feat: re-introduce substrait submodule

vbarua:
# Problem 
When creating custom extension messages, we ran into a limitation where we could not reference the existing substrait message types easily as they were not included in the substrait-go library. The workaround for this, vendoring the messages outside of the library, is worse that just including them in the substrait-go. This reduces the number of things that need to be kept in sync.

# PR
This PR
* Adds the substrait submodule back into substrait-go
* Adds a utility script `./update_substrait` to make it easy to update the protos.


update_substrait
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+git submodule update --remote substrait
+buf generate substrait
```
EpsilonPrime, line 3: Not sure if this repo is fanatic about having linefeeds at the end of the file.

EpsilonPrime (approved):
LGTM.  As mentioned in Slack, we will likely want a CI to enforce that the protos are updated when the files are.  The CI will probably also need to make sure that buf (and any related dependencies) is available.

vibhatha:
Just a thought: Also would it be better to keep the `submodule` inside a directory `third_party`?

vbarua:
Both [substrait-java](https://github.com/substrait-io/substrait-java) and [substrait-rs](https://github.com/substrait-io/substrait-rs) have the submodule in `substrait`, and it is the only submodule.

[substrait-cpp](https://github.com/substrait-io/substrait-cpp) has it in `third_party/substrait`, but it also has other dependencies as submodules.

I think with substrait-go, `substrait` should be the only submodule needed, so I think having it at the top-level makes sense.

vbarua:
There are alternative mechanisms for accessing the protos.

PR #20, opened 2023-05-11
feat(expr/builders): Add initial expression builder helpers

zeroshade:
only implements builders for ScalarFunction/WindowFunction/Cast/AggregateFunction/Literals


zeroshade:
Will merge at EOD if there's no objections

PR #21, opened 2023-06-02
feat(plan): Plan builders

zeroshade:
Still need to add some documentation and tests, but this is a good first pass at creating plan builders.


zeroshade:
@westonpace @JamesRTaylor when either of you have a chance, I'd love a quick review here just for some validation to ensure that this approach for plan builders makes sense.

Thanks much!

zeroshade:
If there's no comments by EOD today I'll merge this.

PR #22, opened 2023-08-03
fix(builders): Updated Remap functions to test correct output mapping

wackywendell:
Hello! I've been having fun trying out the builders; they're quite useful.

In the process, I noticed that the `ProjectRemap` and `AggregateRemap` builder functions were incorrectly comparing indices with the input length, rather than the output length, which for those two is not the same - before remapping, `Project` operators output (inputColumns…, expressions…), and `Aggregate` outputs `(measures…, groupings…)`. So here is a PR to fix the remapping index checker.

I looked at `CrossRemap` and `NamedScanRemap`, and they look fine; meanwhile, `FilterRemap`, `FetchRemap`, `SortRemap`, and `SetRemap` should be fine as their outputs are the same types as their inputs.


zeroshade (approved):
Thanks for catching and fixing this!

PR #29, opened 2024-07-18
Introduce rewrite tools on Rel.

scgkiran:
Add GetInputs, Copy, CopyWithExpressionRewrite to Rel

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -5,11 +5,12 @@ package substraitgo
 import "errors"
 
 var (
-	ErrNotImplemented = errors.New("not implemented")
-	ErrInvalidType    = errors.New("invalid type")
-	ErrInvalidExpr    = errors.New("invalid expression")
-	ErrNotFound       = errors.New("not found")
-	ErrKeyExists      = errors.New("key already exists")
-	ErrInvalidRel     = errors.New("invalid relation")
-	ErrInvalidArg     = errors.New("invalid argument")
+	ErrNotImplemented    = errors.New("not implemented")
+	ErrInvalidType       = errors.New("invalid type")
+	ErrInvalidExpr       = errors.New("invalid expression")
+	ErrNotFound          = errors.New("not found")
+	ErrKeyExists         = errors.New("key already exists")
+	ErrInvalidRel        = errors.New("invalid relation")
+	ErrInvalidArg        = errors.New("invalid argument")
+	ErrInvalidInputCount = errors.New("invalid input count")
 )
--- a/plan/plan.go
+++ b/plan/plan.go
@@ -278,6 +278,16 @@ type Rel interface {
 	GetAdvancedExtension() *extensions.AdvancedExtension
 	ToProto() *proto.Rel
 	ToProtoPlanRel() *proto.PlanRel
+
+	// Copy creates a copy of this relation with new inputs
+	Copy(newInputs ...Rel) (Rel, error)
+
+	// GetInputs returns a list of zero or more inputs for this relation
+	GetInputs() []Rel
+
+	// CopyWithExpressionRewrite rewrites all expression trees in this Rel. Returns nil
+	// if no changes were made, otherwise a newly created rel that includes the given expressions
+	CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), inputs ...Rel) (Rel, error)
 }
 
 func RelFromProto(rel *proto.Rel, reg expr.ExtensionRegistry) (Rel, error) {
--- a/plan/plan.go
+++ b/plan/plan.go
@@ -244,6 +244,8 @@ func (r *Root) RecordType() types.NamedStruct {
 	}
 }
 
+type RewriteFunc func(expr.Expression) (expr.Expression, error)
+
 // Rel is a relation tree, representing one of the expected Relation
 // types such as Fetch, Sort, Filter, Join, etc.
 //
@@ -285,9 +287,9 @@ type Rel interface {
 	// GetInputs returns a list of zero or more inputs for this relation
 	GetInputs() []Rel
 
-	// CopyWithExpressionRewrite rewrites all expression trees in this Rel. Returns nil
+	// CopyWithExpressionRewrite rewrites all expression trees in this Rel. Returns original Rel
 	// if no changes were made, otherwise a newly created rel that includes the given expressions
-	CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), inputs ...Rel) (Rel, error)
+	CopyWithExpressionRewrite(rewriteFunc RewriteFunc, newInputs ...Rel) (Rel, error)
 }
 
 func RelFromProto(rel *proto.Rel, reg expr.ExtensionRegistry) (Rel, error) {
--- a/plan/relations.go
+++ b/plan/relations.go
@@ -122,7 +122,7 @@ func (b *baseReadRel) GetInputs() []Rel {
 	return []Rel{}
 }
 
-func (b *baseReadRel) copyExpressions(rewriteFunc func(expr.Expression) (expr.Expression, error)) ([]expr.Expression, error) {
+func (b *baseReadRel) copyExpressions(rewriteFunc RewriteFunc) ([]expr.Expression, error) {
 	filter, err := rewriteFunc(b.filter)
 	if err != nil {
 		return nil, err
@@ -131,10 +131,15 @@ func (b *baseReadRel) copyExpressions(rewriteFunc func(expr.Expression) (expr.Ex
 	if err != nil {
 		return nil, err
 	}
-	if filter == b.filter && bestEffortFilter == b.bestEffortFilter {
-		return nil, nil
-	}
 	return []expr.Expression{filter, bestEffortFilter}, nil
+}
+
+func (b *baseReadRel) getExpressions() []expr.Expression {
+	return []expr.Expression{b.filter, b.bestEffortFilter}
+}
+
+func (b *baseReadRel) updateFilters(filters []expr.Expression) {
+	b.filter, b.bestEffortFilter = filters[0], filters[1]
 }
 
 // NamedTableReadRel is a named scan of a base table. The list of strings
@@ -176,21 +181,20 @@ func (n *NamedTableReadRel) ToProto() *proto.Rel {
 	}
 }
 
-func (n *NamedTableReadRel) Copy(newInputs ...Rel) (Rel, error) {
+func (n *NamedTableReadRel) Copy(_ ...Rel) (Rel, error) {
 	return n, nil
 }
 
-func (n *NamedTableReadRel) CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), newInputs ...Rel) (Rel, error) {
-	newExprs, err := n.baseReadRel.copyExpressions(rewriteFunc)
+func (n *NamedTableReadRel) CopyWithExpressionRewrite(rewriteFunc RewriteFunc, _ ...Rel) (Rel, error) {
+	newExprs, err := n.copyExpressions(rewriteFunc)
 	if err != nil {
 		return nil, err
 	}
-	if newExprs == nil {
-		return nil, nil
+	if slices.Equal(newExprs, n.getExpressions()) {
+		return n, nil
 	}
 	nt := *n
-	nt.filter = newExprs[0]
-	nt.bestEffortFilter = newExprs[1]
+	nt.updateFilters(newExprs)
 	return &nt, nil
 }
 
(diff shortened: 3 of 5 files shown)


vbarua:
This is very promising! We have something similar to this in substrait-java in the [RelCopyOnWriteVisitor](https://github.com/substrait-io/substrait-java/blob/main/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java) and [ExpressionCopyOnWriteVisitor](https://github.com/substrait-io/substrait-java/blob/main/core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java) which allows us to implement arbitrary tree rewrites quite easily.

errors.go
```diff
@@ -5,11 +5,12 @@ package substraitgo
 import "errors"
 
 var (
-	ErrNotImplemented = errors.New("not implemented")
```
jacques-n, line 8: A bunch of extraneous reformatting here.

plan/plan.go
```diff
@@ -278,6 +278,16 @@ type Rel interface {
...
+
+	// Copy creates a copy of this relation with new inputs
+	Copy(newInputs ...Rel) (Rel, error)
+
+	// GetInputs returns a list of zero or more inputs for this relation
+	GetInputs() []Rel
+
+	// CopyWithExpressionRewrite rewrites all expression trees in this Rel. Returns nil
+	// if no changes were made, otherwise a newly created rel that includes the given expressions
+	CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), inputs ...Rel) (Rel, error)
```
jacques-n, line 290: As I've thought about this more, I think we should change. Let's just return original rel if no changes were needed. We can update the description to make this clear.

plan/relations.go
```diff
@@ -156,6 +176,24 @@ func (n *NamedTableReadRel) ToProto() *proto.Rel {
...
+	return n, nil
+}
+
+func (n *NamedTableReadRel) CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), newInputs ...Rel) (Rel, error) {
+	newExprs, err := n.baseReadRel.copyExpressions(rewriteFunc)
+	if err != nil {
+		return nil, err
+	}
+	if newExprs == nil {
+		return nil, nil
```
jacques-n, line 189: per comment above, let's avoid returning nil in the case that expressions are unchanged and return original.

plan/relations.go
```diff
@@ -1136,6 +1570,62 @@ func (mr *MergeJoinRel) ToProtoPlanRel() *proto.PlanRel {
...
+	}
+	merge := *mr
+	merge.left = newInputs[0]
+	merge.right = newInputs[1]
+	merge.postJoinFilter = newExpr
+	return &merge, nil
+}
+
+func areExpressionsEqual(exprs []expr.Expression, exprs2 []expr.Expression) bool {
```
jacques-n, line 1605: Equal might not be the best name here. Maybe adjust to something like Same or Identical with a doc describing behavior. same for are rels equal

plan/relations.go
```diff
@@ -1136,6 +1570,62 @@ func (mr *MergeJoinRel) ToProtoPlanRel() *proto.PlanRel {
...
+	}
+	for i, e := range exprs {
+		if e != exprs2[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func areRelsEqual(slice1 []Rel, slice2 []Rel) bool {
```
jacques-n, line 1617: as above

jacques-n (commented):
Mostly looking good. Few minor proposed changes.

jacques-n:
> ...something similar to this in substrait-java...

Lol, why invent when you can just copy useful patterns, right? :D

plan/relations.go
```diff
@@ -117,6 +118,25 @@ func (b *baseReadRel) toReadRelProto() *proto.ReadRel {
...
+		return nil, err
+	}
+	bestEffortFilter, err := rewriteFunc(b.bestEffortFilter)
+	if err != nil {
+		return nil, err
+	}
+	if filter == b.filter && bestEffortFilter == b.bestEffortFilter {
+		return nil, nil
+	}
+	return []expr.Expression{filter, bestEffortFilter}, nil
```
zeroshade, line 137: could we shift this check up and just return `[]expr.Expression{filter, bestEffortFilter}` regardless (same as with the `CopyWithExpressionRewrite` method, just return the original ones if no changes are needed

plan/relations.go
```diff
@@ -156,6 +176,24 @@ func (n *NamedTableReadRel) ToProto() *proto.Rel {
...
+	newExprs, err := n.baseReadRel.copyExpressions(rewriteFunc)
+	if err != nil {
+		return nil, err
+	}
+	if newExprs == nil {
+		return nil, nil
+	}
+	nt := *n
+	nt.filter = newExprs[0]
+	nt.bestEffortFilter = newExprs[1]
```
zeroshade, line 193:
```suggestion
	nt.filter, nt.bestEffortFilter = newExprs[0], newExprs[1]
```

plan/relations.go
```diff
@@ -194,6 +232,24 @@ func (v *VirtualTableReadRel) ToProtoPlanRel() *proto.PlanRel {
...
+	if err != nil {
+		return nil, err
+	}
+	if newExprs == nil {
+		return nil, nil
+	}
+	nt := *v
+	nt.filter = newExprs[0]
+	nt.bestEffortFilter = newExprs[1]
+	return &nt, nil
```
zeroshade, line 250: Rather than repeating this over and over, it feels like we could easily write a simple generic helper function, something like:

```go
func (b *baseReadRel) updateFilters(filter, bestEffortFilter expr.Expression) {
    b.filter, b.bestEffortFilter = filter, bestEffortFilter
}

type exprRewriter interface {
    copyExpressions(func(expr.Expression) (expr.Expression, error)) ([]expr.Expression, error)
    updateFilters(filter, bestEffortFilter expr.Expression)
}

func copyRewriteHelper[T exprRewriter](r *T, rewriteFunc func(expr.Expression) (expr.Expression, error)) (*T, error) {
    newExprs, err := r.copyExpressions(rewriteFunc)
    if err != nil {
        return nil, err
    }

    if newExprs == nil {
        return r, nil
    }

    nt := *r
    nt.updateFilters(newExprs[0], newExprs[1])
    return &nt, nil
}

...

func (v *VirtualTableReadRel) CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), newInputs ...Rel) (Rel, error) {
    return copyRewriteHelper(v, rewriteFunc)
}
```

plan/relations.go
```diff
@@ -194,6 +232,24 @@ func (v *VirtualTableReadRel) ToProtoPlanRel() *proto.PlanRel {
 	}
 }
 
+func (v *VirtualTableReadRel) Copy(newInputs ...Rel) (Rel, error) {
+	return v, nil
+}
+
+func (v *VirtualTableReadRel) CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), newInputs ...Rel) (Rel, error) {
```
zeroshade, line 239: for cases where we know that we're going to ignore / not use the `newInputs` parameter, you should use `_` as the var name to indicate that. 

i.e. 

```go
func (v *VirtualTableReadRel) CopyWithExpressionRewrite(rewriteFunc func(expr.Expression) (expr.Expression, error), _ ...Rel) (Rel, error) {
```

plan/plan.go
zeroshade, line 290: Given the frequency with which we use `func(expr.Expression) (expr.Expression, error)` should we alias that to some name?

```go
type RewriteFunc func(expr.Expression) (expr.Expression, error)

...

CopyWithExpressionRewrite(fn RewriteFunc, inputs ...Rel) (Rel, error)
```
?

plan/relations.go
```diff
@@ -564,6 +690,44 @@ func (j *JoinRel) ToProtoPlanRel() *proto.PlanRel {
...
+	}
+	currExprs := []expr.Expression{j.expr, j.postJoinFilter}
+	exprs := make([]expr.Expression, len(currExprs))
+	for i, e := range currExprs {
+		newExpr, err := rewriteFunc(e)
+		if err != nil {
+			return nil, err
+		}
+		exprs[i] = newExpr
+	}
```
zeroshade, line 719: perhaps:

```go
var (
    exprs = make([]expr.Expression, 2)
    err error
)

for i, e := range []expr.Expression{j.expr, j.postJoinFilter} {
    if exprs[i], err = rewriteFunc(e); err != nil {
        return nil, err
    }
}
```

plan/relations.go
```diff
@@ -449,6 +541,40 @@ func (p *ProjectRel) ToProtoPlanRel() *proto.PlanRel {
...
+	for i, e := range p.exprs {
+		newExpr, err := rewriteFunc(e)
+		if err != nil {
+			return nil, err
+		}
+		exprs[i] = newExpr
+	}
+	if areExpressionsEqual(exprs, p.exprs) && areRelsEqual(newInputs, p.GetInputs()) {
+		return nil, nil
+	}
```
zeroshade, line 571: If these are only checking that the underlying pointers the same you can just use `slices.Equal(exprs, p.exprs)` instead and avoid having to make your own functions for this. Same goes for the rels.

plan/relations.go
```diff
@@ -564,6 +690,44 @@ func (j *JoinRel) ToProtoPlanRel() *proto.PlanRel {
...
+		exprs[i] = newExpr
+	}
+	if areExpressionsEqual(exprs, currExprs) && areRelsEqual(newInputs, j.GetInputs()) {
+		return nil, nil
+	}
+	join := *j
+	join.left = newInputs[0]
+	join.right = newInputs[1]
+	join.expr = exprs[0]
+	join.postJoinFilter = exprs[1]
```
zeroshade, line 727: Could we do

```go
join := j.Copy(newInputs...)
join.expr, join.postJoinFilter = exprs[0], exprs[1]
```

instead?

plan/relations.go
```diff
@@ -745,6 +953,63 @@ func (ar *AggregateRel) ToProtoPlanRel() *proto.PlanRel {
...
+			if err != nil {
+				return nil, err
+			}
+			if newExpr != e {
+				groupsAreEqual = false
+			}
+			newGroup[j] = newExpr
+		}
+		newGroups[i] = newGroup
+	}
```
zeroshade, line 988: Similar to the above we can simplify this:

```go
var (
    newGroups = make([][]expr.Expressions, len(ar.groups)
    groupsAreEqual = true
    err error
)

for i, g := range ar.groups {
    newGroups[i] = make([]expr.Expression, len(g))
    for j, e := range g {
        if newGroups[i][j], err = rewriteFunc(e); err != nil {
             return nil, err
        }
        groupsAreEqual = groupsAreEqual && newGroups[i][j] == e
    }
}
```

Something similar can be done for the measures loop below

plan/relations.go
```diff
@@ -787,6 +1052,47 @@ func (sr *SortRel) ToProtoPlanRel() *proto.PlanRel {
...
+			return nil, err
+		}
+		if newExpr != s.Expr {
+			sortExpressionsAreEqual = false
+		}
+		sorts[i] = expr.SortField{
+			Expr: newExpr,
+			Kind: s.Kind,
+		}
+	}
```
zeroshade, line 1086: same simplifications as above

plan/relations.go
scgkiran: changed this to use slices.Equal()

plan/relations.go
scgkiran: Did this partially. Using j.Copy asks for a type cast to JoinRel.

plan/relations.go
scgkiran: I hit some issues when I tried this.
1. `func copyRewriteHelper[T exprRewriter](r *T, rewriteFunc RewriteFunc) (*T, error)`
       - r is a pointer to interface, complains about `unresolved copyReference`
       - r is a pointer to interface, so `nt := *r` won't copy the object?
2. `func copyRewriteHelper[T exprRewriter](r T, rewriteFunc RewriteFunc) (*T, error)`
       - `nt := *r` this throws error `invalid indirect of type T`. Looks like we cannot copy object using interface type variable.

scgkiran: Below works, but still some code duplication of is there.

```
type exprRewriter interface {
	copyExpressions(rewriteFunc RewriteFunc) ([]expr.Expression, error)
	getExpressions() []expr.Expression
	updateFilters(filter []expr.Expression)
	copy() Rel
}

func copyRewriteHelper(n exprRewriter, rewriteFunc RewriteFunc) (Rel, error) {
	newExprs, err := n.copyExpressions(rewriteFunc)
	if err != nil {
		return nil, err
	}
	if slices.Equal(newExprs, n.getExpressions()) {
		return n.(Rel), nil
	}
	nt := n.copy()
	nt.(exprRewriter).updateFilters(newExprs)
	return nt, nil
}

func (n *NamedTableReadRel) copy() Rel {
	ntr := *n
	return &ntr
}

func (n *NamedTableReadRel) CopyWithExpressionRewrite1(rewriteFunc RewriteFunc, _ ...Rel) (Rel, error) {
	return copyRewriteHelper(n, rewriteFunc)
}
```

jacques-n (approved):
This looks good to me. I see most of the feedback that zeroshade provided has been addressed. I think we should merge this. If the remaining items are not sufficient, let's address in a follow-up ticket.

PR #37, opened 2024-07-29
ci(build-test): Update codecov to ignore protobuf files and String() methods

jacques-n:


jacques-n:
Not entirely sure this will work but don't think I can tell until merged into main so merging despite lack of confidence around codecov.yml format.

PR #38, opened 2024-07-29
ci(build-test): Use grep to exclude protobuf from coverage report

jacques-n:


jacques-n:
Validated that coverage went up as expected.

Issue #33, opened 2024-07-29
Introduce a Literal package and provide clean utility methods for construction of literals


jacques-n:
Literals right now are a bit difficult to work with. Constructing them generally requires a bunch of sleuthing. There aren't a lot of docs and protobuf objects are leaked through the provided apis (e.g. `ProtoLiteral`).

Simple example. Assume we have a variable days which is an int32 of days since epoch. 

Current
```
func NewLiteral[T allLiteralTypes](val T, nullable bool) (Literal, error)

// example usage.
expr.NewLiteral[types.Date](types.Date(days), true)
```

New Utility Method

```
func NewDate(days int) (Literal, error)

// example usage.
literal.NewDate(days)
```

We can also add additional methods for common conversion cases. For example, maybe we add:

```
NewDecimalFromTwosComplement(scale int, []byte twoscomplement)
NewDecimalFromString(str string)
```

etc.

As part of this we should also ensure we have reasonable test coverage for the new methods

jacques-n:
Completed in #41

PR #41, opened 2024-08-07
feat(type): add support for type PrecisionTimestamp and PrecisionTimestampTz

anshuldata:
* Introduced type "PrecisionTimeStampType" and "PrecisionTimeStampTzType" which implements [Type](https://github.com/substrait-io/substrait-go/blob/0ea5482e061033854f9931e2134a1bf91a5bbb54/types/types.go#L255) interface
* Support for Literal expression using PrecisionTimestamp and PrecisionTimestampTz
* Fixed "ProtoLiteral" ToProtoLiteral API to switch on type instead of value. Felt this is cleaner and extensible since converting based on type. Used "ProtoLiteral" to create Literal interface for PrecisionTimestamp and PrecisionTimestampTz


go.mod
```diff
@@ -6,23 +6,29 @@ go 1.20
 
 require (
 	github.com/alecthomas/participle/v2 v2.0.0
+	github.com/cockroachdb/errors v1.11.3
```
jacques-n, line 9: i don't think it makes sense to source the cockroachdb errors package for a test. Let's revert the go.mod/go.sum changes.

expr/literals.go
```diff
@@ -924,6 +946,42 @@ func LiteralFromProto(l *proto.Expression_Literal) Literal {
...
+		}
+		if precTimeStamp.Value < 0 {
+			return nil
+		}
+		return NewPrecisionTimestampTzLiteral(uint64(precTimeStamp.Value), precision, nullability)
 	}
 	panic("unimplemented literal type")
 }
+
+func NewPrecisionTimestampLiteral(value uint64, precision types.TimePrecision, n types.Nullability) Literal {
```
jacques-n, line 973: please add docs to this function and the one below.

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,137 @@
...
+	case 0:
+		return Seconds, nil
+	case 3:
+		return Milliseconds, nil
+	case 6:
+		return Microseconds, nil
+	case 9:
+		return Nanoseconds, nil
+	}
+	return UnknownPrecision, errors.Newf("invalid TimePrecision value %d", val)
```
jacques-n, line 44: I believe 0-9 are all allowed. Why throw here? Are you making the design decision that substrait-go won't support 1,2,4,5,7,8? If so, maybe update methods to say that in the error.

jacques-n (commented):
It looks like a bunch of the code you introduced [doesn't have test cases](https://app.codecov.io/gh/substrait-io/substrait-go/pull/41?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=substrait-io)

Can you resolve that as well as the other comments? We should at least have test cases for the production paths (even if we don't for things like String() and weird error conditions)

go.mod
anshuldata: This is a utility library which has useful function to construct meaningful error. So good to keep it

expr/literals.go
anshuldata: done

types/precison_timestamp_types.go
anshuldata: fixed it

anshuldata:
>  We should at least have test cases for the production paths (even if we don't for things like String() and weird error conditions)

Added unit tests for below parts
1. Creation of PrecisionTimeStamp and PrecisionTimeStampTz type
2. Conversion to proto from both types
3. Conversion from proto for both types

I found a bug too in my code because of above. Thanks

anshuldata:
We will add tests for other types (existing ones) as part of #33

expr/literals.go
```diff
@@ -7,13 +7,13 @@ package expr
 import (
 	"bytes"
 	"fmt"
+	"google.golang.org/protobuf/types/known/anypb"
```
zeroshade, line 10: why did this import move? Can we please put it back in the import list with the rest of the imports that are not stdlib imports?

expr/literals.go
```diff
@@ -404,6 +404,10 @@ type ProtoLiteral struct {
 func (*ProtoLiteral) isRootRef()            {}
 func (t *ProtoLiteral) GetType() types.Type { return t.Type }
 func (t *ProtoLiteral) String() string {
+	switch literalType := t.Type.(type) {
+	case types.PrecisionTimeStampType, types.PrecisionTimeStampTzType:
+		return fmt.Sprintf("%s(%d)", literalType, t.Value.(uint64))
```
zeroshade, line 409: it's likely not necessary to do the type assertion here with Sprintf which should properly handle the interface type anyways.

go.mod
zeroshade: how is it different than just using `fmt.Errorf` instead? I agree with @jacques-n that I don't see the benefit in the cockroachdb errors package. What functionality is it providing that we can't already get from `errors.New` and `fmt.Errorf`?

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	case 6:
+		return MicroSeconds, nil
+	case 7:
+		return EMinus7Seconds, nil
+	case 8:
+		return EMinus8Seconds, nil
+	case 9:
+		return NanoSeconds, nil
+	}
+	return UnknownPrecision, errors.Newf("invalid TimePrecision value %d", val)
```
zeroshade, line 74:
```suggestion
        return UnknownPrecision, fmt.Errorf("invalid TimePrecision value: %d", val)
```

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
+package types
+
+import (
+	"fmt"
+	"github.com/cockroachdb/errors"
+	"github.com/substrait-io/substrait-go/proto"
+)
+
+type TimePrecision int
```
zeroshade, line 9: doc string?

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	case MicroSeconds:
+		return 6
+	case EMinus7Seconds:
+		return 7
+	case EMinus8Seconds:
+		return 8
+	case NanoSeconds:
+		return 9
+	}
+	panic("unreachable")
```
zeroshade, line 48: This isn't unreachable, what if they pass `UnknownPrecision`?

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	case EMinus5Seconds:
+		return 5
+	case MicroSeconds:
+		return 6
+	case EMinus7Seconds:
+		return 7
+	case EMinus8Seconds:
+		return 8
+	case NanoSeconds:
+		return 9
```
zeroshade, line 46: I'm confused by this function, why not change the enum constants so that the values match these values? thus conversion would just be `int(val)` ?

You could do this by making `UnknownPrecision = -1` and then use `iota` starting with `Seconds`

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	DeciSeconds
+	CentiSeconds
+	MilliSeconds
+	EMinus4Seconds // 10^-4 of seconds
+	EMinus5Seconds // 10^-5 of seconds
+	MicroSeconds
+	EMinus7Seconds // 10^-7 of seconds
+	EMinus8Seconds // 10^-8 of seconds
+	NanoSeconds
+)
```
zeroshade, line 23: I'd prefer for `Precision` to be in the names of these. Otherwise `types.Seconds` could potentially be confusing as opposed to `types.SecondsPrecision` or something to that extent. 

Also, is there actually a need for the Deci/Centi/EMinus4/EMinus5, etc. types? Why is Seconds, Milliseconds, Microseconds, Nanoseconds not sufficient?

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	case 5:
+		return EMinus5Seconds, nil
+	case 6:
+		return MicroSeconds, nil
+	case 7:
+		return EMinus7Seconds, nil
+	case 8:
+		return EMinus8Seconds, nil
+	case 9:
+		return NanoSeconds, nil
```
zeroshade, line 72: same point as above, just make the values match so that we can do `TimePrecision(val)` to get the correct time precision, rather than needing the whole switch statement.

types/precison_timestamp_types.go
zeroshade, line 9: the conversion functions take and return `int32` so why are you using an `int` as the type instead of just defining this as an `int32` ?

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return UnknownPrecision, errors.Newf("invalid TimePrecision value %d", val)
+}
+
+type PrecisionTimeStampType struct {
```
zeroshade, line 77: needs a doc string?

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	precision        TimePrecision
+	typeVariationRef uint32
+	nullability      Nullability
+}
+
+func NewPrecisionTimestampType(precision TimePrecision) PrecisionTimeStampType {
```
zeroshade, line 83: doc string?

zeroshade (changes requested):
A few nitpicks

expr/literals.go
jacques-n: as elsewhere, -1 on review comments around import order. If we want a specific order, add as precommit tests.

types/precison_timestamp_types.go
jacques-n: > Why is Seconds, Milliseconds, Microseconds, Nanoseconds not sufficient?

The standard supports all sizes and so do most databases (0..9).

jacques-n: I agree on Precision but I suggest we put first as opposed to last. Makes easier to find similar options in IDE. E.g. types.PrecisionMilliseconds

types/precison_timestamp_types.go
jacques-n: agree on improvement with ^

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,167 @@
...
+	MilliSeconds
+	EMinus4Seconds // 10^-4 of seconds
+	EMinus5Seconds // 10^-5 of seconds
+	MicroSeconds
+	EMinus7Seconds // 10^-7 of seconds
+	EMinus8Seconds // 10^-8 of seconds
+	NanoSeconds
+)
+
+func timePrecisionToProtoVal(val TimePrecision) int32 {
```
jacques-n, line 25: inclined to return error as opposed to panic.

jacques-n:
Note that your patch only appears to have 34% code coverage. We should strive for patches to meet or exceed existing coverage.

go.mod
anshuldata: I felt instead of using some methods from fmt and some from standard "errors". It is good to use one library. This cockroach library has much richer APIs (e.g. adding Hint to errors) too. Hence I used it so that we can use a single lib for errors and augment errors to be more informative
Though, for this task I could use existing fmt.Errorf. So I reverted back to fmt instead of using cockroach/errors. I still feel this API will be useful to report more meaningful error. I will reintroduce in future when I use APIs of this lib which aren't available in standard package

expr/literals.go
anshuldata: yes, it should be fixed by pre-commit. I will add a pre-commit action to detect imports in future PR

expr/literals.go
anshuldata: fixed it

types/precison_timestamp_types.go
anshuldata: Replied in [this](https://github.com/substrait-io/substrait-go/pull/41#discussion_r1710861808) comment
Since currently, reverted to use fmt.Errorf so resolving this comment

types/precison_timestamp_types.go
anshuldata: fixed it, kindly recheck

types/precison_timestamp_types.go
anshuldata: fixed it, kindly recheck

types/precison_timestamp_types.go
anshuldata: fixed it, kindly recheck

types/precison_timestamp_types.go
anshuldata: fixed it, kindly recheck

types/precison_timestamp_types.go
anshuldata: fixed it

types/precison_timestamp_types.go
anshuldata: fixed it

types/precison_timestamp_types.go
anshuldata: fixed it

anshuldata:
> Note that your patch only appears to have 34% code coverage. We should strive for patches to meet or exceed existing coverage.

Fixed it by adding more tests which covers code which was missing coverage.

types/precison_timestamp_types.go
anshuldata: fixed it

types/precison_timestamp_types.go
anshuldata: fixed it

anshuldata:
> how is it different than just using fmt.Errorf instead? I agree with @jacques-n that I don't see the benefit in the cockroachdb errors package. What functionality is it providing that we can't already get from errors.New and fmt.Errorf?

Replied in [this](https://github.com/substrait-io/substrait-go/pull/41#discussion_r1710861808) comment

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"fmt"
+	"github.com/substrait-io/substrait-go/proto"
+)
+
+// TimePrecision is used to represent the precision of a timestamp
+type TimePrecision int32
+
+const PrecisionUnknown TimePrecision = -1
```
zeroshade, line 11: this doesn't need to be a separate block, it can still be grouped in the block below without changing the values

types/precison_timestamp_types.go
```diff
@@ -0,0 +1,147 @@
...
+	PrecisionEMinus8Seconds // 10^-8 of seconds
+	PrecisionNanoSeconds
+)
+
+func timePrecisionToProtoVal(val TimePrecision) int32 {
+	return int32(val)
+}
+
+func ProtoToTimePrecision(val int32) (TimePrecision, error) {
+	if val < 0 || val > 9 {
```
zeroshade, line 32: use the constants here

`if val < PrecisionSeconds || val > PrecisionNanoSeconds {`

zeroshade (changes requested):
Just a few other nitpicks, everything else looks good to me besides agreeing with @jacques-n that we should raise the test coverage

types/precison_timestamp_types.go
anshuldata: I did try this initially, but iota gets reset to zero on first statement of block and subsequently gets incremented for each statement. If I include unknown, it was giving value 1 for seconds ([goPlayground](https://go.dev/play/p/kK4EhOKAUQw))

anshuldata:
> we should raise the test coverage
I did fix it with last commit and now coverage for type change is around 97%.
For literal it covers mostly PreisionTimeStamp/Tz test cases. For other literals, it should get increased by #33 . It is showing up in my PR because I modified function func (t *ProtoLiteral) ToProtoLiteral() and it has other types too

types/precison_timestamp_types.go
zeroshade: ah interesting, I didn't realize it worked out that way. fair enough

zeroshade (approved):

anshuldata:
Can you merge this @zeroshade . I don't have write permission

Thanks

jacques-n (approved):
Looks good. Test coverage is well above codebase. Thanks @anshuldata

PR #43, opened 2024-08-12
Introduce literal package

scgkiran:


literal/decimal_util.go
```diff
@@ -0,0 +1,73 @@
+package literal
+
+import (
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+func decimalStringToBytes(decimalStr string) ([16]byte, int32, int32, error) {
```
jacques-n, line 9: doc plz

literal/decimal_util.go
```diff
@@ -0,0 +1,73 @@
...
+	var result [16]byte
```
jacques-n, line 10: does it makes since to use a utility library such as https://github.com/cockroachdb/apd avoid hand coding this?

literal/decimal_util.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Split the string into integer and fractional parts
+	parts := strings.Split(decimalStr, ".")
```
jacques-n, line 13: should we trim spaces?

literal/decimal_util.go
jacques-n: also, I suggest this function starts with checking the pattern. e.g. the only valid pattern is
`[+-][\d]{0,38}.[\d]{0,38}` or something like that.

literal/decimal_util_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_decimalStringToBytes(t *testing.T) {
+	tests := []struct {
+		input        string
+		want         string
```
jacques-n, line 15: would be clearer if that had been called something like hexwant. took a minute to grok.

literal/decimal_util_test.go
```diff
@@ -0,0 +1,122 @@
...
+		{"0", "00000000000000000000000000000000", 1, 0, false},
+		{"12345678901234567890123456789012345678901234", "", 44, 0, true},
+		{"abc", "", 0, 0, true},
+		{"12.34.56", "", 0, 0, true},
+		{"199999999999999999999999999999999999999", "", 0, 0, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.input, func(t *testing.T) {
+			got, precision, scale, err := decimalStringToBytes(tt.input)
```
jacques-n, line 55: please extract the actual test of one case into a separate function. makes it much easier to separate business logic from the framework of running tests.

literal/decimal_util_test.go
```diff
@@ -0,0 +1,122 @@
...
+			if (err != nil) != tt.wantErr {
```
jacques-n, line 56: why not use assert utilities more here and below? e.g. asserEqual, assertError, etc.

literal/utils.go
```diff
@@ -0,0 +1,131 @@
...
+
+func NewTimeFromMicros(micros int64) (expr.Literal, error) {
+	if micros < 0 || micros >= (24*time.Hour).Microseconds() {
+		return nil, fmt.Errorf("invalid time value %d", micros)
+	}
+	return expr.NewLiteral[types.Time](types.Time(micros), false)
+}
+
+func NewTimestamp(tm time.Time) (expr.Literal, error) {
+	return expr.NewLiteral[types.Timestamp](types.Timestamp(tm.UnixMicro()), false)
```
jacques-n, line 61: this or above should probably confirm that days are only included in one and not the other. also, need better logic around things like timezone. What is expectation here if someone passes in a timezone?

if time.Time can represent many things, it seems like we need to make sure that when used to construct literals, it represents only right things and translation is reliable.

literal/utils.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func NewTimeFromMicros(micros int64) (expr.Literal, error) {
+	if micros < 0 || micros >= (24*time.Hour).Microseconds() {
+		return nil, fmt.Errorf("invalid time value %d", micros)
+	}
+	return expr.NewLiteral[types.Time](types.Time(micros), false)
+}
+
+func NewTimestamp(tm time.Time) (expr.Literal, error) {
```
jacques-n, line 60: For these ones with conversions, we should clarify how conversion works in doc

literal/utils.go
```diff
@@ -0,0 +1,131 @@
...
+	return expr.NewLiteral[types.Timestamp](types.Timestamp(tm.UnixMicro()), false)
+}
+
+func NewTimestampFromMicros(micros int64) (expr.Literal, error) {
+	return expr.NewLiteral[types.Timestamp](types.Timestamp(micros), false)
+}
+
+func NewTimestampTZ(tm time.Time) (expr.Literal, error) {
```
jacques-n, line 68: don't forget the variable precision timestamps (just added to substrait-go)

literal/decimal_util.go
zeroshade: We already include "math/big", we can use big.ParseFloat, or we can use the Go Arrow library decimal128 package which has a FromString method, etc. Personally I'd prefer those

literal/utils.go
zeroshade: I agree with @jacques-n here, that we need to clarify what happens if the passed in time has a location that isn't UTC. Do we ignore it and just use the UTC value? Do we handle it? etc.

literal/decimal_util.go
zeroshade: Another example that can be used is https://github.com/zeroshade/iceberg-go/blob/main/literals.go#L996

literal/decimal_util_test.go
```diff
@@ -0,0 +1,122 @@
...
+				t.Errorf("decimalStringToBytes() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if tt.wantErr {
+				return
+			}
+			assert.Len(t, got, 16)
+			if !strings.EqualFold(tt.want, toHex(got[:])) {
+				t.Errorf("decimalStringToBytes() got = %v, want %v", toHex(got[:]), tt.want)
+			}
```
zeroshade, line 66: rather than converting the value to hex and doing strings.EqualFold, it might make more sense to convert the `want` from hex back to bytes, and then do `assert.Equal(....)` instead

literal/decimal_util_test.go
```diff
@@ -0,0 +1,122 @@
...
+	// Convert the byte array to a big.Int
+	intValue := new(big.Int).SetBytes(decimalBytes[:])
+
+	// Convert the big.Int to a string
+	decimalStr := intValue.String()
+
+	// Insert the decimal point at the appropriate position
+	if scale > 0 {
+		decimalStr = decimalStr[:len(decimalStr)-int(scale)] + "." + decimalStr[len(decimalStr)-int(scale):]
+	}
```
zeroshade, line 104: given the precision and scale, you could use `big.Float` to perform the string conversion for you rather than having to implement it yourself.

literal/utils.go
scgkiran: added doc

literal/utils.go
scgkiran: I added docs.

Time literal only represents time of the day (in microseconds)
Both Timestamp and TimestampTz use only microseconds in proto. No timezone field in proto for TimestampTz. Both APIs that I added uses microsecs since epoch of UTC. 
Timestamp (is timezone unaware)
TimestampTz (timezone aware) 

My understanding: A column with Timestamp type cannot be used with a timezone. It is a timestamp in unspecified time zone. 
A column with TimestampTz has a timestamp in micros seconds since epoch in UTC timezone. It can be localised to any timezone, as translations and comparison can be done correctly.

literal/utils.go
scgkiran: I will add them in next PR

literal/utils.go
jacques-n: Can we error on arguments that don't make sense? For example, if I pass in a time.Time that includes days to to the time literal creator.

Also, if time.Time is carrying a timezone, how is that interpereted in the timezoneless time? Converted to UTC first?

scgkiran: time.Time internally stores only seconds, nanoSeconds (since 1 Jan 1885) and location. time.UnixMicro() gives micro seconds since 1 Jan 1971 UTC. This is what we need in Timestamp & TimestampTz.

Logically time.Time always stores (secs & nanosecs) in UTC. So no timezone conversion needed when getting UnixMicro().

location in Time used only to get Year(), Month(), Day(), hour(), min() 

Given the above details. I feel NewTime(time.Time) need not throw error and just take timeOfTheDay in micro seconds to create the literal. I think I captured this in the doc. May be the doc can be improved?
 or may be get rid of this API if this is going to be confusing for users? Just provide only NewTimeFromMicros(micros int64).

jacques-n: The confusion I have is if I pass in a time for something like 12/30/2024 2:30pm into `NewTime`, what do I get? Would I get 2:30pm and it would silently ignore the date? Would it store the date? From my pov, I would expect to error and say something like "time not within allowed range". Looking at the spec, it is listed at the following valid timerange: 0..86,399,999,999 microseconds. Outside that should error.

scgkiran: We do range check in NewTimeFromMicros.

NewTime(time.Time) current code takes 2.30 pm and ignores the date part. My point is if we throw error in this case, this api is not so useful. For some one to use this api, they have to create a time like "1 Jan 1971 2:30PM" and pass it. Should we get rid of this API?

May be a better api is NewTime(hours, minutes, seconds, microsecs int32).

jacques-n: yeah, that sounds like better api.

we should range check it. someone can't say 25 hours, for example.

literal/decimal_util.go
```diff
@@ -0,0 +1,68 @@
+package literal
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/cockroachdb/apd"
```
zeroshade, line 8: If we're going to use this, we should use the most recent version (https://pkg.go.dev/github.com/cockroachdb/apd/v3) please change the import accordingly (add the `/v3`)

literal/decimal_util.go
```diff
@@ -0,0 +1,68 @@
...
+// This function also returns the precision and scale of the decimal value.
+// The precision is the total number of digits in the decimal value. The precision is limited to 38 digits.
+// The scale is the number of digits to the right of the decimal point. The scale is limited to the precision.
+func decimalStringToBytes(decimalStr string) ([16]byte, int32, int32, error) {
+	var result [16]byte
+
+	strings.Trim(decimalStr, " ")
+	if !decimalPattern.MatchString(decimalStr) {
+		return result, 0, 0, fmt.Errorf("invalid decimal string")
+	}
```
zeroshade, line 24: does the `apd.NewFromString` perform this validation for us such that we don't need to explicitly do it ourselves?

literal/decimal_util_test.go
```diff
@@ -0,0 +1,103 @@
...
+			testDecimalStringToBytes(t, tt.input, tt.hexWant, tt.expPrecision, tt.expScale)
+		})
+	}
+
+	badInputs := []string{
+		"12345678901234567890123456789012345678901234",
+		"abc",
+		"12.34.56",
+		"199999999999999999999999999999999999999",
+		"1.23e20",
```
zeroshade, line 59: This should be a valid input IMHO, why would this be invalid?

zeroshade (changes requested):

literal/decimal_util_test.go
scgkiran: Not sure about this. ChatGpt says it is not a valid literal for decimal type.
I tried this on duckdb: select `1023e23` it recognised this as `double`. When I tried to insert this in a `decimal(38, 9)` column, it actually stored `123000000000000003008.758808576`
Postgres support this notation.
apd.NewFromString also supports this notation.

literal/decimal_util.go
scgkiran: not exactly it supports scientific notation, inf, nan

literal/decimal_util_test.go
jacques-n: I think it is okay for our utility library to provide more options for input than what a random database may provide. They are utility functions afterall. We should just document what is allowed.

literal/decimal_util.go
scgkiran: done

literal/decimal_util_test.go
scgkiran: added

jacques-n:
I got confused and merged this patch as part of #47 .  Didn't realize that was a branch built on this change.

In future, @scgkiran can you avoid building PRs that incorporate other PRs? Thanks

scgkiran:
> I got confused and merged this patch as part of #47 . Didn't realize that was a branch built on this change.
> 
> In future, @scgkiran can you avoid building PRs that incorporate other PRs? Thanks

Ok

PR #44, opened 2024-08-12
feat(type): Add support for interval compound literal

anshuldata:
* Introduced APIs to create Literal of type IntervalCompound and IntervalYearToMonth

### DIFF
--- a/expr/interval_compound_literal.go
+++ b/expr/interval_compound_literal.go
@@ -69,22 +69,22 @@ func newIntervalPartsValInternal(datePartsOptions ...intervalDatePartsOptions) *
 	return intervalPartsVal
 }
 
-func intervalPartsValToProto(idp *intervalDateParts, ict *types.IntervalCompoundType) *proto.Expression_Literal_IntervalCompound {
+func (m *intervalDateParts) ToProto(ict *types.IntervalCompoundType) *proto.Expression_Literal_IntervalCompound {
 	intrCompPB := &proto.Expression_Literal_IntervalCompound{}
 
-	if idp.years > 0 || idp.months > 0 {
+	if m.years > 0 || m.months > 0 {
 		yearToMonthProto := &proto.Expression_Literal_IntervalYearToMonth{
-			Years:  idp.years,
-			Months: idp.months,
+			Years:  m.years,
+			Months: m.months,
 		}
 		intrCompPB.IntervalYearToMonth = yearToMonthProto
 	}
 
 	dayToSecondProto := &proto.Expression_Literal_IntervalDayToSecond{
-		Days:          idp.days,
-		Seconds:       idp.seconds,
+		Days:          m.days,
+		Seconds:       m.seconds,
 		PrecisionMode: &proto.Expression_Literal_IntervalDayToSecond_Precision{Precision: ict.GetPrecisionProtoVal()},
-		Subseconds:    idp.subSeconds,
+		Subseconds:    m.subSeconds,
 	}
 	intrCompPB.IntervalDayToSecond = dayToSecondProto
 	return intrCompPB
--- a/expr/interval_compound_literal.go
+++ b/expr/interval_compound_literal.go
@@ -1,0 +1,156 @@
+package expr
+
+import (
+	"errors"
+	"github.com/substrait-io/substrait-go/proto"
+	"github.com/substrait-io/substrait-go/types"
+)
+
+type intervalDateParts struct {
+	years      int32
+	months     int32
+	days       int32
+	seconds    int32
+	subSeconds int64
+}
+
+type intervalDatePartsOptions func(parts *intervalDateParts)
+
+func WithIntervalCompoundYears(years int32) func(*intervalDateParts) {
+	return func(idp *intervalDateParts) {
+		idp.years = years
+	}
+}
+
+func WithIntervalCompoundMonths(months int32) func(*intervalDateParts) {
+	return func(idp *intervalDateParts) {
+		idp.months = months
+	}
+}
+
+func WithIntervalCompoundDays(days int32) func(*intervalDateParts) {
+	return func(idp *intervalDateParts) {
+		idp.days = days
+	}
+}
+
+func WithIntervalCompoundSeconds(seconds int32) func(*intervalDateParts) {
+	return func(idp *intervalDateParts) {
+		idp.seconds = seconds
+	}
+}
+
+func WithIntervalCompoundSubSeconds(subSeconds int64) func(*intervalDateParts) {
+	return func(idp *intervalDateParts) {
+		idp.subSeconds = subSeconds
+	}
+}
+
+// NewIntervalLiteralUptoSubSecondPrecision creates an interval literal which allows upto subsecond precision
+// arguments: precision and nullable property (n)
+// datePartsOptions is options to set value parts (month, year, day, seconds, subseconds).
+// If multiple options of same types (e.g. multiple second options) are provided only value of last part is considered
+func NewIntervalLiteralUptoSubSecondPrecision(precision types.TimePrecision, n types.Nullability, datePartsOptions ...intervalDatePartsOptions) Literal {
+	intervalCompoundType := types.NewIntervalCompoundType(precision).WithNullability(n)
+	intervalPartsVal := newIntervalPartsValInternal(datePartsOptions...)
+	return &ProtoLiteral{
+		Value: intervalPartsVal,
+		Type:  intervalCompoundType,
+	}
+}
+
+func newIntervalPartsValInternal(datePartsOptions ...intervalDatePartsOptions) *intervalDateParts {
+	intervalPartsVal := &intervalDateParts{}
+	for _, datePart := range datePartsOptions {
+		datePart(intervalPartsVal)
+	}
+	return intervalPartsVal
+}
+
+func intervalPartsValToProto(idp *intervalDateParts, ict *types.IntervalCompoundType) *proto.Expression_Literal_IntervalCompound {
+	intrCompPB := &proto.Expression_Literal_IntervalCompound{}
+
+	if idp.years > 0 || idp.months > 0 {
+		yearToMonthProto := &proto.Expression_Literal_IntervalYearToMonth{
+			Years:  idp.years,
+			Months: idp.months,
+		}
+		intrCompPB.IntervalYearToMonth = yearToMonthProto
+	}
+	if idp.days == 0 && idp.seconds == 0 && idp.subSeconds == 0 {
+		// all parts are >= month granularity so no need to set daytosecond component
+		return intrCompPB
+	}
+
+	dayToSecondProto := &proto.Expression_Literal_IntervalDayToSecond{
+		Days:          idp.days,
+		Seconds:       idp.seconds,
+		PrecisionMode: &proto.Expression_Literal_IntervalDayToSecond_Precision{Precision: ict.GetPrecisionProtoVal()},
+		Subseconds:    idp.subSeconds,
+	}
+	intrCompPB.IntervalDayToSecond = dayToSecondProto
+	return intrCompPB
+}
+
+func intervalCompoundLiteralFromProto(protoVal *proto.Expression_Literal_IntervalCompound, nullability types.Nullability) Literal {
+	var datePartsOptions []intervalDatePartsOptions
+	if protoVal.IntervalYearToMonth != nil {
+		if protoVal.IntervalYearToMonth.Years > 0 {
+			datePartsOptions = append(datePartsOptions, WithIntervalCompoundYears(protoVal.IntervalYearToMonth.Years))
+		}
+		if protoVal.IntervalYearToMonth.Months > 0 {
+			datePartsOptions = append(datePartsOptions, WithIntervalCompoundMonths(protoVal.IntervalYearToMonth.Months))
+		}
+	}
+	idts := protoVal.IntervalDayToSecond
+	err := validateIntervalDayToSecondProto(idts)
+	if err != nil {
+		return nil
+	}
+	precision, err := intervalCompoundPrecisionFromProto(idts)
+	if err != nil {
+		return nil
+	}
+	if idts.Days > 0 {
+		datePartsOptions = append(datePartsOptions, WithIntervalCompoundDays(protoVal.IntervalDayToSecond.Days))
+	}
+	if idts.Seconds > 0 {
+		datePartsOptions = append(datePartsOptions, WithIntervalCompoundSeconds(protoVal.IntervalDayToSecond.Seconds))
+	}
+	if idts.Subseconds > 0 {
+		datePartsOptions = append(datePartsOptions, WithIntervalCompoundSubSeconds(protoVal.IntervalDayToSecond.Subseconds))
+	} else if val, ok := idts.PrecisionMode.(*proto.Expression_Literal_IntervalDayToSecond_Microseconds); ok {
+		// deprecated field microsecond is set, set its value as subsecond
+		datePartsOptions = append(datePartsOptions, WithIntervalCompoundSubSeconds(int64(val.Microseconds)))
+	}
+	return NewIntervalLiteralUptoSubSecondPrecision(precision, nullability, datePartsOptions...)
+}
+
+func validateIntervalDayToSecondProto(idts *proto.Expression_Literal_IntervalDayToSecond) error {
+	if idts.PrecisionMode == nil {
+		// error, precision mode must be set for intervalCompound
+		return errors.New("missing precision mode for interval compound")
+	}
+	if _, ok := idts.PrecisionMode.(*proto.Expression_Literal_IntervalDayToSecond_Microseconds); ok {
+		// if microsecond precision then subseconds must be set to zero
+		if idts.Subseconds > 0 {
+			return errors.New("both deprecated microseconds and subseconds can't be non zero")
+		}
+	}
+	return nil
+}
+
+func intervalCompoundPrecisionFromProto(protoVal *proto.Expression_Literal_IntervalDayToSecond) (types.TimePrecision, error) {
+	var precisionVal int32
+	switch pmt := protoVal.PrecisionMode.(type) {
+	case *proto.Expression_Literal_IntervalDayToSecond_Precision:
+		precisionVal = pmt.Precision
+	case *proto.Expression_Literal_IntervalDayToSecond_Microseconds:
+		precisionVal = types.PrecisionMicroSeconds.ToProtoVal()
+	}
+	precision, err := types.ProtoToTimePrecision(precisionVal)
+	if err != nil {
+		return types.PrecisionUnknown, err
+	}
+	return precision, nil
+}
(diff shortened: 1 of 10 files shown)


zeroshade (changes requested):

expr/interval_year_to_month.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// NewIntervalLiteralUptoMonth creates an interval literal which allows setting only year and month
+// arguments: nullable property (n), years and months
+func NewIntervalLiteralUptoMonth(n types.Nullability, years int32, months int32) Literal {
+	intervalCompoundType := types.NewIntervalYearToMonthType().WithNullability(n)
+	intervalPartsVal := &intervalYearMonthVal{
+		years:  years,
+		months: months,
+	}
```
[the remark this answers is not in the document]
anshuldata: this structure represents value. proto carries type related information too (e.g. [Precision](https://github.com/substrait-io/substrait/blob/28025cbaa8dc3c65b736d8a68fa7070c465fb494/proto/substrait/algebra.proto#L924)). My intention is to construct value by value related fields.
We have another object to represent type so wanted to keep it separate so that separation of value/type remain clear

expr/interval_year_to_month.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+type intervalYearMonthVal struct {
+	years  int32
+	months int32
+}
+
+// NewIntervalLiteralUptoMonth creates an interval literal which allows setting only year and month
+// arguments: nullable property (n), years and months
+func NewIntervalLiteralUptoMonth(n types.Nullability, years int32, months int32) Literal {
```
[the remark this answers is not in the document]
anshuldata: fixed it, Thanks

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,154 @@
...
+
+func newIntervalPartsValInternal(datePartsOptions ...intervalDatePartsOptions) *intervalDateParts {
+	intervalPartsVal := &intervalDateParts{}
+	for _, datePart := range datePartsOptions {
+		datePart(intervalPartsVal)
+	}
+	return intervalPartsVal
+}
+
+func intervalPartsValToProto(idp *intervalDateParts, ict *types.IntervalCompoundType) *proto.Expression_Literal_IntervalCompound {
```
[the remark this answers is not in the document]
anshuldata: fixed it, Thanks

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/substrait-io/substrait-go/types"
+)
+
+type intervalDateParts struct {
+	years      int32
+	months     int32
+	days       int32
+	seconds    int32
+	subSeconds int64
+}
```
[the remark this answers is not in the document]
anshuldata: this structure represents value. proto carries type related information too (e.g. [Precision](https://github.com/substrait-io/substrait/blob/28025cbaa8dc3c65b736d8a68fa7070c465fb494/proto/substrait/algebra.proto#L924)). My intention is to construct value by value related fields.
We have another object to represent type so wanted to keep it separate so that separation of value/type remain clear

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,154 @@
...
+	return func(idp *intervalDateParts) {
+		idp.seconds = seconds
+	}
+}
+
+func WithIntervalCompoundSubSeconds(subSeconds int64) func(*intervalDateParts) {
+	return func(idp *intervalDateParts) {
+		idp.subSeconds = subSeconds
+	}
+}
```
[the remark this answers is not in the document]
anshuldata: My thought was to keep it simple for user (set what is needed) and to not expose internal representation. Following reasons

1. I went to Option Pattern since there are more then two fields (more the fields then chances of mistakes are more) to set here and user will typically will set only some of them (depending on interval). If user needs to set Month they just need to give option "WithIntervalCompoundMonths" without worrying about other fields.

2. If we represent it as publicly exposed fields of "IntervalDatePartsConfig" and directly using it in our object. We would have to live with always taking this as input which would have made it hard to change internal representation in future. For internal representation to change, we would have to either copy these structure internally or made a breaking change.

Hence I felt option pattern is more maintainable and easier to use for user in this case

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,154 @@
...
+
+func WithIntervalCompoundSeconds(seconds int32) func(*intervalDateParts) {
+	return func(idp *intervalDateParts) {
+		idp.seconds = seconds
+	}
+}
+
+func WithIntervalCompoundSubSeconds(subSeconds int64) func(*intervalDateParts) {
```
jacques-n, line 44: not a fan of this. think it should probably be WithMilliseconds, WithMicroseconds, WithNanoseconds.

plan/relations.go
```diff
@@ -585,9 +585,9 @@ const (
 	JoinTypeOuter       = proto.JoinRel_JOIN_TYPE_OUTER
 	JoinTypeLeft        = proto.JoinRel_JOIN_TYPE_LEFT
 	JoinTypeRight       = proto.JoinRel_JOIN_TYPE_RIGHT
-	JoinTypeSemi        = proto.JoinRel_JOIN_TYPE_SEMI
-	JoinTypeAnti        = proto.JoinRel_JOIN_TYPE_ANTI
-	JoinTypeSingle      = proto.JoinRel_JOIN_TYPE_SINGLE
+	JoinTypeRightSemi   = proto.JoinRel_JOIN_TYPE_RIGHT_SEMI
```
jacques-n, line 588: it looks like some other changes are in this patch. Maybe it needs to be rebased?

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,154 @@
+package expr
+
+import (
+	"errors"
+
+	"github.com/substrait-io/substrait-go/proto"
+	"github.com/substrait-io/substrait-go/types"
+)
+
+type intervalDateParts struct {
```
jacques-n, line 10: I think there is too much composition and cuteness going on in this patch.

Let's just write a simple struct with public fields that implements Literal interface. We shouldn't use protoliteral. We shouldn't use private structs inside of something else.

Sometimes the simplest thing is the best. Let's keep it simple.

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,154 @@
...
+	years      int32
+	months     int32
+	days       int32
+	seconds    int32
+	subSeconds int64
+}
+
+type intervalDatePartsOptions func(parts *intervalDateParts)
+
+func WithIntervalCompoundYears(years int32) func(*intervalDateParts) {
```
jacques-n, line 20: Let's just hang these methods off of the literal object. So if I wanted, I could do:

new IntervalCompoundLiteral{}.WithMillis(1).WithSeconds(2)...etc.

expr/interval_year_to_month.go
```diff
@@ -0,0 +1,36 @@
+package expr
+
+import (
+	"github.com/substrait-io/substrait-go/proto"
+	"github.com/substrait-io/substrait-go/types"
+)
+
+type intervalYearMonthVal struct {
```
jacques-n, line 8: Same feedback as other literal. Let's just create a Literal struct impl that is public and give it utility methods as appropriate.

plan/relations.go
anshuldata: This is needed because of proto update to v.54. [This](https://github.com/substrait-io/substrait/commit/bed84ecb6193c22bf2ff83dc3a391ec5a9a3aa68) PR updated these types

expr/interval_compound_literal.go
anshuldata: fixed it

expr/interval_compound_literal.go
anshuldata: fixed it

expr/interval_year_to_month.go
anshuldata: fixed it

anshuldata:
I updated PR to do away with Option and have separate Literal Impl for IntervalDayToMonth and intervalCompound type. Kindly check @jacques-n @zeroshade

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,246 @@
...
+	return IntervalCompoundLiteral{
+		years:       years,
+		months:      m.months,
+		days:        m.days,
+		seconds:     m.seconds,
+		subSeconds:  m.subSeconds,
+		precision:   m.precision,
+		nullability: m.nullability,
+	}
+}
```
zeroshade, line 32: All of these can be made simpler because you're using a value receiver.

```go
func (m IntervalCompoundLiteral) WithYears(years int32) IntervalCompoundLiteral {
   m.years = years
   return m
}
```

The value receiver is going to create a copy when this is called, so you don't need to explicitly make the copy yourself. Same can be done for the rest of the methods

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		return nil
+	}
+
+	// get subSecond/precision value from proto. To get value it takes care of deprecated microseconds
+	precision, subSeconds, err := intervalCompoundPrecisionSubSecondsFromProto(dayToSecond)
+	if err != nil {
+		return nil
+	}
+	return icLiteral.WithDays(dayToSecond.Days).WithSeconds(dayToSecond.Seconds).WithSubSecond(subSeconds, precision)
```
zeroshade, line 190: we should avoid the numerous copies this will make and just construct the compound literal directly here instead. As written, this is going to create at least 4 copies of this struct which seems kinda silly

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,246 @@
...
+	return false
+}
+
+func (m IntervalCompoundLiteral) ToProtoFuncArg() *proto.FunctionArgument {
+	return &proto.FunctionArgument{
+		ArgType: &proto.FunctionArgument_Value{Value: m.ToProto()},
+	}
+}
+
+func (m IntervalCompoundLiteral) Visit(VisitFunc) Expression { return m }
```
zeroshade, line 212: shouldn't we return `visit(m)` ?

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,246 @@
...
+// IntervalCompoundLiteral creates an interval compound literal
+type IntervalCompoundLiteral struct {
+	years       int32
+	months      int32
+	days        int32
+	seconds     int32
+	subSeconds  int64
+	precision   types.TimePrecision
+	nullability types.Nullability
+}
```
zeroshade, line 20: since all of the members are private, how does a user actually view or get the value in this literal? As written, it's impossible for any library to do anything with this literal except convert it to and from the protobuf object which makes it sort of useless unless we either make the members public or create accessor methods. 

This is why I was using the protobuf object directly, so I didn't have to duplicate all of the code to create the accessor methods

expr/interval_year_to_month.go
```diff
@@ -0,0 +1,92 @@
...
+	nullability types.Nullability
+}
+
+func (m IntervalYearToMonthLiteral) WithYear(years int32) IntervalYearToMonthLiteral {
+	return IntervalYearToMonthLiteral{
+		years:       years,
+		months:      m.months,
+		nullability: m.nullability,
+	}
+}
```
zeroshade, line 23: same comment as with IntervalCompoundLiteral

types/interval_compound_type.go
```diff
@@ -0,0 +1,68 @@
...
+	return m.precision.ToProtoVal()
+}
+
+func (IntervalCompoundType) isRootRef() {}
+func (m IntervalCompoundType) WithNullability(n Nullability) Type {
+	return IntervalCompoundType{
+		precision:   m.precision,
+		nullability: n,
+	}
+}
```
zeroshade, line 31: same comment as earlier

types/interval_compound_type.go
```diff
@@ -0,0 +1,68 @@
...
+	precision        TimePrecision
+	typeVariationRef uint32
+	nullability      Nullability
+}
+
+// NewIntervalCompoundType creates a type of new interval compound.
+func NewIntervalCompoundType() IntervalCompoundType {
+	return IntervalCompoundType{}
+}
+
```
zeroshade, line 20: should the `typeVariationRef` be a parameter? Or do we need a `WithTypeVariationRef` method?

types/interval_year_month_type.go
```diff
@@ -0,0 +1,60 @@
...
+func (IntervalYearToMonthType) isRootRef() {}
+func (m IntervalYearToMonthType) WithNullability(n Nullability) Type {
+	return m.withNullability(n)
+}
+
+func (m IntervalYearToMonthType) withNullability(n Nullability) IntervalYearToMonthType {
+	return IntervalYearToMonthType{
+		nullability: n,
+	}
+}
```
zeroshade, line 32: why both?

types/precison_timestamp_types.go
```diff
@@ -9,20 +9,19 @@ import (
...
+	PrecisionDeciSeconds    TimePrecision = 1
+	PrecisionCentiSeconds   TimePrecision = 2
+	PrecisionMilliSeconds   TimePrecision = 3
+	PrecisionEMinus4Seconds TimePrecision = 4 // 10^-4 of seconds
+	PrecisionEMinus5Seconds TimePrecision = 5 // 10^-5 of seconds
+	PrecisionMicroSeconds   TimePrecision = 6
+	PrecisionEMinus7Seconds TimePrecision = 7 // 10^-7 of seconds
+	PrecisionEMinus8Seconds TimePrecision = 8 // 10^-8 of seconds
+	PrecisionNanoSeconds    TimePrecision = 9
 )
```
zeroshade, line 25: why this change?

plan/relations.go
jacques-n: Best to do that as a separate patch.

types/precison_timestamp_types.go
jacques-n: This was a review comment that I failed to submit on the precision patch. It's unrelated to this and should be done separately.

I felt like the implicit iota didn't clarify that the numbers are meaningful. It seemed too easy for a future editor to change things and not realize the importance. Generally, I don't we should iota unless the number is meaningless (just an artifact of go enum pattern) to our code/business logic. I generally follow a similar pattern across languages. For example, in Java, one should store concrete values if they want them and not try to overload ordinal.

expr/interval_compound_literal.go
jacques-n: @anshuldata, my request was to make the fields and structure public. No reason to hide.

jacques-n: Also suggest we name the precision field  as subsecond precision or similar. Make it clear that only relates to that field...

expr/interval_year_to_month.go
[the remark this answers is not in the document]
anshuldata: Resolving this comment, since I assume (based on other comments) that we agree to not use Proto to construct literal and have a separate struct which implements literal is ok

expr/interval_compound_literal.go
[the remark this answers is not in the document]
anshuldata: Resolving this, since I feel (based on later comments) that we agree to not have proto to represent values

expr/interval_compound_literal.go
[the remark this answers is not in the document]
anshuldata: Resolving this. Since it changed now to use builder pattern and public fields

plan/relations.go
anshuldata: Raised [this](https://github.com/substrait-io/substrait-go/pull/48) PR for proto update. Once it is merged, I will rebase this PR.

expr/interval_compound_literal.go
anshuldata: fixed it, Thanks

expr/interval_compound_literal.go
anshuldata: fixed it

expr/interval_compound_literal.go
anshuldata: I kept is similar to other literal implementations (e.g. [here](https://github.com/substrait-io/substrait-go/blob/cbd28cb19499af1923484ec82540350528249075/expr/literals.go#L501))

expr/interval_compound_literal.go
anshuldata: Fixed both
1. Made fields name Public
2. Renamed "precision" to "SubSecondPrecision"

types/interval_compound_type.go
anshuldata: fixed it

types/interval_compound_type.go
anshuldata: added `WithTypeVariationRef`

types/precison_timestamp_types.go
anshuldata: Raised [this](https://github.com/substrait-io/substrait-go/pull/49) PR to fix it. Removed from this PR

expr/interval_year_to_month.go
anshuldata: fixed it

anshuldata:
I have raised separate PR (#48) for proto update. Once it is merged I will rebase this PR

anshuldata:
rebased PR. It is ready for re-review @jacques-n @zeroshade

anshuldata:
rebased it

expr/interval_compound_literal.go
```diff
@@ -0,0 +1,192 @@
...
+func (m IntervalCompoundLiteral) WithSubSecond(subSeconds int64, precision types.TimePrecision) IntervalCompoundLiteral {
+	m.SubSeconds = subSeconds
+	m.SubSecondPrecision = precision
+	return m
+}
+
+func (m IntervalCompoundLiteral) WithNullability(nullability types.Nullability) IntervalCompoundLiteral {
+	m.Nullability = nullability
+	return m
+}
```
zeroshade, line 69: since the members are all public now, do we still need these at all?

expr/interval_compound_literal_test.go
```diff
@@ -0,0 +1,298 @@
...
+	var origYearVal, origMonthVal, origDayVal, origSecondsVal int32
+	var origSubSecondsVal int64
+	origYearVal = 1
+	origMonthVal = -2
+	origDayVal = 3
+	origSecondsVal = -4
+	origNullabilityVal := types.NullabilityRequired
+	origPrecisionVal := types.PrecisionEMinus7Seconds
+	origSubSecondsVal = 5
+	origLiteral := IntervalCompoundLiteral{Years: origYearVal, Months: origMonthVal, Days: origDayVal, Seconds: origSecondsVal, SubSeconds: origSubSecondsVal, SubSecondPrecision: origPrecisionVal, Nullability: origNullabilityVal}
```
zeroshade, line 34: expand this into a multi-line literal to make it easier to read please?

zeroshade (approved):

expr/interval_compound_literal.go
anshuldata: yes, not useful now. removed them

expr/interval_compound_literal_test.go
anshuldata: since remove With* Methods so this test case isn't needed. Removed this test case

anshuldata:
can you merge this @zeroshade, I don't have write access

PR #48, opened 2024-08-19
Update proto definition to v0.55.0

anshuldata:
* Fixed name of Join type which was breaking change. Updated in [This](https://github.com/substrait-io/substrait/commit/bed84ecb6193c22bf2ff83dc3a391ec5a9a3aa68) PR updated these types PR


plan/relations.go
```diff
@@ -607,14 +607,14 @@ func (j *JoinRel) RecordType() types.StructType {
 	switch j.joinType {
```
EpsilonPrime, line 607: Are you going to need JoinTypeRightSingle  and JoinTypeRightAnti for all of the options to be handled without a default clause?

anshuldata: This PR I intended to keep for proto change so I planned to handle them in a separate PR. I updated PR to add missing ones to indicate unsupported. Kindly check

jacques-n (approved):

zeroshade (approved):
Is there something that needs to be fixed in the CI?

PR #52, opened 2024-08-30
Support for Parameterized type

anshuldata:
Enhancements:

1. Added interface "FuncDefArgType". This represents argument of a function in substrait extension

2. Function implementations input argument are feched from parser via method ArgType method. ArgType returns interface "FuncDefArgType". Added another method to parser expression "RetType" to return Type (concrete) as return type. This is temporary and should be fixed with Type Derivation. To keep behavior same for return type I kept "RetType" implementations same to return "Type"

3.  Added support for parameterized type
      * ParameterizedFixedChar/VarChar/FixedBinary/Decimal
      * Parameterized/PrecisionTimestamp/PrecisionTimestampTz
      * Parameterized/Decimal/List/Struct/Map Type
      
     
4. Added interface "LeafParameter".  This indicates parameter which are not nested 
        * These can be of two type, concrete and abstract (name in code is "ConcreteIntParam" and "VariableIntParam")
       * This allows representation of a leaf parameter. This also helps in evaluating if a given parameter is leaf or not (a parameter can be FuncDefArgType too)

5. Renamed "ParameterizedType" to "CompositeType". This was representing concrete serializable  type so changed name to reflect so
6. Fixed string representation of type to be uniform for fields having more than one type (map, struct) to have space between different parameters
7. Enhanced existing PrecisionTimestampType, PrecisionTimestampTzType, StructType, ListType, MapType to indicate they are concrete parameterized type (added BaseString and ParameterString) method
8. Separate "AnyType" from existing nonParamType. This will be helpful in match method in follow up PR to match argument against parameter
9. Added API "HasSyncParams". This takes []FuncDefArgType as argument and returns true if any abstract leaf parameter is shared across different arguments. This will be used in Match API in follow up PR to act on such FunctionVariants


types/parameterized_decimal_type.go
```diff
@@ -0,0 +1,55 @@
+package types
+
+import (
+	"fmt"
+
+	"github.com/substrait-io/substrait-go/proto"
+)
+
+type ParameterizedDecimalType struct {
```
zeroshade, line 9: docstring?

types/any_type.go
```diff
@@ -0,0 +1,58 @@
...
+// This type is not used in function invocation. It is only used in function definition
+type AnyType struct {
+	Name        string
+	Nullability Nullability
+}
+
+func (*AnyType) isRootRef() {}
+func (m *AnyType) WithNullability(nullability Nullability) Type {
+	m.Nullability = nullability
+	return m
```
zeroshade, line 19: Shouldn't this return a copy with the changed nullability rather than change the one being used? It feels like this should be a value receiver, not a pointer receiver

types/parameterized_decimal_type.go
```diff
@@ -0,0 +1,55 @@
...
+	Nullability      Nullability
+	TypeVariationRef uint32
+	Precision        IntegerParam
+	Scale            IntegerParam
+}
+
+func (*ParameterizedDecimalType) isRootRef() {}
+func (m *ParameterizedDecimalType) WithNullability(n Nullability) Type {
+	m.Nullability = n
+	return m
```
zeroshade, line 19: same as previous comment, this should probably be a value receiver, not a pointer receiver

types/parameterized_decimal_type.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func (*ParameterizedDecimalType) ToProtoFuncArg() *proto.FunctionArgument {
+	// parameterized type are never on wire so to proto is not supported
+	panic("not supported")
+}
+
+func (m *ParameterizedDecimalType) ShortString() string {
+	t := &DecimalType{}
+	return t.ShortString()
```
zeroshade, line 41: this doesn't need to be a pointer. Could probably just be 

```go
var t DecimalType
return t.ShortString()
```

types/parameterized_types.go
```diff
@@ -0,0 +1,109 @@
+package types
+
+import (
+	"fmt"
+
+	"github.com/substrait-io/substrait-go/proto"
+)
+
+type IntegerParam struct {
```
zeroshade, line 9: docstring

types/parameterized_types.go
```diff
@@ -0,0 +1,109 @@
...
+type ParameterizedTypeSingleIntegerParam[T VarCharType | FixedCharType | FixedBinaryType | PrecisionTimestampType | PrecisionTimestampTzType] struct {
+	Nullability      Nullability
+	TypeVariationRef uint32
+	IntegerOption    IntegerParam
+}
+
+func (m *ParameterizedTypeSingleIntegerParam[T]) WithIntegerOption(integerOption IntegerParam) ParameterizedSingleIntegerType {
+	m.IntegerOption = integerOption
+	return m
+}
```
zeroshade, line 34: same as before, shouldn't this be a value receiver and not change the param but return a copy instead?

types/parameterized_types.go
```diff
@@ -0,0 +1,109 @@
...
+		return t.ShortString()
+	case *ParameterizedPrecisionTimestampType:
+		t := &PrecisionTimestampType{}
+		return t.ShortString()
+	case *ParameterizedPrecisionTimestampTzType:
+		t := &PrecisionTimestampTzType{}
+		return t.ShortString()
+	default:
+		panic("unknown type")
+	}
```
zeroshade, line 78: Same comment as before, it doesn't need to be pointers you can just use the type without needing the `&`

zeroshade (changes requested):

types/parameterized_decimal_type.go
anshuldata: fixed it

types/any_type.go
anshuldata: fixed it

types/parameterized_decimal_type.go
anshuldata: fixed it

types/parameterized_decimal_type.go
anshuldata: fixed it

types/parameterized_types.go
anshuldata: fixed it

types/parameterized_types.go
anshuldata: fixed it

types/any_type.go
```diff
@@ -0,0 +1,58 @@
...
+	return true
+}
+
+func (AnyType) ToProtoFuncArg() *proto.FunctionArgument {
+	panic("not allowed")
+}
+
+func (AnyType) ToProto() *proto.Type {
+	panic("not allowed")
+}
```
zeroshade, line 39: It feels like these should be removed from this type entirely and we should perhaps split the Type interface to create an interface that contains *just* the `ToProtoFuncArg` and `ToProto` methods, and then just use type assertion to do the serialization, enabling proper error reporting if the type doesn't meet the interface (rather than us just panic'ing if they use an AnyType). What do you think?

types/any_type.go
```diff
@@ -0,0 +1,58 @@
...
+func (t AnyType) ShortString() string { return t.Name }
+func (t AnyType) String() string {
+	return fmt.Sprintf("%s%s", t.Name, strNullable(t))
+}
+
+// Below methods are for parser Def interface
+
+func (AnyType) Optional() bool {
+	panic("not allowed")
+}
```
zeroshade, line 50: We have a nullability member, why is `Optional` not allowed?

types/parameterized_types.go
```diff
@@ -0,0 +1,112 @@
...
+	Name string
+}
+
+func (m IntegerParam) Equals(o IntegerParam) bool {
+	return m == o
+}
+
+func (p IntegerParam) ToProto() *proto.ParameterizedType_IntegerParameter {
+	panic("not implemented")
+}
```
zeroshade, line 21: same comment as before

types/parameterized_types.go
```diff
@@ -0,0 +1,112 @@
...
+	if o, ok := rhs.(ParameterizedTypeSingleIntegerParam[T]); ok {
+		return o == m
+	}
+	return false
+}
+
+func (ParameterizedTypeSingleIntegerParam[T]) ToProtoFuncArg() *proto.FunctionArgument {
+	// parameterized type are never on wire so to proto is not supported
+	panic("not supported")
+}
```
zeroshade, line 60: same comment

types/any_type.go
anshuldata: agree, it is much cleaner this way. I was initially contemplating to do it as part of this PR or separate PR. I now fixed as part of this PR. Basically introduced interface "FuncArgType" which represents a type which can be function argument (basically can be on wire)

types/parameterized_decimal_type.go
```diff
@@ -0,0 +1,58 @@
...
+	if o, ok := rhs.(ParameterizedDecimalType); ok {
+		return o == m
+	}
+	return false
+}
+
+func (ParameterizedDecimalType) ToProtoFuncArg() *proto.FunctionArgument {
+	// parameterized type are never on wire so to proto is not supported
+	panic("not supported")
+}
```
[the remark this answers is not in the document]
anshuldata: removed it since now Type interface doesn't have proto methods

types/parameterized_types.go
anshuldata: removed it since now Type interface doesn't have proto methods

types/parameterized_types.go
anshuldata: removed it since now Type interface doesn't have proto methods

types/parameterized_types.go
```diff
@@ -0,0 +1,101 @@
+package types
+
+import (
+	"fmt"
+)
+
+// IntegerParam represents a single integer parameter for a parameterized type
+// Example: VARCHAR(L1) -> L1 is the integer parameter
+type IntegerParam struct {
```
jacques-n, line 9: why is this a struct as opposed to a type alias of string?

types/parser/type_parser.go
```diff
@@ -3,6 +3,7 @@
 package parser
```
jacques-n, line 3: I generally think we're just creating more debt by maintaining multiple parsers in the project rather than just using antlr everywhere.

types/parameterized_types.go
```diff
@@ -0,0 +1,101 @@
...
+func (m IntegerParam) Equals(o IntegerParam) bool {
+	return m == o
+}
+
+func (m IntegerParam) String() string {
+	return m.Name
+}
+
+// ParameterizedTypeSingleIntegerParam This is a generic type to represent parameterized type with a single integer parameter
+type ParameterizedTypeSingleIntegerParam[T VarCharType | FixedCharType | FixedBinaryType | PrecisionTimestampType | PrecisionTimestampTzType] struct {
```
jacques-n, line 22: why export?

extensions/simple_extension.go
```diff
@@ -57,6 +59,7 @@ type TypeVariation struct {
 
 type Argument interface {
 	toTypeString() string
+	ArgType() (types.Type, error)
```
jacques-n, line 62: I don't get this method on this interface. What's the point? It seems like you've replaced type assertions with custom method that returns errors. How's that better than doing a type assertion on ValueArg and then having ValueArg have a utility method? 

As part of this, I suggest we use unexported marker methods on Argument so it is clear that people can't extend it. There are only three types of arguments: ValuesArg, TypeArg and EnumArg.

jacques-n (commented):
I'm struggling with how this pattern extends to things like List<ANY>, LIST<T1>, STRUCT<T1, T2>, etc.

types/parameterized_types.go
anshuldata: fixed it

types/parser/type_parser.go
anshuldata: We will fix it in follow up PR to rely on antlr

types/parameterized_types.go
anshuldata: fixed it

extensions/simple_extension.go
anshuldata: Removed this method and added marker

anshuldata:
@jacques-n @zeroshade 
Kindly re-review. I have overhauled this PR to support
* nested parameterized type (struct, map, list)
* support for parameterized decimal type where one of precision and scale may be concrete

I have captured changes in PR description. Kindly read it first to get an high level idea of changes.

Thanks

types/parameter_types/abstract_parameter_type.go
```diff
@@ -0,0 +1,23 @@
+package parameter_types
+
+// LeafIntParamAbstractType represents an integer parameter for a parameterized type
+// Example: VARCHAR(L1) -> L1 is an LeafIntParamAbstractType
+// DECIMAL<P, 0> --> P Is an LeafIntParamAbstractType
+type LeafIntParamAbstractType string
```
zeroshade, line 6: is there a corresponding non-`Leaf` param type? If not, do we really need to call it a "Leaf" type?

Is `AbstractType` what it is called in the substrait spec? Would `IntParamVar` be a better name if not?

types/parameter_types/concrete_parameter_type.go
```diff
@@ -0,0 +1,23 @@
...
+		return t == m
+	}
+	return false
+}
+
+func (m LeafIntParamConcreteType) String() string {
+	return fmt.Sprintf("%d", m)
+}
+
+func (m LeafIntParamConcreteType) ToProtoVal() int32 {
```
zeroshade, line 21: does the Abstract type version need a `ToProtoVal` method also?

types/parameter_types/leaf_parameter_type.go
```diff
@@ -0,0 +1,18 @@
+package parameter_types
+
+// LeafParameter represents a parameter type
+// parameter can of concrete (38) or abstract type (P)
+// or another parameterized type like VARCHAR<"L1">
+type LeafParameter interface {
+	// IsCompatible is type compatible with other
+	// compatible is other can be used in place of this type
+	IsCompatible(other LeafParameter) bool
+	String() string
```
zeroshade, line 10: `fmt.Stringer` ?

types/parameter_types/concrete_parameter_type.go
```diff
@@ -0,0 +1,23 @@
+package parameter_types
+
+import "fmt"
+
+// LeafIntParamConcreteType represents a single integer concrete parameter for a concrete type
+// Example: VARCHAR(6) -> 6 is an LeafIntParamConcreteType
+// DECIMAL<P, 0> --> 0 Is an LeafIntParamConcreteType but P not
+type LeafIntParamConcreteType int32
```
zeroshade, line 8: `ConcreteIntParam` ?

types/parameter_types/leaf_parameter_type.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+// AbstractParameterType represents a parameter type which is abstract.
+// it can be a leaf parameter (LeafIntParamAbstractType)
+// or another abstract type like "DECIMAL<P, S>"
+type AbstractParameterType interface {
```
zeroshade, line 16: `AbstractParameter`? I don't think `Type` needs to be part of the name here

functions/types.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	substraitgo "github.com/substrait-io/substrait-go"
 	"github.com/substrait-io/substrait-go/types"
+	"github.com/substrait-io/substrait-go/types/parameter_types"
```
zeroshade, line 9: We should add the license identifier `// SPDX-License-Identifier: Apache-2.0` to this file

types/any_type.go
```diff
@@ -0,0 +1,35 @@
+package types
```
zeroshade, line 1: Should add license identifier `// SPDX-License-Identifier: Apache-2.0` to the file

types/any_type_test.go
```diff
@@ -0,0 +1,36 @@
+package types_test
```
zeroshade, line 1: license identifier header, etc...

types/parameter_types/abstract_parameter_type.go
```diff
@@ -0,0 +1,23 @@
+package parameter_types
```
zeroshade, line 1: since this is already in the `types` package, could we shorten this package name to just `parameters` ?

Also need to add the license identifier header

types/parameterized_decimal_type.go
zeroshade: this got converted back to a pointer receiver, please change it to a value receiver?

types/parameterized_list_type.go
```diff
@@ -0,0 +1,59 @@
...
+	Nullability      Nullability
+	TypeVariationRef uint32
+	Type             ParameterizedAbstractType
+}
+
+func (*ParameterizedListType) isRootRef() {}
+func (m *ParameterizedListType) WithNullability(n Nullability) Type {
+	m.Nullability = n
+	return m
+}
```
zeroshade, line 23: value receiver

types/parameterized_map_type.go
```diff
@@ -0,0 +1,65 @@
...
+	TypeVariationRef uint32
+	Key              Type
+	Value            Type
+}
+
+func (*ParameterizedMapType) isRootRef() {}
+func (m *ParameterizedMapType) WithNullability(n Nullability) Type {
+	m.Nullability = n
+	return m
+}
```
zeroshade, line 22: value receiver

types/parameterized_single_integer_param_type.go
```diff
@@ -0,0 +1,101 @@
+package types
+
+import (
+	"fmt"
+
+	"github.com/substrait-io/substrait-go/types/parameter_types"
+)
+
+// parameterizedTypeSingleIntegerParam This is a generic type to represent parameterized type with a single integer parameter
+type parameterizedTypeSingleIntegerParam[T VarCharType | FixedCharType | FixedBinaryType | PrecisionTimestampType | PrecisionTimestampTzType] struct {
```
zeroshade, line 10: we should shift these to a constraint to shorten this line and make it easier for reuse. It also would allow us to add the common functions to the interface constraint and make the subsequent code simpler.

```go
type singleIntParamType interface {
        VarCharType | FixedCharType | FixedBinaryType |
                PrecisionTimestampType | PrecisionTimestampTzType

         ShortString() string
         BaseString() string
}

type parameterizedTypeSingleIntegerParam[T singleIntParamType] struct {
```

types/parameterized_single_integer_param_type.go
```diff
@@ -0,0 +1,101 @@
...
+		return t.ShortString()
+	case ParameterizedPrecisionTimestampType:
+		t := PrecisionTimestampType{}
+		return t.ShortString()
+	case ParameterizedPrecisionTimestampTzType:
+		t := PrecisionTimestampTzType{}
+		return t.ShortString()
+	default:
+		panic("unknown type")
+	}
```
zeroshade, line 58: if you follow the suggestion above to add `ShortString() string` and `BaseString() string` to the constraint, then you can rewrite this function as such:

```go
var z T
return z.ShortString()
```

the `z` is for "zero value".

types/parameterized_single_integer_param_type.go
```diff
@@ -0,0 +1,101 @@
...
+		return t.BaseString()
+	case ParameterizedPrecisionTimestampType:
+		t := PrecisionTimestampType{}
+		return t.BaseString()
+	case ParameterizedPrecisionTimestampTzType:
+		t := PrecisionTimestampTzType{}
+		return t.BaseString()
+	default:
+		panic("unknown type")
+	}
```
zeroshade, line 88: if you follow the suggestion above, this entire function can be rewritten to:

```go
var z T
return z.BaseString()
```

types/parameterized_struct_type.go
```diff
@@ -0,0 +1,80 @@
...
+	Nullability      Nullability
+	TypeVariationRef uint32
+	Type             []Type
+}
+
+func (*ParameterizedStructType) isRootRef() {}
+func (m *ParameterizedStructType) WithNullability(n Nullability) Type {
+	m.Nullability = n
+	return m
+}
```
zeroshade, line 23: value receiver

types/parameterized_struct_type.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+		}
+		return true
+	}
+	return false
+}
+
+func (m *ParameterizedStructType) ShortString() string {
+	t := StructType{}
+	return t.ShortString()
```
zeroshade, line 48: Maybe use

```go
return (StructType{}).ShortString()
```

or 

```go
var t StructType
return t.ShortString()
```

types/parameterized_struct_type.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// ParameterizedStructType is a struct having at least one parameter of type ParameterizedAbstractType
+// example: Struct<Decimal(P,S), INT8>.
+// If All arguments are concrete they are represented by StructType
+type ParameterizedStructType struct {
+	Nullability      Nullability
+	TypeVariationRef uint32
+	Type             []Type
+}
```
zeroshade, line 17: since this is exactly the same definition as `StructType` maybe we just do

```go
type ParameterizedStructType struct {
        StructType
}
```

and just implement any functions that are different. or 

```go
type ParameterizedStructType StructType
```

either way that'll let us avoid having to manually create `StructType` instances in the methods.

types/parser/type_parser.go
zeroshade: Is there an antlr implementation for Go that we can use which is easy to utilize and plug into the types?

zeroshade (changes requested):
Made a new pass on this, left a bunch of comments

types/parser/type_parser.go
jacques-n: Antlr has generators for most languages, including go. Have used it extensively elsewhere and much more reliable/easy to maintain than writing parsers in multiple languages.

jacques-n (changes requested):
I think the overloading of Type here is wrong. You're upgrading Type to mean both concrete or parameterized. I think that has undesirable side effects and should be avoided.

A better way to approach is to keep Type to mean concrete types only. That's what users are going to be using in most of the substrait-go. Let's keep that narrow. 

We then have ParameterizedType. It could also be called FunctionDefinitionArgumentType since that it is the only place this concept needs to exist. This is a type that can be used in the definition of function arguments and can be parameterized or concrete. FunctionDefinitionArgumentType can't be used in plans and doesn't need to be converted to Protobuf binary values. In the protobuf files, this type is called [ParameterizedType](https://github.com/substrait-io/substrait/blob/main/proto/substrait/parameterized_types.proto#L13C9-L13C26). Note that Parameterized types should not be confused with [Compound Types](https://substrait.io/types/type_classes/#compound-types) which are types that are concrete types that have multiple components.

The third thing is what we call type expressions. Each function variant should have a type expression. This is mostly missing from substrait-go. That is the expression that is handed the concrete types and works with the FunctionDefinitionArgumentType values to figure things out.

Please update so that Type means what it means today and let's upgrade FunctionVariant to be defined using a new ParametertizedType or FunctionDefintitionArugmentType (or another name if you prefer).

A good test is all existing substrait-go type construction/manipulation code (outside of function variant construction) should have no changes. For example, [this shouldn't be needed](https://github.com/substrait-io/substrait-go/pull/52/files/a6ef7c8e3da9c1a8e9848a046f564b1c73f53ea1).

types/parameterized_decimal_type.go
anshuldata: I deliberately converted it to pointer receiver. This is to catch error of using wrong type at compile time instead of run time.
Basically, if a interface is implemented with a pointer receiver. Value receiver won't implement this. 
Since pointer of ParameterizedDecimalType implements "FuncDefArgType" so value receiver won't satisfy interface "FuncDefArgType"
I feel this is helpful if we have "FuncDefArgType" and need to switch case on "type". We can get compile error than
(sample program is [here](https://go.dev/play/p/-N-XAX8DrVm))

Let me know if this sounds ok

types/parameterized_list_type.go
anshuldata: I deliberately converted it to pointer receiver. This is to catch error of using wrong type at compile time instead of run time.
Basically, if a interface is implemented with a pointer receiver. Value receiver won't implement this.
Since pointer of ParameterizedListType implements "FuncDefArgType" so value receiver won't satisfy interface "FuncDefArgType"
I feel this is helpful if we have "FuncDefArgType" and need to switch case on "type". We can get compile error than
(sample program is [here](https://go.dev/play/p/-N-XAX8DrVm))

Let me know if this sounds ok

types/parameter_types/abstract_parameter_type.go
anshuldata: renamed it to ConcreteIntParam

types/parameter_types/leaf_parameter_type.go
anshuldata: fixed it

types/parameter_types/concrete_parameter_type.go
anshuldata: renamed it to ConcreteIntParam

types/parameter_types/concrete_parameter_type.go
anshuldata: with new change (separation of Type and ParameterizedType (called FuncDefArgType) interface) this proto method is not needed. 
FuncDefArgType implementations will never be on wire so proto method not needed

types/parameter_types/leaf_parameter_type.go
anshuldata: This is not used anymore

functions/types.go
anshuldata: added, Thanks

types/any_type.go
anshuldata: added, Thanks

types/any_type_test.go
anshuldata: added, Thanks

types/parameter_types/abstract_parameter_type.go
anshuldata: added license
renamed this package to "leaf_parameters" since it now only contains leaf parameters

types/parameterized_map_type.go
anshuldata: same reasoning as previous

types/parameterized_single_integer_param_type.go
anshuldata: This comment is pending, I will address in follow up change (either in next iteration of review comments or next PR whichever comes first).

types/parameterized_single_integer_param_type.go
anshuldata: This comment is pending, I will address in follow up change (either in next iteration of review comments or next PR whichever comes first).

types/parameterized_struct_type.go
anshuldata: As mentioned above this is deliberate change to catch errors at compile time

types/parameterized_struct_type.go
anshuldata: this method is no more needed in FuncDefArgType

types/parameterized_struct_type.go
anshuldata: It was a mistake to use []Type as Type. It is []FuncDefArgType. Fixed it in updated PR to use FuncDefArgType

types/parameterized_struct_type.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/substrait-io/substrait-go/types/parameter_types"
+)
+
+// ParameterizedStructType is a struct having at least one parameter of type ParameterizedAbstractType
+// example: Struct<Decimal(P,S), INT8>.
+// If All arguments are concrete they are represented by StructType
+type ParameterizedStructType struct {
+	Nullability      Nullability
+	TypeVariationRef uint32
+	Type             []Type
```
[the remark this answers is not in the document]
anshuldata: yes, it was a mistake. Fixed it to use "FuncDefArgType"

anshuldata:
> Please update so that Type means what it means today and let's upgrade FunctionVariant to be defined using a new ParametertizedType or FunctionDefintitionArugmentType (or another name if you prefer).

Updated PR to use interface "FuncDefArgType" for function parameters. Kindly check @jacques-n

extensions/simple_extension.go
```diff
@@ -69,6 +70,8 @@ func (EnumArg) toTypeString() string {
 	return "req"
 }
 
+func (v EnumArg) marker() {}
```
jacques-n, line 73: marker method should be named for the kind of mark. E.g. argumentMarker() here.

types/parameterized_single_integer_param_type.go
```diff
@@ -0,0 +1,63 @@
...
+package types
+
+import (
+	"fmt"
+
+	"github.com/substrait-io/substrait-go/types/leaf_parameters"
+)
+
+// parameterizedTypeSingleIntegerParam This is a generic type to represent parameterized type with a single integer parameter
+type parameterizedTypeSingleIntegerParam[T VarCharType | FixedCharType | FixedBinaryType | PrecisionTimestampType | PrecisionTimestampTzType] struct {
```
jacques-n, line 12: I feel like the constraint is kind of pointless since the type is not exported.

types/leaf_parameters/leaf_parameter_type.go
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package leaf_parameters
+
+import "fmt"
+
+// LeafParameter represents a parameter type
+// parameter can of concrete (38) or abstract type (P)
+// or another parameterized type like VARCHAR<"L1">
+type LeafParameter interface {
```
jacques-n, line 10: Why not call this IntegerParameter? Or is this intended to cover type parameters as well? E.g. `STRUCT<T,X>`. 

If it covers type parameters as well, why would it be called leaf? It is only integers, why not use IntegerParameter instead of Leaf?

types/types.go
```diff
@@ -369,14 +371,24 @@ type (
...
 		ParameterString() string
 		BaseString() string
 	}
 
+	// FuncDefArgType this represents a type used in function argument
+	// These type can't be present in plan (not serializable)
+	FuncDefArgType interface {
+		fmt.Stringer
+		SetNullability(Nullability) FuncDefArgType
+		HasParameterizedParam() bool
```
jacques-n, line 386: let's add doc to these methods.

types/types.go
```diff
@@ -369,14 +371,24 @@ type (
 		WithNullability(Nullability) Type
 	}
 
-	ParameterizedType interface {
+	// CompositeType this represents a concrete type having components
+	CompositeType interface {
 		Type
 		ParameterString() string
```
jacques-n, line 377: add docs to these methods

jacques-n: also, how important is it that these methods are exported?

types/types.go
```diff
@@ -369,14 +371,24 @@ type (
...
 		BaseString() string
 	}
 
+	// FuncDefArgType this represents a type used in function argument
+	// These type can't be present in plan (not serializable)
+	FuncDefArgType interface {
```
jacques-n, line 383: this is much cleaner, thanks

jacques-n (commented):
a few small comments but looks pretty good from my pov.

extensions/simple_extension.go
anshuldata: fixed it

types/parameterized_single_integer_param_type.go
anshuldata: yes, not really required to strict export. so removed types as constraint but added interface as constraint to simplify baseString as mentioned in [this](https://github.com/substrait-io/substrait-go/pull/52#discussion_r1752303432) comment

types/types.go
anshuldata: added

types/types.go
anshuldata: added doc

these are used in function registry which is in separate package. Basically, to convert substrait type to local name and to initialize local type registry.

types/parameterized_single_integer_param_type.go
anshuldata: fixed this

types/parameterized_single_integer_param_type.go
anshuldata: fixed it

types/parameterized_single_integer_param_type.go
anshuldata: ShortString method is no more part of interface implemented by this type so this is not relevant anymore. resolving this comment

jacques-n (approved):
LGTM. Thanks for the iterations!

anshuldata:
Kindly check updated PR @zeroshade.
If it gets merged I would like to rebase #53 with this PRs changes

Thanks

types/parameterized_decimal_type.go
zeroshade: The problem is that now, unlike with other types, calling `WithNullability` actually *changes* `ParameterizedDecimalType` as opposed to nearly all other cases where calling that just returns a copy with the nullability updated.

This inconsistency will be a problem for users

PR #53, opened 2024-09-02
Add Match method to FuncDefArgType interface

anshuldata:
* Added MatchWithNullability and MatchWithoutNullability methods on FuncDefArgType interface. This method returns true if argument "t" is type compatible with this type otherwise it returns false
* These two APIs are necessary considering Match behavior will differ based on function's nullability handling (MIRROR, DECLARED_OUTPUT, DISCRETE).
* Nullability "MIRROR" and "DECLARED_OUTPUT" requires MatchWithoutNullability
* Nullability "DISCRETE" will require MatchWithNullability
* Also updated go version to 1.22 to fix golint error with reflect API TypeFor
* In follow up PR I will add Match method on FunctionVariant interface

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -394,11 +394,26 @@ type (
 		// GetParameterizedParams returns all parameterized parameters
 		// it doesn't return concrete parameters
 		GetParameterizedParams() []interface{}
+
+		// MatchWithNullability This API return true if Type argument
+		// is compatible with this param otherwise it returns false.
+		// This method expects that nullability of argument is same as this type.
+		MatchWithNullability(ot Type) bool
+		// MatchWithoutNullability This API return true if Type argument
+		// is compatible with this param otherwise it returns false.
+		// This method ignores nullability for matching.
+		MatchWithoutNullability(ot Type) bool
 	}
 
 	FixedType interface {
 		CompositeType
 		WithLength(int32) FixedType
+		GetLength() int32
+	}
+
+	timestampPrecisionType interface {
+		CompositeType
+		GetPrecision() TimePrecision
 	}
 )
 
@@ -644,6 +659,20 @@ func (s *PrimitiveType[T]) SetNullability(n Nullability) FuncDefArgType {
 	return s
 }
 
+func (s *PrimitiveType[T]) MatchWithNullability(ot Type) bool {
+	if s.Nullability != ot.GetNullability() {
+		return false
+	}
+	return s.MatchWithoutNullability(ot)
+}
+
+func (s *PrimitiveType[T]) MatchWithoutNullability(ot Type) bool {
+	if _, ok := ot.(*PrimitiveType[T]); ok {
+		return true
+	}
+	return false
+}
+
 // create type aliases to the generic structs
 type (
 	BooleanType                           = PrimitiveType[bool]
@@ -728,6 +757,10 @@ func (s *FixedLenType[T]) WithLength(length int32) FixedType {
 	out := *s
 	out.Length = length
 	return &out
+}
+
+func (s *FixedLenType[T]) GetLength() int32 {
+	return s.Length
 }
 
 // DecimalType is a decimal type with concrete precision and scale parameters, e.g. Decimal(10, 2).
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@
 
 module github.com/substrait-io/substrait-go
 
-go 1.21
+go 1.22
 
 require (
 	github.com/alecthomas/participle/v2 v2.0.0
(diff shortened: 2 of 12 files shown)


types/parameterized_types.go
```diff
@@ -89,6 +89,38 @@ func (m ParameterizedTypeSingleIntegerParam[T]) ParameterString() string {
...
+	case PrecisionTimestampType:
+		if _, ok := ot.(*PrecisionTimestampType); ok {
+			return true
+		}
+	case PrecisionTimestampTzType:
+		if _, ok := ot.(*PrecisionTimestampTzType); ok {
+			return true
+		}
+	default:
+		return false
```
zeroshade, line 119: It feels like we could achieve this in a simpler way by using reflect? 

```go
return reflect.TypeFor[T]() == reflect.ValueOf(ot).Elem().Type()
```

types/parameterized_types.go
```diff
@@ -89,6 +89,38 @@ func (m ParameterizedTypeSingleIntegerParam[T]) ParameterString() string {
...
+	}
+	return false
```
zeroshade, line 121: this is entirely unnecessary since we have the `default` case already, remove this

types/types.go
```diff
@@ -369,6 +369,7 @@ type (
 		// WithNullability returns a copy of this type but with
 		// the nullability set to the passed in value
 		WithNullability(Nullability) Type
+		Match(ot Type) bool
```
zeroshade, line 372: how is Match different from Equals? It looks like most of the implementations make it identical?

If `Match` is different from `Equals` then we should add a comment explaining why you would use one over the other.

types/interval_compound_type.go
```diff
@@ -71,3 +71,7 @@ func (m IntervalCompoundType) String() string {
 	return fmt.Sprintf("intervalcompound%s<%d>", strNullable(m),
 		m.precision.ToProtoVal())
 }
+
+func (m IntervalCompoundType) Match(ot Type) bool {
+	return m == ot
```
zeroshade, line 76: delegate to `m.Equals(ot)`?

types/interval_year_month_type.go
```diff
@@ -58,3 +58,7 @@ func (IntervalYearToMonthType) ShortString() string { return "intrytm" }
 func (m IntervalYearToMonthType) String() string {
 	return fmt.Sprintf("intervalyeartomonth%s", strNullable(m))
 }
+
+func (m IntervalYearToMonthType) Match(ot Type) bool {
+	return m == ot
```
zeroshade, line 63: delegate to `m.Equals(ot)`?

types/types.go
```diff
@@ -823,6 +845,27 @@ func (t *StructType) String() string {
...
+	if other, ok := ot.(*StructType); ok {
+		switch {
+		case t.Nullability != other.Nullability:
+			return false
+		case t.TypeVariationRef != other.TypeVariationRef:
+			return false
+		}
+		if len(t.Types) != len(other.Types) {
+			return false
+		}
```
zeroshade, line 858: I would add this to the cases above instead of having it be a separate `if` block

zeroshade (changes requested):

types/parameterized_types.go
anshuldata: fixed it, Thanks

types/types.go
anshuldata: I fixed PR to have "MatchWithNullability" and "MatchWithoutNullability" methods instead of "Match" method.
I missed it earlier. Based on Option Nullability on function, either MatchWithNullability or MatchWithoutNullability will be called.
So considering this, Match* and equal won't be same

types/interval_compound_type.go
anshuldata: not needed since API changed to "MatchWithNullability" and "MatchWithoutNullability"

types/interval_year_month_type.go
anshuldata: not needed since API changed to "MatchWithNullability" and "MatchWithoutNullability"

types/types.go
anshuldata: with updated PR this won't be needed. So resolving it

types/parameterized_types.go
anshuldata: with updated change, this won't be needed. Thanks

zeroshade:
Using `reflect.TypeFor` requires updating the module to `go1.22` instead of `go1.21`, that's why the linter is failing most likely

anshuldata:
I will un-draft once #52 is merged and is rebased for this PR

jacques-n (approved):
Feels like it is pretty clean and simple now and all outstanding requested changes have been dealt with.

The biggest controversial thing is upgrading to go 1.22. Looked at only 1.22 and 1.23 are now supported by community so that seems fine.

PR #54, opened 2024-09-04
Add Match and MatchAt API to FunctionVariant interface

anshuldata:
* fixed default nullability for ScalarFunctionImpl if nullability is not provided

### DIFF
--- a/extensions/variants.go
+++ b/extensions/variants.go
@@ -21,6 +21,19 @@ type FunctionVariant interface {
 	URI() string
 	ResolveType(argTypes []types.Type) (types.Type, error)
 	Variadic() *VariadicBehavior
+	// Match this function matches input arguments against this functions parameter list
+	// returns (true, nil) if all input argument can type replace the function definition argument
+	// returns (false, err) for invalid input argument. For e.g. if input argument nullability is not correctly
+	// set this function will return error
+	// returns (false, nil) valid input arguments and argument list type replace parameter list
+	Match(argumentTypes []types.Type) (bool, error)
+	// MatchAt this function matches input argument at position against definition of this
+	// functions argument at same position
+	// returns (true, nil) if input argument can type replace the function definition argument
+	// returns (false, err) for invalid input argument. For e.g. if input arg position is negative or
+	// argument nullability is not correctly set this function will return error
+	// returns (false, nil) valid input argument type and argument can't type replace parameter at argPos
+	MatchAt(typ types.Type, pos int) (bool, error)
 }
 
 func EvaluateTypeExpression(nullHandling NullabilityHandling, expr parser.TypeExpression, paramTypeList ArgumentList, actualTypes []types.Type) (types.Type, error) {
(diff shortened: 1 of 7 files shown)


extensions/variants.go
```diff
@@ -83,6 +86,43 @@ func EvaluateTypeExpression(nullHandling NullabilityHandling, expr parser.TypeEx
...
+	}
+	return true, nil
+}
+
+func matchArgumentAt(actualType types.Type, argPos int, nullability NullabilityHandling, paramTypeList ArgumentList) (bool, error) {
+	if argPos < 0 {
+		return false, errors.New("non-zero argument position")
+	}
+	if argPos >= len(paramTypeList) {
+		return false, fmt.Errorf("%w: argument position %d out of range", substraitgo.ErrNotFound, argPos)
```
jacques-n, line 110: this should work with variadic functions. I think this failure is incorrect in that case.

extensions/variants.go
```diff
@@ -140,24 +180,35 @@ type ScalarFunctionVariant struct {
...
-func (s *ScalarFunctionVariant) URI() string                      { return s.uri }
+func (s *ScalarFunctionVariant) Name() string                { return s.name }
+func (s *ScalarFunctionVariant) Description() string         { return s.description }
+func (s *ScalarFunctionVariant) Args() ArgumentList          { return s.impl.Args }
+func (s *ScalarFunctionVariant) Options() map[string]Option  { return s.impl.Options }
+func (s *ScalarFunctionVariant) Variadic() *VariadicBehavior { return s.impl.Variadic }
+func (s *ScalarFunctionVariant) Deterministic() bool         { return s.impl.Deterministic }
+func (s *ScalarFunctionVariant) SessionDependent() bool      { return s.impl.SessionDependent }
+func (s *ScalarFunctionVariant) Nullability() NullabilityHandling {
+	if s.impl.Nullability == "" {
```
jacques-n, line 191: we shouldn't need to handle this here. we should figure out a way to handle on unmarshal. I think something like github.com/creasty/defaults will allow us to use tags to handle defaults rather than having to hand code fixes. (Not sure if that is the best library, just an option).

extensions/variants.go
```diff
@@ -140,24 +180,35 @@ type ScalarFunctionVariant struct {
...
-	return EvaluateTypeExpression(s.impl.Nullability, s.impl.Return, s.impl.Args, argumentTypes)
+	return EvaluateTypeExpression(s.Nullability(), s.impl.Return, s.impl.Args, argumentTypes)
 }
 func (s *ScalarFunctionVariant) CompoundName() string {
 	return s.name + ":" + s.impl.signatureKey()
 }
 func (s *ScalarFunctionVariant) ID() ID {
 	return ID{URI: s.uri, Name: s.CompoundName()}
 }
+func (s *ScalarFunctionVariant) Match(argumentTypes []types.Type) (bool, error) {
```
jacques-n, line 206: In doc it would be good to give information on what kinds of errors could be return here. It's not immediately evident.

extensions/variants.go
```diff
@@ -140,24 +180,35 @@ type ScalarFunctionVariant struct {
...
+	return matchArguments(s.Nullability(), s.impl.Args, argumentTypes)
```
jacques-n, line 207: maybe add a todo here or somewhere that returns error if the function requires argument coordination. For example `append(list LIST<T1>, item T1)`

functions/dialect_test.go
```diff
@@ -564,3 +566,123 @@ func checkCompoundNames(t *testing.T, compoundNames []string, expectedNames []st
 		assert.Contains(t, compoundNames, name)
 	}
 }
+
+func TestScalarFunctionsMatch(t *testing.T) {
```
jacques-n, line 570: rather than have a mega yaml and many test cases, can we split into little yamls for each test case?

anshuldata:
I will un-draft once #53  is merged and is rebased for this PR

extensions/variants.go
anshuldata: I have marked a TODO for variadic function. I need to understand behavour of variadic function more to fix

extensions/variants.go
anshuldata: fixed it to use "github.com/creasty/defaults "

extensions/variants.go
anshuldata: added doc string

extensions/variants.go
anshuldata: added check if HasSyncParams then return unsupported error and a test case too

functions/dialect_test.go
anshuldata: splitted it

extensions/variants.go
anshuldata: fixed APIs to work with variadic function

anshuldata:
This PR is ready for review

jacques-n (approved):
LGTM. Thanks @anshuldata !

PR #55, opened 2024-09-15
fix: allow variadic variants to evaluate

zeroshade:
During some testing of building some expressions, I discovered that the `EvaluateTypeExpression` function didn't properly handle variadic arguments for functions. For example, the boolean or function in the standard substrait extensions uses 1 argument with the rest defined via variadics with a Min of 0. 

So we need to update the method to properly handle variadic expressions.


anshuldata (approved):
lgtm

PR #56, opened 2024-09-16
treat default of supported_as_column as true in dialect

scgkiran:


functions/dialect.go
```diff
@@ -187,7 +192,14 @@ type dialectFile struct {
 
 type dialectTypeInfo struct {
 	SqlTypeName       string `yaml:"sql_type_name"`
-	SupportedAsColumn bool   `yaml:"supported_as_column"`
+	SupportedAsColumn *bool  `yaml:"supported_as_column"`
```
jacques-n, line 195: can we start using the defaults pattern that we introduced in the matching patches? Example [here](https://github.com/substrait-io/substrait-go/commit/1a800d967cdceef1f29c3a6d7fdd06ac5d37fc30#diff-00e5053c917f74402204735708fe3fb7f91142e641d66f3c1749f8f09f33ccc7R191).

functions/dialect.go
```diff
@@ -145,6 +151,12 @@ func (d *dialectImpl) Load(reader io.Reader) error {
 	if err != nil {
 		return err
 	}
+	for _, ti := range d.file.SupportedTypes {
```
jacques-n, line 154: Can't we just run this against dialectFile? Then we can use it everywhere.

functions/dialect.go
```diff
@@ -145,6 +151,12 @@ func (d *dialectImpl) Load(reader io.Reader) error {
...
+		if err := defaults.Set(ti); err != nil {
```
jacques-n, line 155: Can't we just run this against dialectFile? Then we can use it everywhere.

functions/dialect.go
```diff
@@ -176,18 +188,22 @@ func (d *dialectImpl) buildFunctionInfoMap(functions []dialectFunction) map[exte
...
-	Name               string                     `yaml:"name"`
-	Type               string                     `yaml:"type"`
-	SupportedTypes     map[string]dialectTypeInfo `yaml:"supported_types,omitempty"`
-	ScalarFunctions    []dialectFunction          `yaml:"scalar_functions,omitempty"`
-	AggregateFunctions []dialectFunction          `yaml:"aggregate_functions,omitempty"`
-	WindowFunctions    []dialectFunction          `yaml:"window_functions,omitempty"`
-	Dependencies       map[string]string          `yaml:"dependencies,omitempty"`
+	Name               string                      `yaml:"name"`
+	Type               string                      `yaml:"type"`
+	SupportedTypes     map[string]*dialectTypeInfo `yaml:"supported_types,omitempty"`
```
jacques-n, line 193: why the change to pointer?

functions/dialect.go
```diff
@@ -176,18 +188,22 @@ func (d *dialectImpl) buildFunctionInfoMap(functions []dialectFunction) map[exte
...
+	ScalarFunctions    []dialectFunction           `yaml:"scalar_functions,omitempty"`
+	AggregateFunctions []dialectFunction           `yaml:"aggregate_functions,omitempty"`
+	WindowFunctions    []dialectFunction           `yaml:"window_functions,omitempty"`
+	Dependencies       map[string]string           `yaml:"dependencies,omitempty"`
 }
 
 type dialectTypeInfo struct {
 	SqlTypeName       string `yaml:"sql_type_name"`
-	SupportedAsColumn bool   `yaml:"supported_as_column"`
+	SupportedAsColumn *bool  `yaml:"supported_as_column" default:"true"`
```
jacques-n, line 202: let's avoid change to pointer

functions/dialect.go
```diff
@@ -176,18 +188,22 @@ func (d *dialectImpl) buildFunctionInfoMap(functions []dialectFunction) map[exte
...
+}
+
+func (ti *dialectTypeInfo) isSupportedAsColumn() bool {
```
jacques-n, line 205: shouldn't need to introduce this if we stay away from pointer.

functions/dialect.go
scgkiran: yes, this helped removing one pointer.

functions/dialect.go
scgkiran: Removed it. not needed after running SetDefaults on dialect file.

functions/dialect.go
scgkiran: This is needed for SetDefaults to differentiate the zero value from the provided value.

Without this it is incorrectly setting the value to true even if the input is
`supported_as_column: false`

scgkiran: I tried UnmarshalYAML, one hiccup with this is we cannot use `defaults.Set`  on any struct using dialectTypeInfo. I had to remove `defaults.Set` on dialectFile. 

Looks like whenever we have `default:"true"` the choice is between 1. using pointer vs 2. adding UnmarshalYAML in all the structs (that includes this bool indirectly)  to set the defaults.

LMK 1 or 2?

jacques-n (approved):
lgtm

PR #57, opened 2024-09-18
Add Min/Max ArgumentCount API to functionvariant

anshuldata:


extensions/variants.go
```diff
@@ -218,6 +222,20 @@ func parseFuncName(compoundName string) (name string, args ArgumentList) {
...
+		return len(paramTypeList)
+	}
+	return variadicBehavior.Min
+}
+
+func maxArgumentCount(paramTypeList ArgumentList, variadicBehavior *VariadicBehavior) int {
+	if variadicBehavior == nil {
+		return len(paramTypeList)
+	}
+	return variadicBehavior.Max
```
zeroshade, line 236: shouldn't this be `len(paramTypeList) + variadicBehavior.Max`?

extensions/variants.go
```diff
@@ -218,6 +222,20 @@ func parseFuncName(compoundName string) (name string, args ArgumentList) {
 	return name, args
 }
 
+func minArgumentCount(paramTypeList ArgumentList, variadicBehavior *VariadicBehavior) int {
+	if variadicBehavior == nil {
+		return len(paramTypeList)
+	}
+	return variadicBehavior.Min
```
zeroshade, line 229: shouldn't this be `len(paramTypeList) + variadicBehavior.Min` ?

extensions/variants.go
```diff
@@ -398,6 +424,13 @@ func (s *AggregateFunctionVariant) Match(argumentTypes []types.Type) (bool, erro
 func (s *AggregateFunctionVariant) MatchAt(typ types.Type, pos int) (bool, error) {
 	return matchArgumentAt(typ, pos, s.Nullability(), s.impl.Args, s.impl.Variadic)
 }
+func (s *AggregateFunctionVariant) MinArgumentCount() int {
+	return minArgumentCount(s.impl.Args, s.impl.Variadic)
```
zeroshade, line 428: add a test for this?

extensions/variants.go
```diff
@@ -398,6 +424,13 @@ func (s *AggregateFunctionVariant) Match(argumentTypes []types.Type) (bool, erro
...
+}
+
+func (s *AggregateFunctionVariant) MaxArgumentCount() int {
+	return maxArgumentCount(s.impl.Args, s.impl.Variadic)
```
zeroshade, line 432: add a test?

extensions/variants.go
```diff
@@ -513,6 +546,14 @@ func (s *WindowFunctionVariant) MatchAt(typ types.Type, pos int) (bool, error) {
 	return matchArgumentAt(typ, pos, s.Nullability(), s.impl.Args, s.impl.Variadic)
 }
 
+func (s *WindowFunctionVariant) MinArgumentCount() int {
+	return minArgumentCount(s.impl.Args, s.impl.Variadic)
```
zeroshade, line 550: add a test?

extensions/variants.go
```diff
@@ -513,6 +546,14 @@ func (s *WindowFunctionVariant) MatchAt(typ types.Type, pos int) (bool, error) {
...
+}
+
+func (s *WindowFunctionVariant) MaxArgumentCount() int {
+	return maxArgumentCount(s.impl.Args, s.impl.Variadic)
```
zeroshade, line 554: add a test?

extensions/variants.go
```diff
@@ -218,6 +222,20 @@ func parseFuncName(compoundName string) (name string, args ArgumentList) {
 	return name, args
 }
 
+func minArgumentCount(paramTypeList ArgumentList, variadicBehavior *VariadicBehavior) int {
+	if variadicBehavior == nil {
+		return len(paramTypeList)
```
zeroshade, line 227: your test calls `MinArgumentCount` but somehow this entire function isn't covered by the tests, how is that?

extensions/variants.go
```diff
@@ -218,6 +222,20 @@ func parseFuncName(compoundName string) (name string, args ArgumentList) {
...
+	}
+	return variadicBehavior.Min
+}
+
+func maxArgumentCount(paramTypeList ArgumentList, variadicBehavior *VariadicBehavior) int {
+	if variadicBehavior == nil {
+		return len(paramTypeList)
```
zeroshade, line 234: we are calling `MaxArgumentCount` in the tests, but somehow this entire function isn't covered by the tests according to codecov, how is that?

extensions/variants.go
jacques-n: I think it should actually be len(paramTypeList) + variadicBehavior.Min - 1, no?

zeroshade: going by the example of the [`or` function](https://github.com/substrait-io/substrait/blob/main/extensions/functions_boolean.yaml)

it has one parameter and a `Min` of 0 for the variadic. As a result, the Minimum number of arguments is 1, `len(paramTypeList) + variadicBehavior.Min`

If you use the `- 1` then it would have a minimum of 0 which seems wrong, right?

extensions/variants.go
anshuldata: fixed it

extensions/variants.go
anshuldata: I think this should be "len(paramTypeList) + variadicBehavior.Min".
Fixed it

extensions/variants.go
anshuldata: added tests for this

extensions/variants.go
anshuldata: added tests for this, thanks for pointing this. I missed it

extensions/variants.go
anshuldata: added tests for this, thanks for pointing this. I missed it

extensions/variants.go
anshuldata: added tests for this, thanks for pointing this. I missed it

extensions/variants.go
anshuldata: I think this test coverage script doesn't detect coverage hit across pacakge. I have tried it in IDE and it shows fine.
I will figure out and fix as a separate PR

Screenshot 2024-09-25 at 12 46 51 PM

extensions/variants.go
anshuldata: same as previous comment's response

anshuldata:
is this PR good @zeroshade

zeroshade (approved):

PR #59, opened 2024-09-25
Add substrait as a module

anshuldata:
* Fetched extension definition from substrait module
* Removed "definition" directory which has pre-loaded extension files
* Fixed intervalDay type to take precision.
* For unparsable extension file, logging and ignore failure. This is required since parsing needs to be enhanced to parse extension grammar fully. Below 4 files are non-parsable today
  1. functions_arithmetic_decimal.yaml (need to parse complex return type)
  2.  functions_geometry.yaml (need to parse geometry type)
  3. type_variations.yaml
  4. unknown.yaml


zeroshade:
The `go` toolchain does not work with git submodules which is why we weren't already using a submodule.

Switching to using a git submodule will result in users being required to package the default definitions themselves when they try to use this library instead of being able to get them for free and having them embedded. To that end, the `//go:embed` directive will fail to work unless the user vendors this themselves. 

All in all, using a submodule for this is not viable

jacques-n:
I'd like to propose packaging substrait core repo as a go module that substrait-go depends on. That way we can avoid the copying and have clear connections. I think we can do this by proposing the following in the base repo:

- minimal go.mod for "substrait-core"
- single go file in root that does an //go:embed of the extension files (and tests in future)
- publish that to go package registry
- refer to in substrait-go

We can publish on the same version scheme/timeline as the rest of substrait-core.

anshuldata:
> All in all, using a submodule for this is not viable
Thanks for pointing this @zeroshade . I have fixed it as based on suggestions on @jacques-n 
I have opened [this](https://github.com/substrait-io/substrait/pull/712) PR in substrait.
I tested approach suggested with my own github repo which emulates substrait as go module.
I am marking this PR as draft and will undraft once PR of substrait as go module merges

anshuldata:
This is ready for review @zeroshade @jacques-n '

extensions/extension_mgr.go
```diff
@@ -20,31 +21,46 @@ type AdvancedExtension = extensions.AdvancedExtension
 const SubstraitDefaultURIPrefix = "https://github.com/substrait-io/substrait/blob/main/extensions/"
 
 // DefaultCollection is loaded with the default Substrait extension
-// definitions with the exception of decimal arithmetic. Decimal arithmetic
-// functions are missing as the complex return type expressions are not
-// yet implemented.
+// definitions.
+// Parser needs to enhanced until than below function files are not processed
+// 1. functions_arithmetic_decimal.yaml (need to parse complex return type)
```
jacques-n, line 26: let's remove the specific names here. seems like overkill.

jacques-n (commented):
Generally looks good.

Is the intervalday change a breaking change for downstream consumers? Hard to tell looking at the patch.

zeroshade (approved):

anshuldata:
> Is the intervalday change a breaking change for downstream consumers?
No to be precise not in this PR but it will be in next PR.
In next PR, I will fix IntervalDayType precision to be passed to/from proto ([here](https://github.com/substrait-io/substrait/blob/683f4179a058c2c99c04501b920a48ff372356ff/proto/substrait/type.proto#L138)). 
Before this PR, IntervalDayType didn't have precision field which in proto mean microsecond precision. After next PR is merged, if Consumer hasn't set precision in IntervalDayType it will mean 0  (i.e. Second) .

extensions/extension_mgr.go
anshuldata: I will fix it in next PR

anshuldata: fixed in #61

PR #60, opened 2024-10-10
feat: support user-defined types in YAML extensions

kadinrabo:
Support parsing `UserDefinedType` (denoted by u!) in extension yaml

### DIFF
--- a/extensions/simple_extension_test.go
+++ b/extensions/simple_extension_test.go
@@ -3,9 +3,6 @@
 package extensions_test
 
 import (
-	"github.com/substrait-io/substrait-go/proto"
-	"github.com/substrait-io/substrait-go/types"
-	"github.com/substrait-io/substrait-go/types/parser"
 	"strings"
 	"testing"
 
@@ -13,6 +10,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/substrait-io/substrait-go/extensions"
+	"github.com/substrait-io/substrait-go/proto"
+	"github.com/substrait-io/substrait-go/types"
+	"github.com/substrait-io/substrait-go/types/parser"
 )
 
 func TestUnmarshalSimpleExtension(t *testing.T) {
--- a/extensions/simple_extension_test.go
+++ b/extensions/simple_extension_test.go
@@ -41,6 +41,36 @@ types:
 	assert.Nil(t, f.Types[3].Structure)
 }
 
+func TestUnmarshalCustomScalarFunction(t *testing.T) {
+	const customDef = `
+scalar_functions:
+  - name: "scalar1"
+    impls:
+      - args:
+          - name: arg1
+            value: u!customtype1
+        return: i64
+  - name: "scalar2"
+    impls:
+      - args:
+          - name: arg1
+            value: i64
+        return: u!customtype2?
+`
+
+	var f extensions.SimpleExtensionFile
+	require.NoError(t, yaml.Unmarshal([]byte(customDef), &f))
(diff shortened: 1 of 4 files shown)


vbarua (approved):
Looks reasonable to me, but I'm not super familiar with the Go side of things.

cc: @zeroshade, @jacques-n: should be an easy review.

extensions/simple_extension_test.go
```diff
@@ -41,6 +41,36 @@ types:
...
+
+	assert.Len(t, f.ScalarFunctions, 2)
+	assert.Equal(t, "scalar1", f.ScalarFunctions[0].Name)
+	assert.IsType(t, extensions.ValueArg{}, f.ScalarFunctions[0].Impls[0].Args[0])
+	arg1 := f.ScalarFunctions[0].Impls[0].Args[0].(extensions.ValueArg)
+	assert.Equal(t, "u!customtype1", arg1.Value.String())
+	assert.Equal(t, "scalar2", f.ScalarFunctions[1].Name)
+	assert.IsType(t, extensions.ValueArg{}, f.ScalarFunctions[1].Impls[0].Args[0])
+	assert.Equal(t, "u!customtype2?", f.ScalarFunctions[1].Impls[0].Return.String())
+}
```
zeroshade, line 72: Can we also test that the type of the argument is correctly produced as a `*UserDefinedType`?

types/parser/type_parser.go
```diff
@@ -122,7 +122,7 @@ func (t *typename) Capture(values []string) error {
 }
 
 type nonParamType struct {
-	TypeName    typename `parser:"@(IntType | Boolean | FPType | Temporal | BinaryType)"`
+	TypeName    typename `parser:"@(IntType | Boolean | FPType | Temporal | BinaryType | UserDefinedType)"`
```
zeroshade, line 125: `RetType()` probably needs to be updated to handle a user-defined type scenario. I'm pretty sure `SimpleTypeNameToType` will not currently return a user-defined type, it'll error out.

zeroshade (changes requested):

jacques-n:
I believe @anshuldata is or will be working on a patch to replace the hand parsing with antlr grammar. That should address several gaps in the current function parsing. I'd be inclined to wait until that is merged before trying to solve this gap (as some of it may come for cheap/free).

vbarua:
@jacques-n do you think that will be available next week? We have some stuff internally that's blocked on us being able to handle user-defined types. Aside from potentially redundant work on our end, would there be any harm in updating the hand parser.

extensions/simple_extension_test.go
kadinrabo: I added a nullability check too 👍

zeroshade:
@vbarua It's unlikely that something as large as replacing with an antlr grammar will be merged next week as I'm sure that would need significant review and testing. My personal opinion would be that it makes sense to update this if you need it. I doubt there's any harm besides the redundant work.

@jacques-n thoughts?

jacques-n:
I think this patch is building debt on debt. 

In general, I find it to be a bad habit to accept something we know we're going to delete in a couple of weeks (unless we're fixing a critical regression). It sets a bad precedent for new contributors on how we work. If we know we're going in a certain direction and one contributor wants to go there faster than another, let's just have them take us to the target outcome.

I'm -0 on this patch. I'm not going to block.

zeroshade (approved):
LGTM assuming that @jacques-n has no objections

vbarua:
Based on 
> I'm -0 on this patch. I'm not going to block.

I'm going to merge this in.

PR #63, opened 2024-10-30
fix(plan,expr): use Expressions fields in VirtualTable instead of Literal

anshuldata:
* This change migrates VirtualTable to not use deprecated field and instead use newly introduced field in [this](https://github.com/substrait-io/substrait/pull/727) PR
* Fix VirtualTable proto to generate Expression Struct instead of Literal Struct
* This PR doesn't have breaking change
* This PR doesn't introduce PublicAPI to use Expression in building VirtualTableRelation. I will introduce it as separate PR


anshuldata:
Converting this PR to draft. I will raise separate PR for proto update. I will rebase and undraft this PR once proto PR is merged

EpsilonPrime:
This will remove the literal support so consuming older plans won't work but that should be okay given the current usage of this feature.

anshuldata:
> This will remove the literal support so consuming older plans won't work but that should be okay given the current usage of this feature.

@EpsilonPrime this PR does consume older plans. I have added [this](https://github.com/substrait-io/substrait-go/pull/63/files#diff-e2f5ca32a8d7e47b673731a102aaa16a4437822a31a34e1cb2ce5477a90c3334R33) test case to make sure this. Let me know if I am missing something

expr/literals.go
```diff
@@ -45,25 +45,6 @@ type (
 	Null types.Type
 )
 
-func StructLiteralFromProto(s *proto.Expression_Literal_Struct) StructLiteralValue {
```
zeroshade, line 48: Since this is currently exported, removing it is a breaking change. 

My preference would be to just add a `// Deprecated: use VirtualTableExprFromProto instead` and not remove it. 

If we're going to remove this, then we need to mark this as breaking.

zeroshade (changes requested):

expr/literals.go
anshuldata: ok, my thought was, though this API is public but only usage of this API can be to construct a plan from proto. So I felt this won't be used from outside substrait-go.  So I felt safe to remove this API and change usage of this to by introducing `VirtualTableExprFromLiteralProto(s *proto.Expression_Literal_Struct) VirtualTableExpressionValue`.
To signal this intent I have migrated new API introduced in this PR to "plan/internal" package in latest commit

Also,  I have accommodated changes to mark this API deprecated

anshuldata:
@zeroshade we need to change github repository setting to not allow merge if change is requested.
Screenshot 2024-11-08 at 5 33 20 PM

We need to add "Branch protection rules" rules as below (branches --> Branch protection rules)

Screenshot 2024-11-08 at 5 41 00 PM

zeroshade:
@anshuldata I don't have the permissions for that.

@jacques-n could you make the suggested change to the branch protection rules?

expr/literals.go
zeroshade: Looks good to me. Thanks for adding the `Deprecated` mark

zeroshade (approved):

jacques-n:
> could you make the suggested change to the branch protection rules?

It looks like it is already added. Maybe someone else took the initiative? Or I'm looking in the wrong place...

EpsilonPrime:
I added them.  Looks like we need some added test coverage to merge this PR.

anshuldata:
> Looks like we need some added test coverage to merge this PR.
 code coverage doesn't recognize coverage in tests across package. So test in plan_builder_test coverage which covers expression.go changes is not accounted. 
I need to figure out coverage option to detect it

PR #64, opened 2024-10-30
feat: Use antlr parser to parse the types in extensions

scgkiran:
1. This needs a `go generate` changes to pull the grammar from substarit-core and auto generate antlr parser. For now I included the files in this PR.
2. This is just parser replacement. No new functionality added w.r.to loading new extensions or supporting type derivation.


grammar/Makefile
```diff
@@ -0,0 +1,9 @@
+ANTLR_JAR=antlr-4.13.2-complete.jar
```
jacques-n, line 1: let's put this in a go generate block instead of adding make to the project.

grammar/Makefile
```diff
@@ -0,0 +1,9 @@
...
+GRAMMARS=SubstraitLexer.g4 SubstraitType.g4
+OUTPUT_DIR=../parser/types/baseparser
+
+generate:
```
jacques-n, line 5: we should be download grammar from substrait core rather than checking into this repo.

grammar/Makefile
jacques-n: For example, a block something like this should work:

```
//go:generate wget -nc https://www.antlr.org/download/antlr-4.11.1-complete.jar
//go:generate -command antlr java -Xmx500M -cp "./antlr-4.11.1-complete.jar:$CLASSPATH" org.antlr.v4.Tool
//go:generate antlr -Dlanguage=Go -package substrait bar/foo.g4 -Werror -long-messages -o ./
```

jacques-n (commented):
Let's update codecov as part of this patch to ignore the newly generated files.

grammar/Makefile
scgkiran: Waiting on this [grammar PR](https://github.com/substrait-io/substrait/pull/730)  to get merged

scgkiran:
> Let's update codecov as part of this patch to ignore the newly generated files.

Done

grammar/Makefile
scgkiran: Fixed it.

grammar/Makefile
scgkiran: Done

types/parameterized_user_defined_type.go
```diff
@@ -0,0 +1,204 @@
...
+	toTypeParam() (TypeParam, error)
+	MatchWithoutNullability(param TypeParam) bool
+	MatchWithNullability(param TypeParam) bool
+}
+
+type DataTypeUDTParam struct {
+	Type FuncDefArgType
+}
+
+func (d *DataTypeUDTParam) isTypeParameter() {}
```
EpsilonPrime, line 22: Is this supposed to return a boolean?

types/parser/baseparser/SubstraitLexer.tokens
EpsilonPrime, line 1: Shouldn't need to be checked in.

types/parser/baseparser/SubstraitLexer.interp
EpsilonPrime, line 1: Shouldn't need to be checked in.

types/parser/baseparser/substraittype_base_listener.go
EpsilonPrime, line 1: If we're checking these in to ease go module behavior we should probably have a README.md in this file to tell folks about these files and how/when they should need to update.

types/parser/parse.go
```diff
@@ -0,0 +1,111 @@
...
+	substraitgo "github.com/substrait-io/substrait-go"
+	"github.com/substrait-io/substrait-go/types"
+	baseparser2 "github.com/substrait-io/substrait-go/types/parser/baseparser"
+)
+
+type TypeExpression struct {
+	ValueType types.FuncDefArgType
+}
+
+func (t *TypeExpression) MarshalYAML() (interface{}, error) {
```
EpsilonPrime, line 16: A test to test the linkage for marshaling into/out of YAML would likely be useful.

types/parser/parse.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (l *simpleErrorListener) ErrorCount() int {
+	return l.errorCount
+}
+
+func (l *simpleErrorListener) GetError() string {
+	return l.errors[0]
+}
+
+func (l *simpleErrorListener) GetErrors() []string {
```
EpsilonPrime, line 67: I would expect tests to use GetErrors() instead of GetError() to detect when multiple errors appear (not just the one expected one).

types/parser/parse.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (l *simpleErrorListener) ReportContextSensitivity(recognizer antlr.Parser, dfa *antlr.DFA, startIndex, stopIndex, prediction int, configs *antlr.ATNConfigSet) {
+}
+
+func (l *simpleErrorListener) ErrorCount() int {
+	return l.errorCount
+}
+
+func (l *simpleErrorListener) GetError() string {
```
EpsilonPrime, line 63: This shouldn't be called if ErrorCount() is not positive.  I'd argue that the caller can easily get this on their own since they're also the one doing the count check.   Remove it?

types/parser/visitor.go
```diff
@@ -0,0 +1,319 @@
...
+
+func (v *TypeVisitor) VisitMap(ctx *baseparser2.MapContext) interface{} {
+	nullability := types.NullabilityRequired
+	if ctx.GetIsnull() != nil {
+		nullability = types.NullabilityNullable
+	}
+	keyType, keyOk := v.Visit(ctx.GetKey()).(types.FuncDefArgType)
+	valueType, valueOk := v.Visit(ctx.GetValue()).(types.FuncDefArgType)
+	if !keyOk || !valueOk {
+		panic("Map key or value type is not a FuncDefArgType")
```
EpsilonPrime, line 295: Do we want production code to have to deal with panics?  An alternative would be to use antlr's error handling mechanism.

types/parser/parse.go
scgkiran: added

types/parser/baseparser/substraittype_base_listener.go
scgkiran: added readme.

types/parameterized_user_defined_type.go
scgkiran: yes, fixed it.

types/parser/baseparser/SubstraitLexer.interp
scgkiran: removed it

types/parser/baseparser/SubstraitLexer.tokens
scgkiran: removed it

EpsilonPrime (approved):

PR #65, opened 2024-11-01
feat: switch type of fetch args in builder to match proto

EpsilonPrime:
Recently FetchRel's definition was modified to add -1 as a valid value for count.  This allows one to specify an offset but return all the remaining rows.

### DIFF
--- a/plan/builders.go
+++ b/plan/builders.go
@@ -80,8 +80,8 @@ type Builder interface {
 	CreateTableAsSelect(input Rel, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error)
 	CrossRemap(left, right Rel, remap []int32) (*CrossRel, error)
 	Cross(left, right Rel) (*CrossRel, error)
-	FetchRemap(input Rel, offset, count uint64, remap []int32) (*FetchRel, error)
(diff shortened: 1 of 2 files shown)


plan/builders.go
```diff
@@ -80,8 +80,8 @@ type Builder interface {
 	CreateTableAsSelect(input Rel, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error)
 	CrossRemap(left, right Rel, remap []int32) (*CrossRel, error)
 	Cross(left, right Rel) (*CrossRel, error)
-	FetchRemap(input Rel, offset, count uint64, remap []int32) (*FetchRel, error)
-	Fetch(input Rel, offset, count uint64) (*FetchRel, error)
+	FetchRemap(input Rel, offset, count int64, remap []int32) (*FetchRel, error)
```
jacques-n, line 83: I suggest we add a FETCH_COUNT_ALL_RECORDS const or similar and add doc on this function and the below mentioning it's use.

jacques-n (commented):
Would be good to update PR subject to match commit standard. One minor doc & const request, otherwise looks good.

jacques-n (approved):
LGTM. Thanks @EpsilonPrime !

PR #68, opened 2024-11-08
feat:  Add a way to construct a field ref using a known type

EpsilonPrime:


expr/field_reference.go
```diff
@@ -540,6 +540,15 @@ func NewRootFieldRef(ref Reference, baseSchema *types.StructType) (*FieldReferen
 	return NewFieldRef(RootReference, ref, baseSchema)
 }
 
+// NewRootFieldRefFromType creates a new field reference with a specific known type.  Prefer using NewRootFieldRef.
+func NewRootFieldRefFromType(ref Reference, t types.Type) (*FieldReference, error) {
```
zeroshade, line 544: Would it make sense to also have a `NewFieldRefFromType` function and just have this delegate to that (in the same way that `NewRootFieldRef` delegates to `NewFieldRef`)?

zeroshade (approved):
Just one question, otherwise LGTM

expr/field_reference.go
EpsilonPrime: Added.  It feels gutted if it isn't going to do any work like determine the type of an expression reference.

PR #71, opened 2024-11-20
feat:  Add new Remap interface to relations

EpsilonPrime:
Remap() is an alternative to the current Remap variants currently found on the Builder interface.  It preserves the existing behavior including checking the proposed mapping for validity.  It also modifies the relation instead of returning a copy.  All of the existing Remap function variants have been marked as deprecated in favor of Remap.

### DIFF
--- a/plan/builders.go
+++ b/plan/builders.go
@@ -71,13 +71,18 @@ type Builder interface {
 	// from the output.
 
 	Project(input Rel, exprs ...expr.Expression) (*ProjectRel, error)
+	// Deprecated: Use Project(...).ChangeMapping() instead.
 	ProjectRemap(input Rel, remap []int32, exprs ...expr.Expression) (*ProjectRel, error)
+	// Deprecated: Use AggregateColumns(...).ChangeMapping() instead.
 	AggregateColumnsRemap(input Rel, remap []int32, measures []AggRelMeasure, groupByCols ...int32) (*AggregateRel, error)
 	AggregateColumns(input Rel, measures []AggRelMeasure, groupByCols ...int32) (*AggregateRel, error)
+	// Deprecated: Use AggregateExprs(...).ChangeMapping() instead.
 	AggregateExprsRemap(input Rel, remap []int32, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error)
 	AggregateExprs(input Rel, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error)
+	// Deprecated: Use CreateTableAsSelect(...).ChangeMapping() instead.
 	CreateTableAsSelectRemap(input Rel, remap []int32, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error)
 	CreateTableAsSelect(input Rel, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error)
+	// Deprecated: Use Cross(...).ChangeMapping() instead.
 	CrossRemap(left, right Rel, remap []int32) (*CrossRel, error)
 	Cross(left, right Rel) (*CrossRel, error)
 	// FetchRemap constructs a fetch relation providing an offset (skipping some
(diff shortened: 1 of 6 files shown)


plan/relations.go
```diff
@@ -581,6 +593,18 @@ func (p *ProjectRel) CopyWithExpressionRewrite(rewriteFunc RewriteFunc, newInput
 	return &proj, nil
 }
 
+func (p *ProjectRel) ChangeMapping(mapping []int32) error {
+	noutput := int32(len(p.Remap(p.RecordType()).Types))
```
jacques-n, line 597: If this operation is already mapped, we should be applying the mapping on top of the existing one, not replacing it.

plan/plan.go
```diff
@@ -275,6 +275,7 @@ type Rel interface {
 	// record type from the output RecordType of the underlying Relation.
 	// If the output mapping is nil, then this just returns the input
 	// type.
+	// TODO - Rename this ApplyMap.
```
jacques-n, line 278: Add api to interface.

Also, I think we should be copying, not overwriting. Overwriting/mutating nodes in a plan can be dangerous given the possibility of duplicated subplans, etc.

Maybe call CopyWithRemap() or similar.

plan/relations.go
```diff
@@ -148,6 +148,18 @@ func (b *baseReadRel) updateFilters(filters []expr.Expression) {
 	b.filter, b.bestEffortFilter = filters[0], filters[1]
 }
 
+func (b *baseReadRel) ChangeMapping(mapping []int32) error {
+	noutput := int32(len(b.Remap(b.RecordType()).Types))
```
jacques-n, line 152: We should be able to write this logic once, not many times.

plan/plan.go
EpsilonPrime: Remap doesn't really make sense on write relations so I specifically created an alternative interface.

I'm less inclined to worry about copying as these are more of a build time artifact (i.e. not protos).  If one wants to duplicate a subplan that should do the copying.

plan/relations.go
EpsilonPrime: I don't see any utility to remap something that's already been mapped.  I could add a check to return an error if someone attempts to remap.

plan/relations.go
```diff
@@ -741,6 +776,12 @@ func (j *JoinRel) CopyWithExpressionRewrite(rewriteFunc RewriteFunc, newInputs .
 	return &join, nil
 }
 
+func (j *JoinRel) ChangeMapping(mapping []int32) error {
+	newMapping, err := ChangeMapping(j, mapping)
+	j.mapping = newMapping
```
jacques-n, line 781: If there is an error, we lose the old mapping if I reading this right. If err, I would expect no change.

plan/plan.go
```diff
@@ -271,6 +271,16 @@ type Rel interface {
...
+	// ClearMapping resets the mapping for this relation.
+	ClearMapping()
+	// ChangeMapping modifies the current relation by applying the provided
+	// mapping to the current relation.  Typically used to remove any unneeded
+	// columns or provide them in a different order.  If there already is a
+	// mapping on this relation, this provides mapping over the current mapping.
+	//
+	// If any column numbers specified are outside the currently available input
+	// range an error is returned and the mapping is left unchanged.
+	ChangeMapping(mapping []int32) error
```
jacques-n, line 283: Can we have this make a copy as opposed to an inplace edit? 

I'm also more OverlayMapping or ApplyMapping as a name.

Also, would be good to use vararg.

plan/plan.go
```diff
@@ -271,6 +271,16 @@ type Rel interface {
 	// result should be 3 columns consisting of the 5th, 2nd and 1st
 	// output columns from the underlying relation.
 	OutputMapping() []int32
+	// ClearMapping resets the mapping for this relation.
+	ClearMapping()
```
jacques-n, line 275: I don't think we should add this.

plan/plan.go
```diff
@@ -271,6 +271,16 @@ type Rel interface {
 	// result should be 3 columns consisting of the 5th, 2nd and 1st
 	// output columns from the underlying relation.
 	OutputMapping() []int32
```
jacques-n, line 273: I think we should deprecate this.

EpsilonPrime: done.

plan/relations.go
EpsilonPrime: Fixed the bug in the ChangeMapping helper and added a test.

plan/builders.go
```diff
@@ -71,44 +71,60 @@ type Builder interface {
...
+	// Deprecated: Use Fetch(...).ChangeMapping() instead.
 	JoinAndFilter(left, right Rel, condition, postJoinFilter expr.Expression, joinType JoinType) (*JoinRel, error)
+	// Deprecated: Use Join(...).ChangeMapping() instead.
 	JoinRemap(left, right Rel, condition expr.Expression, joinType JoinType, remap []int32) (*JoinRel, error)
 	Join(left, right Rel, condition expr.Expression, joinType JoinType) (*JoinRel, error)
+	// Deprecated: Use NamedScan(...).ChangeMapping() instead.
 	NamedScanRemap(tableName []string, schema types.NamedStruct, remap []int32) (*NamedTableReadRel, error)
 	NamedScan(tableName []string, schema types.NamedStruct) *NamedTableReadRel
+	// Deprecated: Use NamedWriteMap(...).ChangeMapping() instead.
 	NamedWriteRemap(input Rel, op WriteOp, tableName []string, schema types.NamedStruct, remap []int32) (*NamedTableWriteRel, error)
```
jacques-n, line 114: Looks like deprecation message is wrong and NamedWrite() is missing.

plan/plan.go
EpsilonPrime: Removed.

plan/builders.go
EpsilonPrime: fixed.

plan/plan.go
EpsilonPrime: Renamed to `Remap()`.  There could be some confusion with `remap()` which we could rename to `applyMapping()`.

plan/relations.go
```diff
@@ -212,6 +240,13 @@ func (n *NamedTableReadRel) CopyWithExpressionRewrite(rewriteFunc RewriteFunc, _
 	return &nt, nil
 }
 
+func (n *NamedTableReadRel) Remap(mapping ...int32) (Rel, error) {
+	newMapping, err := RemapHelper(n, mapping)
+	newRel := n
```
jacques-n, line 245: Since this is a pointer receiver, I believe this isn't a copy, right?

plan/relations.go
```diff
@@ -152,6 +152,34 @@ func (b *baseReadRel) updateFilters(filters []expr.Expression) {
 	b.filter, b.bestEffortFilter = filters[0], filters[1]
 }
 
+type MappableRel interface {
+	RecordType() types.RecordType
+	OutputMapping() []int32
+}
+
+// RemapHelper implements the core functionality of RemapHelper for relations.
+// It returns the relation's existing mapping on an error to ease being called.
```
jacques-n, line 161:
```suggestion
// It returns the relation's existing mapping or an error to ease being called.
```

plan/relations.go
```diff
@@ -267,6 +302,13 @@ func (v *VirtualTableReadRel) CopyWithExpressionRewrite(rewriteFunc RewriteFunc,
 	return &vtr, nil
 }
 
+func (v *VirtualTableReadRel) Remap(mapping ...int32) (Rel, error) {
+	newMapping, err := RemapHelper(v, mapping)
```
jacques-n, line 306: If I were you, I'd introduce a non exported set mapping function on relation and then this code could be factored out to a single invocation.


something like 
```
RemapHelper(r Rel, newMapping) (Rel, error) {
  newRel := r.Copy(v.GetInputs())
  err := r.setMapping(newMapping)
  return newRel, err
}
```

plan/relations.go
EpsilonPrime: fixed.

jacques-n (approved):

plan/relations.go
EpsilonPrime: Adding the unexported method to the Rel interface was the trick I needed to get that to work.  Thanks!

PR #73, opened 2024-11-21
feat:  Add expression wrapper to ExprBuilder

EpsilonPrime:
This will allow previously calcuated Expression objects to be used as function arguments.

### DIFF
--- a/expr/builder_test.go
+++ b/expr/builder_test.go
@@ -17,7 +17,10 @@ func TestExprBuilder(t *testing.T) {
 		Reg:        expr.NewEmptyExtensionRegistry(&extensions.DefaultCollection),
 		BaseSchema: &boringSchema.Struct,
 	}
-	precomputedExpression, _ := expr.NewLiteral(int32(3), false)
+	precomputedLiteral, _ := expr.NewLiteral(int32(3), false)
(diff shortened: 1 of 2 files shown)


expr/builder_test.go
```diff
@@ -17,6 +17,7 @@ func TestExprBuilder(t *testing.T) {
 		Reg:        expr.NewEmptyExtensionRegistry(&extensions.DefaultCollection),
 		BaseSchema: &boringSchema.Struct,
 	}
+	precomputedExpression, _ := expr.NewLiteral(int32(3), false)
```
zeroshade, line 20: should we use examples that aren't Literals since wrap methods and such for literals, such as using the builder to create an expression first and then using that expression later, etc.

zeroshade (approved):
Just one nit-pick but overall this looks good to me!

expr/builder_test.go
EpsilonPrime: I was pondering merging Wrap and WrapExpression into Wrap (with the functionality of WrapExpression) since they are essentially the same.

I will also add a non-literal Expression example.

zeroshade:
Looks great, thanks @EpsilonPrime!

PR #74, opened 2024-11-22
feat:  Introduce direct and final output schemas

EpsilonPrime:
Corrects the behavior of RecordType() to return the final output schema.

All internal uses of the x.Remap(x.RecordType()) pattern have been fixed to appropriate use directOutputSchema instead of RecordType.  All external uses of the library should call x.RecordType() to get the final version of the types exposed by a relation.

### DIFF
--- a/plan/plan.go
+++ b/plan/plan.go
@@ -243,7 +243,7 @@ func (r *Root) ToProtoPlanRel() *proto.PlanRel {
 func (r *Root) RecordType() types.NamedStruct {
 	return types.NamedStruct{
 		Names:  r.names,
-		Struct: r.input.Remap(r.input.RecordType()),
+		Struct: r.input.RecordType(),
 	}
 }
 
@@ -271,13 +271,11 @@ type Rel interface {
 	// result should be 3 columns consisting of the 5th, 2nd and 1st
 	// output columns from the underlying relation.
 	OutputMapping() []int32
-	// Remap utilizes the OutputMapping to construct the expected result
-	// record type from the output RecordType of the underlying Relation.
-	// If the output mapping is nil, then this just returns the input
-	// type.
-	Remap(types.StructType) types.StructType
-	// RecordType returns the output record type of the underlying relation
-	// as a struct type.
+	// directOutputSchema returns the output record type of the underlying
+	// relation as a struct type.  Mapping is not applied.
+	directOutputSchema() types.StructType
+	// RecordType returns the types used by all columns returned by
+	// this relation after applying any provided mapping.
 	RecordType() types.StructType
 
 	GetAdvancedExtension() *extensions.AdvancedExtension
@@ -532,9 +530,9 @@ func RelFromProto(rel *proto.Rel, reg expr.ExtensionRegistry) (Rel, error) {
 			return nil, fmt.Errorf("%w: set operation must not be unspecified", substraitgo.ErrInvalidRel)
 		}
 
-		primary := inputs[0].Remap(inputs[0].RecordType())
+		primary := inputs[0].RecordType()
 		for i, in := range inputs[1:] {
-			t := in.Remap(in.RecordType())
+			t := in.RecordType()
 			if !t.Equals(&primary) {
 				return nil, fmt.Errorf("%w: set operation field mismatch found in input #%d, expected %s, got %s",
 					substraitgo.ErrInvalidRel, i+1, &primary, &t)
--- a/plan/relations.go
+++ b/plan/relations.go
@@ -91,8 +91,12 @@ func (b *baseReadRel) fromProtoReadRel(rel *proto.ReadRel, reg expr.ExtensionReg
 	return nil
 }
 
-func (b *baseReadRel) RecordType() types.StructType {
+func (b *baseReadRel) directOutputSchema() types.StructType {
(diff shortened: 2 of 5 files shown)


zeroshade:
Instead of a breaking change, could we just deprecate the `Remap`/`RecordType` methods instead of removing them? (we still update all the calling locations to use the new functions). Or would that be potentially problematic instead?

plan/relations_test.go
```diff
@@ -345,3 +346,116 @@ func TestRelations_Copy(t *testing.T) {
 		})
 	}
 }
+
+// FakeRel is a pretend relation that allows direct control of its direct output schema.
+type FakeRel struct {
```
jacques-n, line 351: Since FakeRel is only used in this test, let's avoid exporting.

plan/plan.go
```diff
@@ -271,14 +271,12 @@ type Rel interface {
...
-	// record type from the output RecordType of the underlying Relation.
-	// If the output mapping is nil, then this just returns the input
-	// type.
-	Remap(types.StructType) types.StructType
-	// RecordType returns the output record type of the underlying relation
-	// as a struct type.
-	RecordType() types.StructType
+	// DirectOutputSchema returns the output record type of the underlying
+	// relation as a struct type.
+	DirectOutputSchema() types.StructType
```
jacques-n, line 276: Should we export this? I'm kind of inclined not to. What's your thoughts?

plan/plan.go
```diff
@@ -271,14 +271,12 @@ type Rel interface {
...
+	// FinalOutputSchema returns the final output schema used by applying
+	// the mapping over the unmapped DirectOutputSchema.
+	FinalOutputSchema() types.StructType
```
jacques-n, line 279: If we don't export above, we can keep this named RecordType(). Seems like a less disruptive breaking change.

jacques-n (commented):
Thanks for going through and fixing this up!

plan/relations_test.go
EpsilonPrime: Done.

plan/plan.go
EpsilonPrime: It does appear that all uses are internal to the package with two instances of it being used to handle an "and Remap" operation that will be obsoleted.  Also, it is easier to later expose this if it turns out to be needed than hide it later.

plan/plan.go
EpsilonPrime: RecordType behaves like DirectOutputSchema (often calling Remap to make it behave like FinalOutputSchema).  For almost all uses of the function that's wrong.  I'm in favor of a bigger break.  It will also get folks in the right place to deal with the upcoming Remap change (where we start pushing folks to using the separate Remap/ChangeMapping method).  Without RecordType() there is no reason for folks to call Remap() any longer which frees that name up (eventually) to use instead of ChangeMapping.

jacques-n: I'm not clear on where you're disagreeing. Let me be more concrete on how I look at this change:

1. RecordType() should have always returned the final output schema of the relation.
1. It didn't and that was a bug. If you only used emit_kind=direct, the bug had no impact
1. You're fixing the bug.
1. When fixing a bug, we shouldn't need to change the name of the function.
1. DirectOutputSchema() is an internal concern of the relation, and should not be exported.

Where is the disagreement?

EpsilonPrime: If you used RecordType() correctly it would be x.Remap(x.RecordType() which would remap the already remapped results.  Nearly all uses of it was as DirectOutputSchema().  In fact all uses of Remap() are with RecordType().

EpsilonPrime: As discussed offline, we can keep the name but we should transition to returning a different type (such as RecordType) instead of using types.StructType.  I did try implementing that here but it became way too large of a change.  My next PR will be to update Remap -- hopefully I can do that there.

plan/relations.go
```diff
@@ -516,6 +524,9 @@ func (p *ProjectRel) RecordType() types.StructType {
 		Types:       output,
 	}
 }
+func (p *ProjectRel) RecordType() types.StructType {
+	return p.Remap(p.directOutputSchema())
```
jacques-n, line 528: Can we unexport `remap()` as part of this change. We've removed it from the interface but I think we should also remove from export.

plan/plan.go
```diff
@@ -271,13 +271,11 @@ type Rel interface {
...
-	// type.
-	Remap(types.StructType) types.StructType
-	// RecordType returns the output record type of the underlying relation
-	// as a struct type.
+	// directOutputSchema returns the output record type of the underlying
+	// relation as a struct type.  Mapping is not applied.
+	directOutputSchema() types.StructType
+	// RecordType returns the types used by all columns returned by
+	// this relation after applying any provided mapping.
 	RecordType() types.StructType
```
jacques-n, line 279: can we redefine this so it returns a RecordType as opposed to StructType. StructType isn't actually right since relations don't return structs, they return sets of records. As a concrete distinction, StructType can be nullable and type variations. Neither of those things are relevant to a record type.

plan/relations.go
EpsilonPrime: Unexported.

jacques-n (approved):
Thanks for carrying this all the way to the end.

PR #76, opened 2024-11-26
feat: Use types.RecordType instead of types.StructType

EpsilonPrime:


types/types.go
```diff
@@ -1337,3 +1337,14 @@ func (n NamedStruct) String() string {
 
 	return b.String()
 }
+
+type RecordType StructType
```
zeroshade, line 1341: What is the explicit difference between a `StructType` and a `RecordType`? Can you add a docstring to the `type` line to explain it? 

Just wondering if this actually is sufficient to demonstrate the difference in usage or if there should be a bit more to the type.

EpsilonPrime: Agreed, there wasn't much behind the previous implementation.  I've updated the PR with a more featured RecordType with many helper methods such as FieldCount and Concat.

jacques-n (approved):
nice!

PR #79, opened 2024-11-29
feat: change ArgumentList name to FuncParameterList

scgkiran:
Fixed bugs in matching variadic arguments.

BREAKING CHANGE: 
- `ArgumentList` has been renamed to `FuncParameterList`. 

BREAKING CHANGE: 
- The `ReturnType` method in the `FuncDefArgType` interface now requires `FuncParameterList` and actual argument types as parameters.

### DIFF
--- a/extensions/simple_extension.go
+++ b/extensions/simple_extension.go
@@ -47,6 +47,7 @@ type TypeVariationFunctions string
 const (
 	TypeVariationInheritsFuncs TypeVariationFunctions = "INHERITS"
 	TypeVariationSeparateFuncs TypeVariationFunctions = "SEPARATE"
+	EnumTypeString                                    = "req" // TODO change this to "enum"
 )
 
 type TypeVariation struct {
@@ -68,7 +69,7 @@ type EnumArg struct {
 }
 
 func (EnumArg) toTypeString() string {
-	return "req"
+	return EnumTypeString
 }
 
 func (v EnumArg) GetTypeExpression() types.FuncDefArgType {
--- a/extensions/simple_extension.go
+++ b/extensions/simple_extension.go
@@ -59,6 +59,7 @@ type TypeVariation struct {
 
 type FuncParameter interface {
 	toTypeString() string
+	argumentMarker() // unexported marker method
 	GetTypeExpression() types.FuncDefArgType
 }
 
@@ -71,6 +72,8 @@ type EnumArg struct {
 func (EnumArg) toTypeString() string {
 	return EnumTypeString
 }
+
+func (v EnumArg) argumentMarker() {}
 
 func (v EnumArg) GetTypeExpression() types.FuncDefArgType {
 	return &types.EnumType{Name: v.Name, Options: v.Options}
@@ -87,6 +90,8 @@ func (v ValueArg) toTypeString() string {
 	return v.Value.ValueType.ShortString()
 }
 
+func (v ValueArg) argumentMarker() {}
+
 func (v ValueArg) GetTypeExpression() types.FuncDefArgType {
 	return v.Value.ValueType
 }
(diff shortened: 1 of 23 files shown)


scgkiran:
No logical change, just cleanup and renaming of variables.

extensions/simple_extension.go
```diff
@@ -55,9 +56,9 @@ type TypeVariation struct {
 	Functions   TypeVariationFunctions
 }
 
-type Argument interface {
+type FuncParameter interface {
 	toTypeString() string
-	argumentMarker() // unexported marker method
```
jacques-n, line 60: why remove the marker? We should keep to encapsulate the set of valid items.

extensions/variants.go
```diff
@@ -72,33 +72,38 @@ func validateType(arg Argument, actual types.Type, idx int, nullHandling Nullabi
...
+	funcParameters FuncParameterList, variadic *VariadicBehavior, argumentTypes []types.Type) (types.Type, error) {
+	if len(funcParameters) != len(argumentTypes) {
 		if variadic == nil {
 			return nil, fmt.Errorf("%w: mismatch in number of arguments provided. got %d, expected %d",
-				substraitgo.ErrInvalidExpr, len(actualTypes), len(paramTypeList))
+				substraitgo.ErrInvalidExpr, len(argumentTypes), len(funcParameters))
 		}
 
-		if !variadic.IsValidArgumentCount(len(actualTypes) - len(paramTypeList) - 1) {
+		if !variadic.IsValidArgumentCount(len(argumentTypes) - len(funcParameters) - 1) {
```
jacques-n, line 87: Is this code right? It looks wrong.

types/types.go
```diff
@@ -689,7 +781,7 @@ func (s *PrimitiveType[T]) MatchWithoutNullability(ot Type) bool {
 	return false
 }
 
-func (s *PrimitiveType[T]) ReturnType() (Type, error) {
+func (s *PrimitiveType[T]) ReturnType([]FuncDefArgType, []Type) (Type, error) {
```
jacques-n, line 784: not sure what this signature change is throughout the patch.

types/parameterized_list_type_test.go
```diff
@@ -7,39 +7,39 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/substrait-io/substrait-go/types"
+	. "github.com/substrait-io/substrait-go/types"
```
jacques-n, line 10: why did you add this cosmetic changes in this patch in multiple places?

types/types.go
```diff
@@ -431,6 +431,98 @@ type (
 	}
 )
 
+type EnumType struct {
```
jacques-n, line 434: docstring

types/types.go
scgkiran: Raised PR #80 for return type derivation

EpsilonPrime (approved):
Some of the changed lines have no coverage but that's not related to this name change PR.

extensions/simple_extension.go
jacques-n: it doesn't accomplish the same. The unexported method guarantees that others can't implement additional implementations outside the package.

scgkiran: Restored the marker method. 
Even without the marker method, this interface has one other unexported method(toTypeString).

extensions/variants.go
scgkiran: forgot to mention this is fixed added some testcases also.

jacques-n:
This should have a breaking change commit message, right?

scgkiran:
> This should have a breaking change commit message, right?

Updated the PR title and description.

PR #83, opened 2024-12-05
feat: update to latest core Substrait

EpsilonPrime:
This PR pulls in the latest testcase changes from Substrait core.

It also modifies the signature of ParseTestCaseFileFromFS so it can additionally be used on other filesystems (including os.DirFS).


zeroshade (approved):
LGTM

PR #84, opened 2024-12-06
feat: avoid emit in remap when unnecessary

EpsilonPrime:


jacques-n (approved):
LGTM. Thanks @EpsilonPrime

PR #87, opened 2024-12-13
fix: avoid duplication of variants in registry when function name is overloaded

scgkiran:
this happens with count_star, where same substrait function name is used for two different functions in the extension


zeroshade (approved):
LGTM

PR #90, opened 2024-12-16
feat: add func type in testcase

scgkiran:


testcases/parser/visitor.go
```diff
@@ -86,9 +98,14 @@ func (v *TestCaseVisitor) VisitScalarFuncTestGroup(ctx *baseparser.ScalarFuncTes
 func (v *TestCaseVisitor) VisitAggregateFuncTestGroup(ctx *baseparser.AggregateFuncTestGroupContext) interface{} {
 	groupDesc := v.Visit(ctx.TestGroupDescription()).(string)
 	groupTestCases := make([]*TestCase, 0, len(ctx.AllAggFuncTestCase()))
+	if v.testFuncType != AggregateFuncType {
+		v.ErrorListener.ReportVisitError(fmt.Errorf("expected %v test case, got aggregate", v.testFuncType))
```
EpsilonPrime, line 102: Can you ever get window?

scgkiran: currently grammar doesn't support window func testcases. We have to modify this when we add support for window functions.

EpsilonPrime: This specific check is checking to see that the function type is aggregate and then complains that it's not aggregate.  It seems like it is expecting an aggregate test case and that the got is the case that needs to have %v.

EpsilonPrime: Oh, this is backwards because we're in an aggregate.  Could we point to where the file's type was printed to make the error message more clear?

scgkiran: `v.functionType` is the expected testcase type based on header in the test file. In VisitVersion we save this.

`VisitAggregateFuncTestGroup`  tells us that we got an aggregate function.

scgkiran: modified the error message

EpsilonPrime (approved):

PR #91, opened 2024-12-16
feat: get scalar function invocation object from testcase

scgkiran:


testcases/parser/nodes.go
```diff
@@ -43,6 +45,77 @@ type TestCase struct {
...
+		if i != 0 {
+			b.WriteByte('_')
+		}
+		b.WriteString(a.ColumnType.ShortString())
+	}
+	return b.String()
+}
+
+func (tc *TestCase) signatureKey() string {
+	if tc.FuncType == ScalarFuncType {
```
EpsilonPrime, line 85: Consider using a switch statement here so that when support for window is added it is handled instead of returning aggregate.

testcases/parser/nodes.go
```diff
@@ -43,6 +45,77 @@ type TestCase struct {
...
+	return tc.aggregateSignatureKey()
+}
+
+func (tc *TestCase) CompoundFunctionName() string {
+	return tc.FuncName + ":" + tc.signatureKey()
+}
+
+func (tc *TestCase) ID() extensions.ID {
+	baseURI := tc.BaseURI
+	if !strings.HasPrefix(baseURI, "https") || !strings.HasPrefix(baseURI, "http") {
```
EpsilonPrime, line 97: This check is being used to check that the URI is not a slash-prefixed file.  So it could check for the prefix being a leading slash.  Otherwise a check for a leading slash could be useful before appending it onto the Substrait default address.

EpsilonPrime (approved):

PR #94, opened 2024-12-18
feat: AggregateRel should use grouping references

srikrishnak:


plan/builders.go
```diff
@@ -75,10 +75,13 @@ type Builder interface {
 	ProjectRemap(input Rel, remap []int32, exprs ...expr.Expression) (*ProjectRel, error)
 	// Deprecated: Use AggregateColumns(...).Remap() instead.
 	AggregateColumnsRemap(input Rel, remap []int32, measures []AggRelMeasure, groupByCols ...int32) (*AggregateRel, error)
+	// Deprecated: Use Aggregate(...) instead.
 	AggregateColumns(input Rel, measures []AggRelMeasure, groupByCols ...int32) (*AggregateRel, error)
 	// Deprecated: Use AggregateExprs(...).Remap() instead.
 	AggregateExprsRemap(input Rel, remap []int32, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error)
+	// Deprecated: Use Aggregate(...) instead.
 	AggregateExprs(input Rel, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error)
+	Aggregate(input Rel, measures []AggRelMeasure, groupingExpressions []expr.Expression, groupings [][]uint32) (*AggregateRel, error)
```
EpsilonPrime, line 84: The design of the new grouping expressions allows one to refer to the same grouping expression without requiring the expression to be copied.  Ideally this would be something that could be hidden under the covers.  If it isn't, it is totally valid to duplicate the expression each time (constructing a new grouping id each time).

An API which hid the grouping references implementation details would be ideal and could look like AggregateExprs.

plan/builders.go
```diff
@@ -75,10 +75,13 @@ type Builder interface {
 	ProjectRemap(input Rel, remap []int32, exprs ...expr.Expression) (*ProjectRel, error)
 	// Deprecated: Use AggregateColumns(...).Remap() instead.
 	AggregateColumnsRemap(input Rel, remap []int32, measures []AggRelMeasure, groupByCols ...int32) (*AggregateRel, error)
+	// Deprecated: Use Aggregate(...) instead.
 	AggregateColumns(input Rel, measures []AggRelMeasure, groupByCols ...int32) (*AggregateRel, error)
 	// Deprecated: Use AggregateExprs(...).Remap() instead.
 	AggregateExprsRemap(input Rel, remap []int32, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error)
+	// Deprecated: Use Aggregate(...) instead.
 	AggregateExprs(input Rel, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error)
```
EpsilonPrime, line 83: Optional:  Is a Grouping something we could provide a builder for instead of requiring the caller to construct a slice of expression slices?

plan/builders.go
srikrishnak: In that case, I won't deprecate AggregateExprs and remove the newly added Aggregate()

plan/builders.go
EpsilonPrime: If I was using the interface I'd prefer to just call:

`func (b *builder) Aggregate(input Rel, measures []AggRelMeasure, groupingExpressions []expr.Expression) (*AggregateRel, error) {
`

Then Aggregate could either:

1.  simply assign ascending values to the grouping references for each expression 1:1 (not gaining any benefit of the identical expressions being able to be deduped and other optimization savings)
2.  assign values to the grouping references for each expression.  If an expression duplicates one seen previously then it isn't added to the expression list and a previous grouping reference value is used instead.

The notion of duplication for the second approach could mark equivalence incorrectly.  (For instance, perhaps marking a literal 1 as identical to another one even though it is of a different unit.). However I'm not sure it matters.

srikrishnak: Adding to your suggestion a bit. How about we do introduce more API on AggregateRel like below?

```
func (b *builder) Aggregate(input Rel, measures []AggRelMeasure, groupingExpressions []expr.Expression) (*AggregateRel, error)

func (ar *AggregateRel) addCube(groupingReferences []uint32) 
func (ar *AggregateRel) addRollup(groupingReferences []uint32)
func (ar *AggregateRel) addGroupingSet(groupingReferences []uint32)
```

EpsilonPrime: I do like the idea of adding a builder to ease construction.  I would try to avoid having to deal with grouping references unless the builder is going to tell me which reference a given expression is.  Perhaps we can return the references as part of adding the expressions?

```
func (arb *AggregateBuilder) addExpression(expr.Expression) uint32
```

plan/builders.go
```diff
@@ -321,18 +324,53 @@ func (b *builder) AggregateExprsRemap(input Rel, remap []int32, measures []AggRe
...
+		groupingReferences:  groupingReferences,
+		groupingExpressions: groupingExpressions,
 	}, nil
 }
 
 func (b *builder) AggregateExprs(input Rel, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error) {
 	return b.AggregateExprsRemap(input, nil, measures, groups...)
 }
 
+func (b *builder) Aggregate(input Rel, measures []AggRelMeasure, groupingExpressions []expr.Expression, groupingReferences [][]uint32) (*AggregateRel, error) {
```
EpsilonPrime, line 341: Should this be part of the interface?

plan/builders.go
```diff
@@ -286,11 +287,13 @@ func (b *builder) AggregateColumnsRemap(input Rel, remap []int32, measures []Agg
 		}
 	}
 
+	groupingExpressions, groupingReferences := groupingExprs(exprs)
```
EpsilonPrime, line 290: We could ditch the Remap versions since they've been deprecated for a while (hehe, for at least two major versions) to avoid needing passing these changes along.

plan/builders.go
srikrishnak: removed.

plan/builders.go
```diff
@@ -73,11 +73,8 @@ type Builder interface {
 	Project(input Rel, exprs ...expr.Expression) (*ProjectRel, error)
 	// Deprecated: Use Project(...).Remap() instead.
 	ProjectRemap(input Rel, remap []int32, exprs ...expr.Expression) (*ProjectRel, error)
-	// Deprecated: Use AggregateColumns(...).Remap() instead.
-	AggregateColumnsRemap(input Rel, remap []int32, measures []AggRelMeasure, groupByCols ...int32) (*AggregateRel, error)
+	// Deprecated: Use Aggregate(...) instead.
```
EpsilonPrime, line 76: AggregateExprs?

EpsilonPrime (approved):

plan/builders.go
srikrishnak: updated. Thanks

PR #95, opened 2024-12-18
feat: get aggregate function invocation object from testcase

scgkiran:


testcases/parser/visitor.go
```diff
@@ -159,7 +159,7 @@ func (v *TestCaseVisitor) VisitMultiArgAggregateFuncCall(ctx *baseparser.MultiAr
 	if ctx.QualifiedAggregateFuncArgs() != nil {
 		args = v.Visit(ctx.QualifiedAggregateFuncArgs()).([]*AggregateArgument)
 	}
-	testcase.FuncName = ctx.Identifier().GetText()
+	testcase.FuncName = strings.ToLower(ctx.Identifier().GetText())
```
EpsilonPrime, line 162: Are function names in Substrait case sensitive because they're effectively part of a URI?  If so we may want to keep the provided case.  If we think it is wrong to have names that aren't in lowercase we could raise an error.

testcases/parser/nodes.go
```diff
@@ -59,6 +59,47 @@ func (tc *TestCase) GetFunctionOptions() []*types.FunctionOption {
...
+func (tc *TestCase) getScalarFuncArgTypes() []types.Type {
+	argTypes := make([]types.Type, len(tc.Args))
+	for i, arg := range tc.Args {
+		argTypes[i] = arg.Type
+	}
+	return argTypes
+}
+
+func (tc *TestCase) getAggregateFuncArgTypes() []types.Type {
+	argTypes := make([]types.Type, len(tc.AggregateArgs))
```
EpsilonPrime, line 71: I've always wondered -- given that both Args and AggregateArgs have the same type why do we have both if only one is populated at a time?

testcases/parser/visitor.go
scgkiran: Reverted it.

EpsilonPrime (approved):

EpsilonPrime (approved):

PR #96, opened 2024-12-20
feat: add api to get local function variant given function invocation

scgkiran:


functions/local_functions_test.go
```diff
@@ -0,0 +1,233 @@
...
+	}{
+		{"avg:fp32", "function variant not found"},
+		{"avg:fp64", ""},
+		{"min:i8", "function variant not found"},
+		{"min:i16", ""},
+		{"min:i32", ""},
+		{"sum:i64", "function variant not found"},
+		{"sum:fp64", ""},
+		{"min:dec", ""},
+		{"min:dec", ""},
```
EpsilonPrime, line 202: Could we add a test with an incompatible option (similar to the test done for scalar above)?

testcases/parser/parse_test.go
```diff
@@ -613,3 +616,57 @@ func TestParseTestCaseFile(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func testGetFunctionInvocation(t *testing.T, tc *TestCase, reg *expr.ExtensionRegistry, registry functions.FunctionRegistry) {
+	switch tc.FuncType {
+	case ScalarFuncType:
+		invocation, err := tc.GetScalarFunctionInvocation(reg, registry)
+		require.NoError(t, err, "GetScalarFunctionInvocation failed with error in test case: %s", tc.CompoundFunctionName())
```
EpsilonPrime, line 651: Is there a test case or documentation somewhere that shows what CompoundFunctionName does?  All I can gather from here is that it doesn't return an error sometimes.

testcases/parser/parse_test.go
```diff
@@ -613,3 +616,57 @@ func TestParseTestCaseFile(t *testing.T) {
...
+		require.Equal(t, tc.ID().URI, invocation.ID().URI)
+	case AggregateFuncType:
+		invocation, err := tc.GetAggregateFunctionInvocation(reg, registry)
```
EpsilonPrime, line 654: Is there a CompoundFunctionName() for aggregates?

testcases/parser/parse_test.go
scgkiran: It is similar to FunctionVariant.[CompoundName](https://github.com/substrait-io/substrait-go/blob/8e12e1e1ae704fb6235e803b9b6586504dc48c58/extensions/variants.go#L334-L334). It is a string with name and argTypes.

Added a test for TestCase.CompoundName. Earlier I was testing with ID()

testcases/parser/parse_test.go
scgkiran: yes, fixed the error message in the assert.

EpsilonPrime (approved):

PR #98, opened 2025-01-02
fix: interval type names

scgkiran:


scgkiran:
Aligning the names according to the docs.

https://substrait.io/extensions/#function-signature-compound-names
https://substrait.io/types/type_classes/

EpsilonPrime (commented):
This change looks good.

However we should fix the build system to stop advancing major versions on every breaking change since doing so requires an update to the versioning to reference the project before we commit this change.

EpsilonPrime (approved):
As discussed externally this is bugfix and not a breaking change so it should not pose an issue.

PR #100, opened 2025-01-07
feat: add decimal literal to string method

EpsilonPrime:


scgkiran (approved):

expr/literals.go
```diff
@@ -406,12 +406,34 @@ type ProtoLiteral struct {
 	Type  types.Type
 }
 
+func DecimalLiteralToString(lit *ProtoLiteral) (string, error) {
```
jacques-n, line 409: I'm inclined to update String() to return just the value and create a TypeString() or similar that returns the value + the type (using String() for value). Users having to find this kind of method is much harder than if it hangs off the struct they are trying to work with.

zeroshade: I agree with @jacques-n here, I'd prefer for functions like this to hang on the struct itself via the String() method and/or another method

EpsilonPrime: I started by trying to update String().  However the current String conversion which includes the type is implemented in many places.  The most obvious consumer is string_test.go where all of the other types (not just Decimal literals) include the type.

Similarly I considered adding a new method to ProtoLiteral but decided not to as part of this PR as the right fix would be to add a new method to all Literal implementations (or at a minimum also add support for all of the other ProtoLiterals).  I suggest adding this function for now and do the requisite refactoring to add a method everywhere later.

EpsilonPrime: The other problem is that this string method can fail because it does checks on the data (so it returns an error).  The error checks can be removed if we assume that the data type was constructed properly (either from an external proto source, a local NewDecimalFrom...() call, or directly via the public struct).

jacques-n: Neither of the two things you discounted seem like large efforts and seem preferable to introducing a public api that we plan on removing.

EpsilonPrime: Moved the new computation to be a method on ProtoLiteral.  The other Literal implementations can be updated later on and then the new method can be added to the Literal interface.

rymurr (approved):

jacques-n (approved):
One thought I had was whether we should add this string to decimal as a literal construction method. I believe we have a builder or api for literal construction, right?

EpsilonPrime:
> One thought I had was whether we should add this string to decimal as a literal construction method. I believe we have a builder or api for literal construction, right?

Yes, that's useful.  There is NewDecimalFromString (literal/utils.go) which utilizes this method.

PR #102, opened 2025-01-09
feat: add ValueString() to all literal types

EpsilonPrime:
Adds the functionality QueryString() functionality requested in #100 .

### DIFF
--- a/expr/literals.go
+++ b/expr/literals.go
@@ -21,7 +21,7 @@ import (
 // via ==
 type PrimitiveLiteralValue interface {
 	bool | int8 | int16 | ~int32 | ~int64 |
-		float32 | float64 | ~string
+	float32 | float64 | ~string
 }
 
 type nestedLiteral interface {
@@ -522,7 +522,7 @@ func getNullability(nullable bool) types.Nullability {
 
 type newPrimitiveLiteralTypes interface {
 	bool | int8 | int16 | ~int32 | ~int64 |
-		float32 | float64 | string
+	float32 | float64 | string
 }
 
 func NewPrimitiveLiteral[T newPrimitiveLiteralTypes](val T, nullable bool) Literal {
@@ -634,9 +634,9 @@ func NewFixedBinaryLiteral(val types.FixedBinary, nullable bool) *ByteSliceLiter
 
 type allLiteralTypes interface {
 	PrimitiveLiteralValue | nestedLiteral | MapLiteralValue |
-		[]byte | types.UUID | types.FixedBinary | *types.IntervalYearToMonth |
-		*types.IntervalDayToSecond | *types.VarChar | *types.Decimal | *types.UserDefinedLiteral |
-		*types.PrecisionTimestamp | *types.PrecisionTimestampTz
+	[]byte | types.UUID | types.FixedBinary | *types.IntervalYearToMonth |
+	*types.IntervalDayToSecond | *types.VarChar | *types.Decimal | *types.UserDefinedLiteral |
+	*types.PrecisionTimestamp | *types.PrecisionTimestampTz
 }
 
 func NewLiteral[T allLiteralTypes](val T, nullable bool) (Literal, error) {
--- a/expr/literals.go
+++ b/expr/literals.go
@@ -6,9 +6,14 @@ package expr
 
 import (
 	"bytes"
+	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strings"
+	"time"
 
+	"cloud.google.com/go/civil"
+	"github.com/google/uuid"
 	substraitgo "github.com/substrait-io/substrait-go/v3"
 	"github.com/substrait-io/substrait-go/v3/proto"
 	"github.com/substrait-io/substrait-go/v3/types"
@@ -84,6 +89,8 @@ type Literal interface {
 	ToProto() *proto.Expression
 	ToProtoLiteral() *proto.Expression_Literal
 	Visit(VisitFunc) Expression
+	// ValueString returns a string representation of the literal's value only.
+	ValueString() string
 }
 
 // A NullLiteral is a typed null, so it just contains its type
@@ -100,6 +107,9 @@ func (*NullLiteral) IsScalar() bool { return true }
 func (*NullLiteral) isRootRef() {}
 func (n *NullLiteral) String() string {
 	return "null(" + n.Type.String() + ")"
+}
+func (n *NullLiteral) ValueString() string {
+	return "null"
 }
 
 func (n *NullLiteral) GetType() types.Type { return n.Type }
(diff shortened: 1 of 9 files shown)


expr/literals.go
```diff
@@ -84,6 +89,8 @@ type Literal interface {
 	ToProto() *proto.Expression
 	ToProtoLiteral() *proto.Expression_Literal
 	Visit(VisitFunc) Expression
+	// ValueString returns a string representation of the literal's value only.
```
zeroshade, line 92: we should probably expand this description a little bit in how we explain that it'll return timestamps/dates/etc. instead of the raw underlying value

expr/literals.go
```diff
@@ -144,7 +154,25 @@ type PrimitiveLiteral[T PrimitiveLiteralValue] struct {
...
+	case types.Time:
+		t := time.UnixMicro(int64(x))
+		return t.UTC().Format(time.TimeOnly)
+	case types.Timestamp:
+		t := time.UnixMicro(int64(x))
+		return t.UTC().Format(time.RFC3339)
+	case types.TimestampTz:
+		t := time.UnixMicro(int64(x))
+		return t.UTC().Format(time.RFC3339)
+	}
```
zeroshade, line 174: I wonder if we should add `String()` methods to the `types.Date`/`types.Time`/etc. types directly which then implements these conversions rather than having to perform the type switch here.

expr/literals.go
```diff
@@ -220,6 +248,17 @@ func (*NestedLiteral[T]) isRootRef() {}
...
 	return fmt.Sprintf("%s(%v)", t.Type, t.Value)
 }
+func (t *NestedLiteral[T]) ValueString() string {
+	switch x := any(t.Value).(type) {
+	case ListLiteralValue:
+		var items []string
+		for _, item := range x {
+			items = append(items, item.ValueString())
+		}
+		return fmt.Sprintf("[%s]", strings.Join(items, ", "))
```
zeroshade, line 258: why not use `strings.Builder` and build the result directly instead of using the intermediate slice?

expr/literals.go
```diff
@@ -408,19 +465,85 @@ type ProtoLiteral struct {
...
+		case types.PrecisionEMinus7Seconds:
+			tm = time.Unix(x/10000000, x%10000000*100)
+		case types.PrecisionEMinus8Seconds:
+			tm = time.Unix(x/100000000, x%100000000*10)
+		case types.PrecisionNanoSeconds:
+			tm = time.Unix(x/1000000000, x%1000000000)
+		default:
+			panic("unsupported precision")
+		}
+		return tm.UTC().Format("2006-01-02 15:04:05.999999999")
```
zeroshade, line 495: this and the part below it are identical, can we push this to a helper function and remove the duplication?

expr/literals.go
EpsilonPrime: I do agree we do need to move this calculation out and the types would be one such place.  But that would mean passing the literal internals to the type class for conversion.  I'd rather see us have literal implementations for all types (perhaps loose wrappers around the existing generics) to hang that calculation on.

expr/literals.go
EpsilonPrime: done

expr/literals.go
EpsilonPrime: done.  can tweak more if desired

expr/literals.go
EpsilonPrime: I used to use that style (building up the string and adding the separator as I visited the elements).  I switched to this style to avoid the is the first element check needed to prevent adding  the duplicate separator.  It may not perform as well but the two activities -- gathering the data and managing the presentation -- are separate.

That said I have no strong preference.  If you'd like it the other way I can make that change.

expr/literals.go
zeroshade: I'm confused, what literal internals would be being passed here? I'm proposing being able to do `return t.Value.String()` 

We could add `String()` to the Type interface, and then implement it for the types, or perhaps just a subset of the types and this could be: 

```go
if s, ok := t.Value.(fmt.Stringer); ok {
     return s.String()
}

return fmt.Sprintf("%v", t.Value)
```

zeroshade: Another alternative might be adding a `ToTime` method on these. For example see https://github.com/apache/iceberg-go/blob/af5a9908a83444ca13b19d24a27200bc7323a2b6/types.go#L519

expr/literals.go
zeroshade: Honestly I mostly agree with you, the only time it would matter is if the number of items is particularly large and they are doing this frequently. 

I say leave it like this for now untill someone comes and complains about memory or performance and shows a use case that really needs it

expr/literals.go
EpsilonPrime: t.Value is an int64 for the time fields, the type is t.Type.  An approach putting the ToTime on the type would result in something like `t.Type.ToTime(t.Value)`.  Note that Iceberg's types are the equivalent of our literals which is where I'd rather see the method.  I suppose we could create a new TimeLiteral interface and add a generic ToTime method but we'd still have the type case switch.

zeroshade: Well, iceberg has literals also, and you can see the way the ToTime is used: https://github.com/apache/iceberg-go/blob/af5a9908a83444ca13b19d24a27200bc7323a2b6/literals.go#L642

Though those literals aren't fully generic like substrait's here. 

That said, if we had the TimeLiteral interface, then we wouldn't need the type switch, it could just be the type assertion example: 

```go
if tm, ok := t.Value.(TimeLiteral); ok {
         return tm.Format(....)
}

return fmt.Sprintf(...)
```

Which now that I see it, I rather like it lol

zeroshade: Keep in mind, I'm suggesting to put the ToTime on `types.Date`, not `types.DateType`

expr/literals.go
```diff
@@ -406,21 +463,68 @@ type ProtoLiteral struct {
 	Type  types.Type
 }
 
+func timeFromPrecisionUnits(units int64, precision types.TimePrecision) time.Time {
```
zeroshade, line 466: I wonder if we could hang a `ToTime(value)` method on `types.TimePrecision` or we could hang `ToTime(precision)` on `types.TimestampTz` and friends....

expr/literals.go
```diff
@@ -406,21 +463,68 @@ type ProtoLiteral struct {
...
+	}
+	return tm
+}
+
 func (t *ProtoLiteral) ValueString() string {
 	switch literalType := t.Type.(type) {
-	case *types.PrecisionTimestampType, *types.PrecisionTimestampTzType:
-		return fmt.Sprintf("%d", t.Value)
+	case *types.PrecisionTimestampType:
+		x, _ := t.Value.(int64)
```
zeroshade, line 498: If you're ignoring the second param here, why bother having this as a separate line? 

Just do `tm := timeFromPrecisionUnits(t.Value.(int64), literalType.Precision)`?

expr/literals.go
EpsilonPrime: done

expr/literals.go
EpsilonPrime: done

expr/literals.go
EpsilonPrime: done

expr/literals.go
```diff
@@ -220,6 +235,17 @@ func (*NestedLiteral[T]) isRootRef() {}
...
+func (t *NestedLiteral[T]) ValueString() string {
+	switch x := any(t.Value).(type) {
+	case ListLiteralValue:
+		var items []string
+		for _, item := range x {
+			items = append(items, item.ValueString())
+		}
+		return fmt.Sprintf("[%s]", strings.Join(items, ", "))
+	}
+	return fmt.Sprintf("%v", t.Value)
```
zeroshade, line 247: Could we add a test that hits this so the codecov warning goes away?

types/types.go
```diff
@@ -1546,3 +1560,69 @@ func (r RecordType) Types() []Type {
 func (r RecordType) Concat(other RecordType) RecordType {
 	return RecordType{types: append(r.Types(), other.Types()...)}
 }
+
+func (d Date) ToTimeString() string {
+	date := civil.Date{Year: 1970, Month: time.January, Day: 1}
+	date = date.AddDays(int(d))
+	return date.String()
```
zeroshade, line 1567: what's the benefit to this over just using `time` like we were doing before?

zeroshade (approved):
Overall looks great to me, just a couple nit-picks

types/types.go
EpsilonPrime: Adding days to time.t can be messed up by daylight savings time.  Civil date's base type is days so it isn't affected.

expr/literals.go
EpsilonPrime: done

EpsilonPrime: ok, not done.  Turns out that map literals go through a different path.  I'm not sure how to get it to visit this line.

PR #107, opened 2025-01-20
fix: tweak the not found in dialect message

EpsilonPrime:


vqmarkman:
PR addresses this problem: https://sundeck.atlassian.net/browse/SD-9378

functions/dialect.go
```diff
@@ -97,6 +97,18 @@ type mapAndSlice[V extensions.FunctionVariant] struct {
 	variantsSlice []V
 }
 
+func nameForInvocationType(invocationType interface{}) string {
```
scgkiran, line 100: is it better to add this to `localFunctionVariant` interface?

EpsilonPrime: Done.  I originally thought about placing it here but didn't like the type methods being mixed with the instance methods.  I gave it a clear name to make the distinction more clear that the method doesn't care about its value.

scgkiran (approved):

PR #110, opened 2025-01-23
fix: set nullability correctly in testcase parser

scgkiran:


testcases/parser/parse_test.go
```diff
@@ -155,17 +155,18 @@ func TestParseTestWithVariousTypes(t *testing.T) {
...
 		{testCaseStr: "f12('P10DT6M7S'::interval_day, 5::i64) = 'P10DT11M7S'::interval_day", expTestStr: "f12('P10DT6M7S'::interval_day<0>, 5::i64) = 'P10DT11M7S'::interval_day<0>"},
-		{testCaseStr: "concat('abcd'::varchar<9>, Null::str) [null_handling:ACCEPT_NULLS] = Null::str", expTestStr: "concat('abcd'::varchar<9>, null::string) [null_handling:ACCEPT_NULLS] = null::string"},
-		{testCaseStr: "concat('abcd'::varchar<9>, null::string) [null_handling:ACCEPT_NULLS] = null::string"},
-		{testCaseStr: "concat('abcd'::vchar<9>, 'ef'::varchar<9>) = Null::vchar<9>", expTestStr: "concat('abcd'::varchar<9>, 'ef'::varchar<9>) = null::varchar<9>"},
-		{testCaseStr: "concat('abcd'::vchar<9>, 'ef'::fixedchar<9>) = Null::fchar<9>", expTestStr: "concat('abcd'::varchar<9>, 'ef'::fixedchar<9>) = null::fixedchar<9>"},
-		{testCaseStr: "concat('abcd'::fbin<9>, 'ef'::fixedbinary<9>) = Null::fbin<9>", expTestStr: "concat('0x61626364'::fixedbinary<9>, '0x6566'::fixedbinary<9>) = null::fixedbinary<9>"},
+		{testCaseStr: "concat('abcd'::varchar<9>, Null::str) [null_handling:ACCEPT_NULLS] = Null::str", expTestStr: "concat('abcd'::varchar<9>, null::string?) [null_handling:ACCEPT_NULLS] = null::string?"},
```
EpsilonPrime, line 164: Implicitly any literal which is null is going to be nullable. I suggest having one test case that illustrates implicit nullability of null literals in addition to the explicit ones.

testcases/parser/parse_test.go
```diff
@@ -611,12 +627,16 @@ func TestParseAggregateTestWithVariousTypes(t *testing.T) {
...
 		{testCaseStr: `DEFINE t1(fp32, fp32) = ((20, 20), (-3, -3), (1, 1), (10,10), (5,5))
 count_star() = 1::fp64`, expTestStr: "((20, 20), (-3, -3), (1, 1), (10, 10), (5, 5)) count_star() = 1::fp64"},
-		{testCaseStr: "f20(('abcd', 'ef')::fchar?<9>) = Null::fchar<9>", expTestStr: "f20(('abcd', 'ef')::fixedchar?<9>) = null::fixedchar<9>"},
-		{testCaseStr: "f20(('abcd', 'ef')::fixedchar<9>) = Null::fchar<9>", expTestStr: "f20(('abcd', 'ef')::fixedchar<9>) = null::fixedchar<9>"},
-		{testCaseStr: "f20(('abcd', 'ef', null)::vchar?<9>) = Null::vchar<9>", expTestStr: "f20(('abcd', 'ef', null)::varchar?<9>) = null::varchar<9>"},
-		{testCaseStr: "f20(('abcd', 'ef')::varchar<9>) = Null::vchar<9>", expTestStr: "f20(('abcd', 'ef')::varchar<9>) = null::varchar<9>"},
-		{testCaseStr: "f20(('abcd', 'ef')::fbin<9>) = Null::fbin<9>", expTestStr: "f20(('abcd', 'ef')::fixedbinary<9>) = null::fixedbinary<9>"},
-		{testCaseStr: "f20(('abcd', 'ef')::fixedbinary?<9>) = Null::fixedbinary<9>", expTestStr: "f20(('abcd', 'ef')::fixedbinary?<9>) = null::fixedbinary<9>"},
+		{testCaseStr: `DEFINE t1(varchar<5>) = (('cat'), ('bat'), ('rat'), (null))
```
EpsilonPrime, line 630: Should this be `varchar<5>?`?

testcases/parser/parse_test.go
```diff
@@ -611,12 +627,16 @@ func TestParseAggregateTestWithVariousTypes(t *testing.T) {
...
+count_star() = 1::fp64`, expTestStr: "(('cat'), ('bat'), ('rat'), (null)) count_star() = 1::fp64"},
```
EpsilonPrime, line 631: This PR addresses parser changes.  In a later PR where should we deal with type mismatches such as null trying to be placed in a non-nullable aggregate data column?

testcases/parser/visitor.go
```diff
@@ -711,7 +727,12 @@ func (v *TestCaseVisitor) VisitLiteral(ctx *baseparser.LiteralContext) interface
 	}
 
 	if ctx.NullLiteral() != nil {
-		return expr.NewNullLiteral(v.getLiteralTypeInContext())
+		nullType := v.getLiteralTypeInContext()
+		if nullType == nil {
+			// Use a dummy type for null literal. This happens in AggregateFuncCall context, where type is not set
```
EpsilonPrime, line 732: Could the aggregate type be pulled (or pushed down) from the parent's context?

testcases/parser/parse_test.go
scgkiran: The type is associated with column argument, in this case number of arguments is zero. 
Added another test with one argument

testcases/parser/visitor.go
scgkiran: 2 out of 3 formats we push down the type. In one case, code is getting messy to push down for very little gain. so I am fixing the type as post processing.

testcases/parser/parse_test.go
EpsilonPrime: Indeed it is, apparently I missed that when I scanned the tests for nulls without question marks.

testcases/parser/parse_test.go
EpsilonPrime: It looks like null is one of the possible values but varchar<5> isn't nullable.  That seems like something that should cause an error.

testcases/parser/visitor.go
EpsilonPrime: It could be one way you could do type checking in place. The type checking also can be done where you post processing (although you might not have as much information about where the offending tokens are).

scgkiran: I see we don't have type checks in this case. Below testcase should fail, but it doesn't.
```
DEFINE t1(fp32) = ((20), ('abc'), (1), (10), (5))
corr(t1.col0) = 1::fp64
```

Is it ok if I take this up in another PR?

testcases/parser/parse_test.go
scgkiran: That needs, substrait PR for grammar changes to be merged, to handle all cases. Currently we have `?` support only for composite types.

EpsilonPrime (approved):

PR #111, opened 2025-01-24
fix: fix testcase literals type to have the requested type parameters

scgkiran:


scgkiran:
This PR is stacked on #110

expr/literals.go
```diff
@@ -625,6 +677,69 @@ func (t *ProtoLiteral) ToProtoFuncArg() *proto.FunctionArgument {
...
+	if _, ok := literal.GetType().(*types.VarCharType); !ok {
+		return nil, fmt.Errorf("literal type is not varchar")
+	}
+	if len(literal.Value.(string)) > int(fixedType.GetLength()) {
+		return nil, fmt.Errorf("varchar litearl value length is greater than type length")
+	}
+	return &ProtoLiteral{Value: literal.Value, Type: fixedType}, nil
+}
+
+func NewPrecisionTimestampWithType(literal *ProtoLiteral, ptsType *types.PrecisionTimestampType) (Literal, error) {
```
EpsilonPrime, line 703: This feels difficult to use as one needs to construct a type object (which isn't easy either) in order to use it.  Since this is a convenience function perhaps just provide the parameter instead?

testcases/parser/nodes.go
```diff
@@ -78,6 +78,25 @@ func (c *CaseLiteral) AsAggregateArgumentString() string {
 	return c.Value.ValueString() + "::" + c.Type.String()
 }
 
+// change the type of the literal to the given type
+func (c *CaseLiteral) fixTypeInTheLiteral() error {
```
EpsilonPrime, line 82: It's not clear to me looking at this function what fixing does.  The same goes for the error message.  It looks like the parser is treating the type declaration as a cast operation.

expr/literals.go
scgkiran: These are convenience functions, current usage has the type available with the caller. Made them local to the module.

EpsilonPrime (approved):

PR #114, opened 2025-01-30
feat: Add AggregateRel.ToBuilder()

EpsilonPrime:


jacques-n (commented):
If I use ToBuilder(), how do I then override measures and/or input?

EpsilonPrime:
> If I use ToBuilder(), how do I then override measures and/or input?

Hmm, true.  The builder doesn't have a way to get rid of the existing measures.  The cleanest way would be to provide a clear method.  The builder doesn't have a way to change the input but AggregateRel.Copy allows the input to change.  I propose adding:

AggregateRelBuilder.ReplaceInput
AggregateRelBuilder.ClearMeasures
AggregateRelBuilder.ClearGroupings (affects both grouping references and grouping expressions)

jacques-n:
What about replacing measures? (That's my specific use case.)

EpsilonPrime:
> What about replacing measures? (That's my specific use case.)

If I implement `AggregateRelBuilder.ClearMeasures` you could first clear them all and then add all of the desired measures (including old ones) next.  If you want to directly replace the third measure I could add `AggregateRelBuilder.SetMeasure(index int, measure)`.

EpsilonPrime:
> > What about replacing measures? (That's my specific use case.)
> 
> If I implement `AggregateRelBuilder.ClearMeasures` you could first clear them all and then add all of the desired measures (including old ones) next. If you want to directly replace the third measure I could add `AggregateRelBuilder.SetMeasure(index int, measure)`.

@jacques-n Is Clear() sufficient for your purposes or do would you like SetMeasure() as well?

EpsilonPrime:
I ended up adding just the clear methods.  @scgkiran could you review please?  Thanks!

scgkiran (approved):

PR #118, opened 2025-02-12
feat: load the default extension collection lazily

scgkiran:


extensions/extension_mgr.go
```diff
@@ -20,45 +22,72 @@ type AdvancedExtension = extensions.AdvancedExtension
...
+	c, err := GetDefaultCollectionWithError()
+	if err != nil {
+		panic(err)
+	}
+	return c
+}
+
+// GetDefaultCollectionWithError returns a Collection that is loaded with the default Substrait extension definitions.
+func GetDefaultCollectionWithError() (*Collection, error) {
+	collectionOnce.Do(func() {
```
EpsilonPrime, line 42: If you use sync.OnceValues you can have it return the data and the error together without needing to keep track of them separately (this obviates the need for all three private global variables).

extensions/extension_mgr.go
```diff
@@ -20,45 +22,72 @@ type AdvancedExtension = extensions.AdvancedExtension
...
-// definitions. Not all files are currently parsable.
-// Parser needs to enhanced to support all files
-var DefaultCollection Collection
+var (
+	defaultCollection   Collection
+	collectionOnce      sync.Once
+	collectionLoadError error
+)
+
+// GetDefaultCollection returns a Collection that is loaded with the default Substrait extension definitions.
```
EpsilonPrime, line 31: Could we add a suggestion that folks start using the signature that returns an error ("this version is provided for the ease of use of legacy code")?

extensions/extension_mgr.go
```diff
@@ -20,45 +22,72 @@ type AdvancedExtension = extensions.AdvancedExtension
...
+		_ = f.Close()
+	}()
+	fileStat, err := f.Stat()
+	if err != nil {
+		return err, true
+	}
+	fileName := path.Base(fileStat.Name())
+	err = defaultCollection.Load(SubstraitDefaultURIPrefix+ent.Name(), f)
+	if err != nil {
+		if fileName == "unknown.yaml" {
```
EpsilonPrime, line 83: Ah, getting rid of this message is going to be nice.  Consider adding a constant map at the top of the file with a list to make this list easier to manipulate (`testcases.go` does this if you'd like something easy to copy).

EpsilonPrime (commented):
Thanks for cleaning this up!

Suggestion for alternative PR title:
```feat: make the default extension collection load lazily```

extensions/extension_mgr.go
scgkiran: Thanks, the code is more simpler now.

extensions/extension_mgr.go
scgkiran: done

EpsilonPrime (approved):
Looks good, thanks!

PR #120, opened 2025-02-24
feat: add is_null support in testfile parser

scgkiran:
Add nullable parameter to literal creation utils
 Literal creation helpers now include an additional boolean parameter for nullability.


expr/literals.go
```diff
@@ -239,6 +239,19 @@ func (t *PrimitiveLiteral[T]) ToProtoFuncArg() *proto.FunctionArgument {
 func (t *PrimitiveLiteral[T]) Visit(VisitFunc) Expression { return t }
 func (*PrimitiveLiteral[T]) IsScalar() bool               { return true }
 
+func (t *PrimitiveLiteral[T]) WithType(newType types.Type) (Literal, error) {
+	switch newType.(type) {
+	case *types.FixedLenType[types.FixedChar]:
+		if val, ok := any(t.Value).(types.FixedChar); ok {
```
EpsilonPrime, line 245: Add a check for this case?

literal/utils.go
```diff
@@ -14,77 +14,77 @@ import (
 	"github.com/substrait-io/substrait-go/v3/types"
 )
 
-func NewBool(value bool) expr.Literal {
-	return expr.NewPrimitiveLiteral[bool](value, false)
+func NewBool(value bool, nullable bool) expr.Literal {
```
EpsilonPrime, line 17: These changes while straightforward are going to break a lot of existing code.  Two alternates would be to add an additional function with the new behavior (NewNullBool creating a nullable Bool or NewNullableBool with the option similar to what is implemented here).  A third is to actually make this a breaking change which would cause a version bump.  A fourth option would be to add a new method to expr.Literal that would change the nullability to be set.

expr/literals.go
scgkiran: done

literal/utils.go
scgkiran: will mark this as breaking change

EpsilonPrime (approved):

PR #121, opened 2025-02-25
Fix the visit function for Cast so it doesn't erase the cast.

EpsilonPrime:


expr/expressions_test.go
```diff
@@ -418,3 +419,37 @@ func TestRoundTripExtendedExpression(t *testing.T) {
...
+		name            string
+		rewriteFunction func(rex expr.Expression) expr.Expression
+		want            string
+	}
+	testCases := []relationTestCase{
+		{"no change", func(ex expr.Expression) expr.Expression { return ex },
+			"literal:{fp64:12 nullable:true}"},
+		{"changed", func(ex expr.Expression) expr.Expression {
+			lit, err := expr.NewLiteral[float64](16.0, true)
+			if err != nil {
```
anshuldata, line 440: nit: Four line of if can be replaced with using require
```suggestion
			require.NoError(t, err)
```

anshuldata (approved):
lgtm

expr/expressions_test.go
EpsilonPrime: done

PR #124, opened 2025-03-07
feat: implement conversion for recently added types to proto and back

EpsilonPrime:


types/types_test.go
```diff
@@ -87,8 +88,11 @@ func TestTypeRoundtrip(t *testing.T) {
 				&FixedBinaryType{Nullability: n, Length: 45},
 				&IntervalDayType{Nullability: n, Precision: 5},
 				&IntervalDayType{Nullability: n, Precision: 0},
+				compound,
```
scgkiran, line 91: do we really need the local variable for this?

scgkiran (approved):

types/types_test.go
EpsilonPrime: Not needed, originally added to get the pointer, but apparently IntervalCompound needs to be real object.

PR #126, opened 2025-04-07
fix: missing args for ParameterizedListType type resolution

miretskiy:
Pass required information when resolving the return type of parameterized list type.


EpsilonPrime (approved):

vbarua (approved):
Thanks @miretskiy ✨

PR #130, opened 2025-04-17
feat: derive ReturnType for AnyTypes nested in composite types

### LINKED ISSUE #129: [bug] ReturnType derivation for AnyType does not work for composite types
https://github.com/substrait-io/substrait-go/blob/1a2a6ae5fd9986045f69d6ec590875c9af45b88f/types/any_type.go#L55

AnyType's ReturnType method does not handle composite types; 
a function such as `element_at(list<any1>, i64) -> any1` would fail to determine
`any1` as a return type.

The same is true for any composite types.

miretskiy:
Fixes #129

Handle composite types when determining ReturnType for AnyType.

### DIFF
--- a/extensions/variants_test.go
+++ b/extensions/variants_test.go
@@ -18,53 +18,150 @@ import (
 
 func TestEvaluateTypeExpression(t *testing.T) {
 	var (
-		i64Null, _    = parser.ParseType("i64?")
-		i64NonNull, _ = parser.ParseType("i64")
-		strNull, _    = parser.ParseType("string?")
+		// Function definition argument type shortcuts.
+		i64Null, _      = parser.ParseType("i64?")
+		i64NonNull, _   = parser.ParseType("i64")
+		strNull, _      = parser.ParseType("string?")
+		strNonNull, _   = parser.ParseType("string")
+		any1NonNull, _  = parser.ParseType("any1")
+		any1listNonNull = mkFuncArgList(any1NonNull)
+
+		// Few shortcut type definitions.
+		i64TypeReq     = &types.Int64Type{Nullability: types.NullabilityRequired}
+		strTypeReq     = &types.StringType{Nullability: types.NullabilityRequired}
+		i64listNonNull = mkList(i64TypeReq)
 	)
 
 	tests := []struct {
-		name     string
-		nulls    extensions.NullabilityHandling
-		ret      types.FuncDefArgType
-		extArgs  extensions.FuncParameterList
-		args     []types.Type
-		expected types.Type
-		err      string
+		name      string
+		nulls     extensions.NullabilityHandling
+		ret       types.FuncDefArgType
+		extArgs   extensions.FuncParameterList
+		args      []types.Type
+		expected  types.Type
+		expectErr string
 	}{
-		{"defaults", extensions.MirrorNullability, i64NonNull, extensions.FuncParameterList{
-			extensions.ValueArg{Value: &parser.TypeExpression{ValueType: i64Null}}},
-			[]types.Type{&types.Int64Type{Nullability: types.NullabilityNullable}},
-			&types.Int64Type{Nullability: types.NullabilityNullable}, ""},
-		{"arg mismatch", extensions.MirrorNullability, strNull, extensions.FuncParameterList{extensions.ValueArg{Value: &parser.TypeExpression{ValueType: strNull}}},
-			[]types.Type{}, nil, "invalid expression: mismatch in number of arguments provided. got 0, expected 1"},
-		{"missing enum arg", extensions.MirrorNullability, i64Null, extensions.FuncParameterList{
-			extensions.ValueArg{Value: &parser.TypeExpression{ValueType: i64NonNull}}, extensions.EnumArg{Name: "foo"}},
-			[]types.Type{&types.Int64Type{}, &types.Int64Type{}}, nil, "invalid type: arg #1 (foo) should be an enum"},
-		{"discrete null handling", extensions.DiscreteNullability, strNull, extensions.FuncParameterList{
-			extensions.ValueArg{Value: &parser.TypeExpression{ValueType: strNull}}},
-			[]types.Type{&types.StringType{Nullability: types.NullabilityRequired}},
-			nil, "invalid type: discrete nullability did not match for arg #0"},
-		{"mirror", extensions.MirrorNullability, strNull, extensions.FuncParameterList{
-			extensions.ValueArg{Value: &parser.TypeExpression{ValueType: i64NonNull}}, extensions.ValueArg{Value: &parser.TypeExpression{ValueType: i64Null}}},
-			[]types.Type{
-				&types.Int64Type{Nullability: types.NullabilityRequired},
-				&types.Int64Type{Nullability: types.NullabilityRequired}},
-			&types.StringType{Nullability: types.NullabilityRequired}, ""},
-		{"declared output", extensions.DeclaredOutputNullability, strNull, extensions.FuncParameterList{
-			extensions.ValueArg{Value: &parser.TypeExpression{ValueType: strNull}}},
-			[]types.Type{&types.StringType{Nullability: types.NullabilityRequired}},
-			&types.StringType{Nullability: types.NullabilityNullable}, ""},
+		{
+			name:     "defaults",
+			nulls:    extensions.MirrorNullability,
+			ret:      i64NonNull,
+			extArgs:  extensions.FuncParameterList{valArg(i64Null)},
+			args:     []types.Type{&types.Int64Type{Nullability: types.NullabilityNullable}},
+			expected: &types.Int64Type{Nullability: types.NullabilityNullable},
+		},
+		{
+			name:      "arg mismatch",
+			nulls:     extensions.MirrorNullability,
+			ret:       strNull,
+			extArgs:   extensions.FuncParameterList{valArg(strNull)},
+			args:      []types.Type{},
+			expectErr: "invalid expression: mismatch in number of arguments provided. got 0, expected 1",
+		},
+		{
+			name:      "missing enum arg",
+			nulls:     extensions.MirrorNullability,
+			ret:       i64Null,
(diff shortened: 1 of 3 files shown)


miretskiy:
Not sure why coverage reports the new lines were not covered by tests -- they are; just in a different directory through variants resolution.

I have made a style change in the TestEvaluateTypeExpression to use explicitly named struct fields in the
table driven tests (I'm sorry for this; but I had really hard time reading this test; hopefully, even though it's more verbose, it's also easier to read now).

extensions/variants_test.go
```diff
@@ -18,53 +18,147 @@ import (
 
 func TestEvaluateTypeExpression(t *testing.T) {
 	var (
-		i64Null, _    = parser.ParseType("i64?")
-		i64NonNull, _ = parser.ParseType("i64")
-		strNull, _    = parser.ParseType("string?")
+		i64Null, _      = parser.ParseType("i64?")
+		i64TypeReq      = &types.Int64Type{Nullability: types.NullabilityRequired}
+		i64NonNull, _   = parser.ParseType("i64")
```
bvolpato, line 23: it's non intuitive how are those different. perhaps a comment should help

extensions/variants_test.go
```diff
@@ -18,53 +18,147 @@ import (
...
-			[]types.Type{
-				&types.Int64Type{Nullability: types.NullabilityRequired},
-				&types.Int64Type{Nullability: types.NullabilityRequired}},
-			&types.StringType{Nullability: types.NullabilityRequired}, ""},
-		{"declared output", extensions.DeclaredOutputNullability, strNull, extensions.FuncParameterList{
-			extensions.ValueArg{Value: &parser.TypeExpression{ValueType: strNull}}},
-			[]types.Type{&types.StringType{Nullability: types.NullabilityRequired}},
-			&types.StringType{Nullability: types.NullabilityNullable}, ""},
+		{
+			name:     "defaults",
```
bvolpato, line 42: thanks for the new tests and formatting - it looks much cleaner

bvolpato (approved):
This looks like a great improvement, and very clear to follow tests, thanks! 🙇

bvolpato:
lint/codecov don't seem happy - do you want to revisit those checks?

extensions/variants_test.go
miretskiy: Done; sorted this list so that function argument definition types are group together, and underlying types are below.

miretskiy:
> lint/codecov don't seem happy - do you want to revisit those checks?

Pushed any_type_test update to keep codecov happy(er).

types/any_type_test.go
```diff
@@ -19,11 +19,83 @@ func TestAnyType(t *testing.T) {
...
-		{"List<any>", "any", []FuncDefArgType{&AnyType{Name: "any"}}, []Type{&ListType{Type: decP30S9}}, &ListType{Type: decP30S9}, NullabilityNullable, "any?"},
-		{"anyrequired", "any2", []FuncDefArgType{&Int16Type{}, &AnyType{Name: "any2"}}, []Type{&Int16Type{}, &Int64Type{}}, &Int64Type{}, NullabilityRequired, "any2"},
-		{"anyOtherName", "any1", []FuncDefArgType{&AnyType{Name: "any1"}, &Int32Type{}}, []Type{varchar37, &Int32Type{}}, varchar37, NullabilityNullable, "any1?"},
-		{"T name", "T", []FuncDefArgType{&AnyType{Name: "U"}}, []Type{varchar37}, nil, NullabilityNullable, "T?"},
+		{
+			testName:           "any",
+			argName:            "any",
+			parameters:         []FuncDefArgType{&AnyType{Name: "any"}},
+			args:               []Type{decP30S9},
+			expectedReturnType: decP30S9,
```
vbarua, line 27: Would it make sense to rename this field to like `expectedConcreteType`. Calling this `expectedReturnType` is confusing IMO because there is no function call, so there's not really a return type involved in this.

types/any_type_test.go
```diff
@@ -19,11 +19,83 @@ func TestAnyType(t *testing.T) {
...
+			nullability:        NullabilityNullable,
+			expectedString:     "any?",
+		},
+		{
+			testName:           "anyrequired",
+			argName:            "any2",
+			parameters:         []FuncDefArgType{&Int16Type{}, &AnyType{Name: "any2"}},
+			args:               []Type{&Int16Type{}, &Int64Type{}},
+			expectedReturnType: &Int64Type{},
+			nullability:        NullabilityRequired,
```
vbarua, line 37: (not technically a part of your change, potentially something to look at as a follow-up)

What's the point of this `nullability`? In the test runner below:
```go
			anyBase := &AnyType{
				Name:        td.argName,
				Nullability: td.nullability,
			}
			anyType := anyBase.SetNullability(td.nullability)
			require.Equal(t, td.nullability, anyType.GetNullability())
```
we set the `nullability` of the `anyType`, and then check that the nullability of the `anyType` we just call `SetNullability` which seems pointless.

types/any_type_test.go
```diff
@@ -19,11 +19,83 @@ func TestAnyType(t *testing.T) {
...
+				&ParameterizedMapType{Key: &StringType{}, Value: &AnyType{Name: "any1"}},
+			},
+			args: []Type{
+				&Int16Type{},
+				&MapType{Key: &StringType{}, Value: &ListType{Type: &Int64Type{}}},
+			},
+			expectedReturnType: &ListType{Type: &Int64Type{}},
+			nullability:        NullabilityRequired,
+			expectedString:     "any1",
+		},
```
vbarua, line 66: Can you add an explicit check for `any1` in the key position. You've gone the trouble of adding all of these nice tests, might as well be extra thorough.

extensions/variants_test.go
```diff
@@ -18,53 +18,150 @@ import (
...
+			expected: i64TypeReq,
+		},
+		{
+			name:     "element_at(list<any1>, i64) -> any1",
+			nulls:    extensions.DeclaredOutputNullability,
+			ret:      any1NonNull,
+			extArgs:  extensions.FuncParameterList{valArg(any1listNonNull), valArg(i64NonNull)},
+			args:     []types.Type{i64listNonNull, i64TypeReq},
+			expected: i64TypeReq,
+		},
```
vbarua, line 98: This test raises a question for me, which is whether NULLABLE or REQUIRED even make sense to apply to the `any1` type.

Specifically, if the first argument to this was `List<i64?>` would the expectation be:
1. This isn't a valid binding for `element_at` because the `any1` value cannot be null.
2. The return type of this functions is `i64?`.

vbarua (changes requested):
Overall looks reasonable. Did find one minor bug and requested one extra test.

types/any_type.go
```diff
@@ -49,23 +49,69 @@ func (m *AnyType) ShortString() string {
...
-		return nil, fmt.Errorf("no matching any type found in function parameters")
-	}
-	if index >= len(argumentTypes) {
-		return nil, fmt.Errorf("no matching argument found for any type")
+	// Didn't find matching AnyType.
+	return nil, nil
+}
+
+func (m *AnyType) ReturnType(funcParameters []FuncDefArgType, argumentTypes []Type) (Type, error) {
+	for i := 0; i < min(len(funcParameters), len(funcParameters)); i++ {
```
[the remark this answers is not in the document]
miretskiy: Of course; this code completion!!! What a silly mistake.
It's the other way around -- more argumentTypes than FuncDefArgTypes;  comment added.

types/any_type_test.go
miretskiy: Agreed; done.

types/any_type_test.go
miretskiy: Added; also using any2 instead of any1.

types/any_type_test.go
miretskiy: This part is definitely pointless; on the other hand it is checking 
if setNullability actually worked.
But also, it's not a great function; 
```
		//SetNullability set nullability as given argument
		SetNullability(Nullability) FuncDefArgType
```
It doesn't specify for example if "this" is mutated.  On the one hand, the function signature seems to imply that it isn't because you return FuncDefArgType; on the other hand -- implementations seem to mutate arg itself and return it; I guess for chaining.

So, at least, the comment should say so; AND the test above should probably do
`anyBase.SetNullability(*)` w/out reassigning to anyType to check that arg (anyBase) was modified with appropriate nullability.  At any rate, I think I'll leave it for now.

miretskiy:
@vbarua -- I slightly modified those error check conditions (the ones that didn't have coverage), and added tests for those. to keep code coverage happy.

types/any_type_test.go
```diff
@@ -15,15 +15,144 @@ func TestAnyType(t *testing.T) {
...
+			testName: "wrong_list<any1>",
+			argName:  "any1",
+			parameters: []FuncDefArgType{
+				&Int16Type{},
+				&ParameterizedListType{Type: &AnyType{Name: "any1"}},
+			},
+			args:           []Type{&Int16Type{}, &ListType{}},
+			nullability:    NullabilityRequired,
+			expectedString: "any1",
+			expectedErr:    "expected ListType to have non-nil 1 parameter, found 1",
```
vbarua, line 63: minor: this error isn't super clear IMO `non-nil 1 parameter, found 1`. In this case it's because the parameter was nil. Can we just print the parameters entirely?

types/any_type_test.go
```diff
@@ -15,15 +15,144 @@ func TestAnyType(t *testing.T) {
...
+				&Int16Type{},
+				&ParameterizedMapType{Key: &AnyType{Name: "any2"}, Value: &StringType{}},
+			},
+			args: []Type{
+				&Int16Type{},
+				&MapType{Value: &StringType{}},
+			},
+			nullability:    NullabilityRequired,
+			expectedString: "any2",
+			expectedErr:    "expected MapType to have 2 non-nil parameters, found 2",
```
vbarua, line 108: Can we just print the parameters here as well?

vbarua (approved):
Changes look good. Thanks for working on this.

PR #131, opened 2025-04-17
build: bump substrait-protobuf to v0.66.1

vbarua:

PR #132, opened 2025-04-21
fix(import): fix import path to match v4

miretskiy:
Since the major version was bumped to v4 (#131)
we need the import path to match accordingly using the /v4 suffix as described
[here](https://go.dev/ref/mod#major-version-suffixes)


.golangci.bck.yml
EpsilonPrime, line 1: Is this a copy of the original?

miretskiy: Sorry, I need to revert that.

miretskiy: Reverted.

miretskiy:
Thans, @EpsilonPrime ;

Do we need to tag this w/ v4.0.1?

Issue #135, opened 2025-06-16
Standardized Visitor Pattern for Substrait Protobuf Traversal


alexanderbianchi:
## Summary

We've developed an efficient visitor pattern implementation for traversing Substrait protobuf plans in Go. We'd like to contribute this to the Substrait community as it solves common traversal needs in an idiomatic Go way.

## Background

When working with Substrait plans, we frequently need to:
- Collect statistics about plan nodes
- Compute hashes for plan comparison/caching
- Extract metadata (table names, function calls, etc.)
- Transform or validate plan structures

The naive approach requires writing custom traversal logic for each use case, leading to:
- Duplicate switch statements across different analyzers
- Error-prone traversal logic
- Missed nodes when plan structures evolve
- Performance issues from repeated traversals

### Core Interfaces

```go
// Visitor is a marker interface for all visitors
type Visitor interface{}

// RelVisitor is implemented by visitors that want to visit relation nodes
type RelVisitor interface {
    VisitRel(rel *substraitpb.Rel)
}

// ExprVisitor is implemented by visitors that want to visit expression nodes
type ExprVisitor interface {
    VisitExpr(expr *substraitpb.Expression)
}
```

### Usage Example

```go
// Visitor that only cares about counting relations
type NodeCounter struct {
    count int
}

func (n *NodeCounter) VisitRel(rel *substraitpb.Rel) {
    n.count++
}

// Visitor that only cares about function names
type FunctionCollector struct {
    functions []string
}

func (f *FunctionCollector) VisitExpr(expr *substraitpb.Expression) {
    if sf := expr.GetScalarFunction(); sf != nil {
        f.functions = append(f.functions, sf.Name)
    }
}

// Usage
counter := &NodeCounter{}
traverse.Walk(plan, counter)
fmt.Printf("Found %d relations\n", counter.count)
```

### Multiple Visitors in Single Traversal

```go
// Combine multiple analyses in one pass
multi := traverse.NewMultiVisitor(
    NewHashVisitor(ctx),
    NewStatsCollector(),
    NewTableExtractor(),
)
traverse.Walk(plan, multi)
```

## Why This Design is Idiomatic Go

1. **Optional Interface Implementation**: Following patterns like `http.ResponseWriter` + `http.Flusher`, visitors only implement the methods they need:
   ```go
   if exprVisitor, ok := visitor.(ExprVisitor); ok {
       exprVisitor.VisitExpr(expr)
   }
   ```

[...]

PR #136, opened 2025-06-17
Add Substrait Protobuf Visitor Framework

alexanderbianchi:
## Summary
We've developed an efficient visitor pattern implementation for traversing Substrait protobuf plans in Go. We'd like to contribute this to the Substrait community as it solves common traversal needs in an idiomatic Go way.

## Issue
Tradeoffs and design is outlined in https://github.com/substrait-io/substrait-go/issues/135


traverse/visitor.go
```diff
@@ -0,0 +1,117 @@
...
+- Interface-based visitor pattern
+- Single tree walk for multiple analyses
+- Sequential execution
+- Read-only operations - NO MUTATIONS
+- Zero allocations per node
+
+## IMPORTANT: Read-Only Analysis Only
+This interface is designed for read-only analysis, and assumes the tree does not change
+as it is being walked. Any modification of the tree could disrupt the walking of the
+tree causing nodes to be missed, visited out of order, or cause panics.
```
EpsilonPrime, line 16: Some modifications would be reasonably safe with this interface.  Here are a few possibilities: https://github.com/voltrondata/spark-substrait-gateway/tree/main/src/transforms

traverse/visitor.go
```diff
@@ -0,0 +1,117 @@
...
+	// Visitor that only cares about relations
+	type NodeCounter struct {
+		count int
+	}
+
+	func (v *NodeCounter) VisitRel(rel *proto.Rel) {
+		v.count++
+	}
+
+	// Visitor that only cares about expressions (e.g., finding all functions)
```
EpsilonPrime, line 45: Presumably for finding all utilized functions versus checking the extension list.

traverse/visitor.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// ExprVisitor is implemented by visitors that want to visit expression nodes.
+type ExprVisitor interface {
+	VisitExpr(expr *proto.Expression)
+}
+
+// Visit provides a convenient way to traverse a Substrait plan.
+// It walks all relations in the plan, calling the appropriate visitor methods
+// based on what interfaces the visitor implements.
```
EpsilonPrime, line 92: Probably worth noting that you have to visit the entire plan.  So if you wanted to stop early once you found something you couldn't handle (such as an unsupported type) you couldn't bail early.

traverse/visitor.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+}
+
+// Walk traverses a Substrait relation tree, applying the visitor to each node.
+// The visitor's methods are called based on which interfaces it implements.
+// This function guarantees zero allocations per node visited.
+//
+// If your visitor needs plan context (for extensions, etc.), construct it
+// with NewPlanContext and pass it to your visitor's constructor.
+func Walk(rel *proto.Rel, visitor Visitor) {
```
EpsilonPrime, line 112: Could you use this to start a visit at any point in the tree?  An alternative name could be VisitRelation since it also visits.

traverse/walk.go
```diff
@@ -0,0 +1,377 @@
+// SPDX-License-Identifier: Apache-2.0
+package traverse
+
+import (
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+)
+
+// walk implements the core traversal logic for Substrait relations.
+// For cycle detection in DAG structures, wrap your visitor with CycleDetectingVisitor.
```
EpsilonPrime, line 9: I don't see `CycleDetectingVisitor` in this PR.  That said are there any use cases where we would need them?

traverse/visitor.go
alexanderbianchi: Fair, yeah I like that name more. Will update.

traverse/walk.go
alexanderbianchi: So, not really? If you were accepting substrait protobufs that were built within the service then there isn't anything in go to stop a cycle, but if you unmarshal the protobuf then it couldn't have a cycle. I think it's just cleaner to remove this for now.

alexanderbianchi:
Thanks @EpsilonPrime for the review, will fix up the comments and push a revision.

PR #140, opened 2025-06-25
fix: set Partitions in windowFuncBuilder Build() method

miretskiy:
Window function builder was incorrectly omitting partitions. Fix this.

Add `fmt.Stringer` implementation for window function `Bound` implementations.  Fix nil panic when constructing Bound from proto; Add `BoundFromProto` tests.


miretskiy:
Apologies for seemingly unrelated changes; whatever `go fmt` says goes.

EpsilonPrime (approved):

vbarua (approved):
LGTM

PR #142, opened 2025-07-03
feat: support aggregate functions in window contexts

miretskiy:
Aggregate functions, such as `SUM`, can be used in window context: `SUM(x) OVER(PARTITION BY ...)`.

Prior to this PR it was impossible to construct such function since attempt to construct window function would fail (sum is an aggregate), and attempts to set window attributes (partitition, etc) on an aggregate would fail too.

Extend this library to support this use case.

### DIFF
--- a/expr/binding_test.go
+++ b/expr/binding_test.go
@@ -39,6 +39,9 @@ var (
 	ntileID = extensions.ID{
 		URI:  extensions.SubstraitDefaultURIPrefix + "functions_arithmetic.yaml",
 		Name: "ntile"}
+	sumID = extensions.ID{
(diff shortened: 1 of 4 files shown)


miretskiy:
@Corentin-Massala -- please see @vbarua comment; see what you can do with that.
It sounds reasonable to me, though it's not clear (esp while on vacation) if one approach is superior to another.  Try it; see what it feels like;

Corentin-Massala:
I'll give it a try !

vbarua:
This was implemented by https://github.com/substrait-io/substrait-go/pull/143

Issue #144, opened 2025-07-11
[Bug] Return type derivation for functions that return compound types and have enum arguments


lorenarosati:
## Description
For functions that return compound types (e.g. precision_timestamp\<P>) that need to be resolved to types with concrete parameters, the call to ReturnType() [here](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/extensions/variants.go#L134) will wrap the return type expression in OutputDerivation ([here](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/types/parameterized_single_integer_param_type.go#L104)) and call ReturnType() on a type types.OutputDerivation ([here](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/types/type_derivation.go#L334)).

This will build a symbol table that maps the parameter names (e.g. "P") to the actual parameter values. This function iterates over the types of the function parameters from the function definitions (funcParameters []FuncDefArgType) and matches its parameterized parameters with the concrete parameters of the argument types (argumentTypes []Type). 

The issue is that, for EnumType, [GetParameterizedParams()](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/types/types.go#L569) returns a list of its options (and this gets called on each function parameter), whereas [GetParameters()](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/types/types.go#L552) returns an empty list (and this gets called on each argument type).  For functions that have enum arguments, [this](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/types/type_derivation.go#L316) statement will result in an error because the lengths of the two lists won't be the same, and the return type will not be able to be resolved.

For example, if we try to call [ResolveType()](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/extensions/variants.go#L331) on the function "round_temporal" ([the implementation that returns precision_timestamp\<P>](https://github.com/substrait-io/substrait/blob/793c64ba26e337c22f5e91b658be58b1eea7efd3/extensions/functions_datetime.yaml#L832)), we'll get the following error:
`function parameter rounding has 4 parameters, but 0 were provided`
This is because rounding has 4 options.

## To Reproduce
[...]

PR #150, opened 2025-08-05
fix: EnumType should not have parameterized parameters

lorenarosati:
EnumType was being treated as if its options were parameterized parameters in GetParameterizedParams(), however, GetParameters() returned an empty array. This PR introduces a breaking change - EnumType has no parameterized parameters, and HasParameterizedParams() returns false. 

This inconsistent behavior was causing an issue with resolving return types of functions - see #144

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -563,15 +563,11 @@ func (e *EnumType) SetNullability(n Nullability) FuncDefArgType {
 }
 
 func (e *EnumType) HasParameterizedParam() bool {
-	return true
+	return false
 }
 
 func (e *EnumType) GetParameterizedParams() []interface{} {
-	params := make([]interface{}, len(e.Options))
-	for i, p := range e.Options {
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -563,15 +563,11 @@ func (e *EnumType) SetNullability(n Nullability) FuncDefArgType {
...
-	return true
+	return false
 }
 
 func (e *EnumType) GetParameterizedParams() []interface{} {
-	params := make([]interface{}, len(e.Options))
-	for i, p := range e.Options {
-		params[i] = p
-	}
-	return params
```
vbarua, line 574: This makes sense to me. Options are not parameterized params.

vbarua (approved):
This looks reasonable to me. I'll leave this open for a couple of days before merging.

As this is a breaking change, I will handle the merge and release because it requires some special handling.

PR #151, opened 2025-08-07
Adds constructors to plan builder for extension rels

Adam-Alani:
Currently, there doesnt seem to be a way to create Extension Rels as extensions fields are private (details and input), This PR adds constructors for these relations like other rels in plan/builder.


wackywendell:
This PR adds reasonable constructors for the current `Extension*Rel` types, but I think the extension relation design needs more flexibility before we bake it into the builder.

### Schema Handling

Currently, the schema for extensions is rigid:
- `ExtensionLeafRel` returns an empty schema and doesn’t allow override.
- `ExtensionSingleRel` returns its input’s schema by default; that's a reasonable default, but should allow override (e.g. ProjectRel does not necessarily output just its input…)
- `ExtensionMultiRel` also defaults to an empty schema, which is a poor default

This is probably wrong most of the time. I suggest that all three extension types should allow the output schema to be explicitly set or overridden, as the default may not be suitable, especially for `ExtensionLeafRel` (and the current default for `ExtensionMultiRel`).

### Expressions

I haven't fully thought this through, but if an extension contains expressions, do we need to hook those into the main system for traversal, rewriting, keeping the extension registry up to date? I'm not sure.

### Proposal

- Introduce a new interface for extension rels, with at least:
  - `Schema(inputs []Rel) types.RecordType`
  - `Build(inputs []Rel) *anypb.Any`
  - (Uncertain) `Expressions(inputs []Rel) []Expression` - returns any expression trees within
  - (Not sure if `inputs []Rel` is necessary / good, but it seems reasonable, especially for `Schema`)
- Update the internals of the `Extension*Rel` types to hold this interface, not just `*anypb.Any`, and to use it for building / producing a schema

### Questions

- For the schemas, should we allow there to be a 'default' if unset?
  - For `ExtensionSingleRel`, there's an obvious one (the input), and for `ExtensionMultiRel`, there's a similarly good one (concatenate inputs, like `Join`), but they should be overrideable.
  - But those are simple enough that perhaps extension creators should just explicitly do that themselves?
- Do we need to keep the `anypb.Any` for backwards-compatibility?
  - I'm guessing not, since it's a private field? Or is there some way to use the existing `Extension*Rel` objects?
- Does having an `Expressions()` method make sense?
  - We don't need it now, but down the line, we might - and then we'd have a backwards compatibility problem if we added it to the interface, right?

### …

Does this approach make sense? What are your thoughts?

Adam-Alani:
> Do we need to keep the anypb.Any for backwards-compatibility?
> I'm guessing not, since it's a private field? Or is there some way to use the existing Extension*Rel objects?

I think we might have to for transformation from proto `RelFromProto` in `plan.go`,

wackywendell:
> I think we might have to for transformation from proto `RelFromProto` in `plan.go`

Ah, good find! Looking at it, to be more precise, it looks like we need to preserve that `RelFromProto` *may* produce an `ExtensionLeafRel`, `ExtensionMultiRel`, or `ExtensionSingleRel`; it doesn't need to have the `detail *anypb.Any` field in that form for backwards compatibility, right? So we could create a an `UnknownExtension` struct that wraps the `detail *anypb.Any` and fulfills the `ExtensionRelDefinition` interface? Does that make sense?

Adam-Alani:
> > I think we might have to for transformation from proto `RelFromProto` in `plan.go`
> 
> Ah, good find! Looking at it, to be more precise, it looks like we need to preserve that `RelFromProto` _may_ produce an `ExtensionLeafRel`, `ExtensionMultiRel`, or `ExtensionSingleRel`; it doesn't need to have the `detail *anypb.Any` field in that form for backwards compatibility, right? So we could create a an `UnknownExtension` struct that wraps the `detail *anypb.Any` and fulfills the `ExtensionRelDefinition` interface? Does that make sense?

Makes sense, I changed this, although i wonder now if this can be a bit confusing that protos use details, and internally we use definitions

plan/relations.go
```diff
@@ -1566,30 +1566,80 @@ func (s *SetRel) Remap(mapping ...int32) (Rel, error) {
...
+// UnknownExtension wraps an *anypb.Any detail for extension relations parsed from protobuf that don't have a definition.
+type UnknownExtension struct {
+	detail *anypb.Any
+}
+
+// Schema returns an empty record type for unknown extensions.
+func (ue *UnknownExtension) Schema(inputs []Rel) types.RecordType {
+	if len(inputs) > 0 {
+		return inputs[0].RecordType()
+	}
```
wackywendell, line 1592: This doesn't match the docstring above. And technically, for backwards-compatibility, this should only trigger for `len(inputs) == 1`, right? (well, technically, that would *still* be wrong for an `ExtensionMultiRel` with 1 input, but I think would be close enough…)

My suggestion: let's get it a bit closer to the previous behavior, and note the situation

```suggestion
	if len(inputs) == 1 {
		// XXX: the UnknownExtension doesn't know its schema type. For now, we
		// guess that if there is exactly one input, its probably that record
		// type, otherwise, empty; that's not a very good choice, but its now
		// what we have backwards-compatibility-wise. We should probably hook
		// this into the ExtensionRegistry or similar so we can look it up and
		// get the correct schema.
		return inputs[0].RecordType()
	}
```

plan/relations.go
```diff
@@ -1566,30 +1566,80 @@ func (s *SetRel) Remap(mapping ...int32) (Rel, error) {
...
-	input  Rel
-	detail *anypb.Any
+	input      Rel
+	definition ExtensionRelDefinition
 }
 
 func (es *ExtensionSingleRel) directOutputSchema() types.RecordType {
+	if es.definition != nil {
+		return es.definition.Schema([]Rel{es.input})
+	}
```
wackywendell, line 1617: Is `nil` a valid value for the `definition`? Shouldn't it always be something? I would suggest that we define `nil` here as invalid, I don't see a case where that is useful

plan/relations.go
```diff
@@ -1679,16 +1741,29 @@ func (el *ExtensionLeafRel) Remap(mapping ...int32) (Rel, error) {
...
-	detail *anypb.Any
+	inputs     []Rel
+	definition ExtensionRelDefinition
 }
 
-func (em *ExtensionMultiRel) directOutputSchema() types.RecordType { return types.RecordType{} }
+func (em *ExtensionMultiRel) directOutputSchema() types.RecordType {
+	if em.definition != nil {
+		return em.definition.Schema(em.inputs)
+	}
```
wackywendell, line 1751:
```suggestion
```
As above, I don't think this should be valid

plan/relations.go
```diff
@@ -1631,21 +1681,33 @@ func (es *ExtensionSingleRel) Remap(mapping ...int32) (Rel, error) {
...
 
-	detail *anypb.Any
+	definition ExtensionRelDefinition
 }
 
-func (el *ExtensionLeafRel) directOutputSchema() types.RecordType { return types.RecordType{} }
+func (el *ExtensionLeafRel) directOutputSchema() types.RecordType {
+	if el.definition != nil {
+		return el.definition.Schema([]Rel{})
+	}
```
wackywendell, line 1690:
```suggestion
```
As above, I don't think this should be valid

wackywendell (commented):
I have several more comments. Structurally I think we're in a good place, and with a bit more tightening of the code this will be ready to go!

plan/relations.go
Broshen: just to clarify - do we want to e.g.
```
if es.definition == nil {
    panic(...)
}
```
then? if a nil definition is invalid
and otherwise, return `es.definition.Schema(...)`?
would this be a breaking change?

wackywendell: No, I don't think you need to check it at all in this function; you can assume its never `nil`. And then the constructors should enforce this, by panicking or returning an error. The Builder constructor for this returns an error, so I think we can just drop this check entirely.

As to the breaking change - the `definition` did not exist previously, and is not a public interface; we're defining it for the first time in this PR. Breaking changes are defined in terms of public interfaces, so if you have a concern there, can you tell me what public interface you are concerned we are breaking, and what previously supported behavior is affected?

Broshen: ahh okay yeah 🤦 my mind's still not used to what's considered breaking or not

my train of thought was that the behavior of `directOutputSchema` changes - now panics if `definition` is not set, whereas before, it would always just return `types.RecordType{}`
and `directOutputSchema` is called by `RecordType()` which is public, so the behavior of that changes
== users can't rely on old behavior == breaking 🤦 😂 

but "breaking changes" is really only wrt code/signatures that would be exposed outside the package directly, not "behavior"... i forget to make the distinction between that

+ i'd imagine the Extension rels wouldn't have been used anywhere right now anyways... they would've been pretty useless if all they ever did was return empty `RecordTypes{}`

wackywendell: > now panics if `definition` is not set, whereas before, it would always just return `types.RecordType{}`

Before, `definition` did not exist, so it would not affect what was returned. If `detail` was `nil`, which was possible as it came directly from the protobuf `detail` field, then it would return `types.RecordType{}`, but `detail` is a different field. We have handled that case - where the protobuf `Detail` field is nil - by wrapping it in an `UnknownExtension` (well, we currently do that with _all_ `detail` fields, nil or not…).

If it helps: a breaking change is when code that previously worked would not compile (due to changes in constraints, parameters, …), or if reasonable (non-buggy) uses of the _public interface_ (exported types, methods, functions) would have different results. In the case I am talking about above, the `Detail` field _of the Protobuf_ is a public interface; a `nil` value there is possible; so that should should not fail, and the resulting built relations should have the same outputs for their schemas. Well, arguably that output is buggy (It pretty much never makes sense for a node to have an empty output) and we aren't *exactly* keeping the same behavior there, but we mostly are, and panicking there would definitely be a negative.

Issue #145, opened 2025-07-11
[Bug] Precision not set properly for IntervalDayToSecond literal


lorenarosati:
When creating a literal of type IntervalDayToSecond with the [NewLiteral()](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/expr/literals.go#L890) function, the final returned literal does not have the precision set to the correct precision. 

This is because this should be taking the precision from val and setting it [here](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/expr/literals.go#L940), but it doesn't do that and the precision defaults to 0:
```
case *types.IntervalDayToSecond:
    return &ProtoLiteral{
        Value: v,
	Type: &types.IntervalDayType{
		Nullability: getNullability(nullable),
	},
    }, nil
```

While precision is optional for this type, if provided, it should still be set properly.

bvolpato:
Resolved by PR #155 (merged Aug 2025).

PR #155, opened 2025-08-11
fix: set Precision on IntervalDayToSecond literals

lorenarosati:
When creating a literal of type IntervalDayToSecond with the [NewLiteral()](https://github.com/substrait-io/substrait-go/blob/ab158b528f583af016d683ab8153cf7f901441c2/expr/literals.go#L890) function, the final returned literal did not have the precision set to the correct precision for its type (its value had a different precision than its type).

See this issue: #145


vbarua (approved):
LGTM. Thanks for fixing this 🙇

PR #163, opened 2025-09-04
Add helper fn to construct VirtualTable from go primitives

### LINKED ISSUE #162: Add helper to construct virtual tables from go primitives
It would be nice to have something like: 
```go
VirtualTableFromGoPrimitives(["name", "age"], [{"Alice", 25}, {"Bob", nil}], [false, true])
```
where:
- first arg is field names
- second arg is table contents
- third arg is nullability of the fields

benbellick:
Closes #162


plan/builders.go
```diff
@@ -124,6 +124,12 @@ type Builder interface {
 	// Deprecated: Use VirtualTableFromExpr(...).Remap() instead.
 	VirtualTableFromExprRemap(fieldNames []string, remap []int32, values ...expr.VirtualTableExpressionValue) (*VirtualTableReadRel, error)
 	VirtualTableFromExpr(fieldNames []string, values ...expr.VirtualTableExpressionValue) (*VirtualTableReadRel, error)
+	// VirtualTableFromGoTypes constructs a VirtualTableReadRel from native Go types.
+	// It accepts field names, tuples of polymorphic Go values, and optional nullability
+	// configuration. The function automatically maps Go types to appropriate Substrait types
+	// and handles nil values by converting them to typed null literals.
+	// If nullableColumns is nil, all columns default to non-nullable (required).
+	VirtualTableFromGoTypes(fieldNames []string, tuples [][]any, nullableColumns []bool) (*VirtualTableReadRel, error)
```
miretskiy, line 132: I'm not one of the approvers (or reviewers) so... take it or not...  I'm not really sure that this kind of logic belongs with the builders.  Go types can get pretty complex pretty fast; not sure the library itself should carry this complexity.

But even ignoring the above, if maintainers think it's useful (and , I do think its definitely applicable and a common enough task for Go users of this library), then at least I'd encourage not to expand the Builder interface.  I think a standalone function should work just fine; And also, we might want to consider adding it to the "literal" package -- it's a utility that constructs virtual scan from literals; maybe that package is a better home for it?

benbellick: Hey, thanks for the comment. I think that is a great point. I'll look into moving it to another place briefly. Thanks!

benbellick: Moved to the `literals` package 👍

benbellick:
Switching to a draft for now as it is low priority.

PR #165, opened 2025-09-16
Gracefully handle URI->URN migration

benbellick:


extensions/extension_mgr.go
```diff
@@ -183,14 +208,15 @@ func checkMaps[T variants](id ID, m map[ID]T, simpleNames map[ID]string) (T, boo
...
 }
 
-func addToMaps[T variants](id ID, fn extFn[T], m map[ID]T, simpleMap map[string]string) {
-	variants := fn.GetVariants(id.URI)
+func addToMaps[T variants](fn extFn[T], m map[ID]T, simpleMap map[string]string, translator *uriUrnTranslator) {
+	variants := fn.GetVariants()
 	for _, v := range variants {
-		id.Name = v.CompoundName()
-		m[id] = v
+		// Each variant should already have its ID set correctly
```
benbellick, line 218: So basically the idea here is that whenever we query something via ID. We really query it by its "canonical" ID

benbellick:
I see that the tests in CI/CD are failing. This is because I rendered some of the proto plans out into JSON, which included the field `"producer":     string("substrait-go darwin/arm64"),` which differs from what is produced in the CI/CD machine. As a fix, I'm going to (force-)push a change that will only use hand constructed plans.

PR #177, opened 2025-11-24
Update UserDefinedType reference only if it is ParameterizedUserDefinedType

Adam-Alani:
This pull request improves the handling of user-defined types (UDTs) in type expression evaluation, specifically addressing how `TypeReference` is set for parameterized UDTs. It also adds new tests to ensure correct behavior when dealing with UDTs returned from generic types like `AnyType`.

Type resolution improvements:

* Updated `EvaluateTypeExpression` in `variants.go` to correctly set the `TypeReference` for `ParameterizedUserDefinedType` instances, ensuring that the reference is filled in using the proper name and URN from the parameterized type expression.


extensions/variants.go
```diff
@@ -140,8 +140,10 @@ func EvaluateTypeExpression(urn string, nullHandling NullabilityHandling, return
 	}
 
 	if udt, ok := outType.(*types.UserDefinedType); ok {
-		name := strings.TrimPrefix(returnTypeExpr.ShortString(), "u!") // short string contains the u! prefix, but type definitions in the extensions don't
-		udt.TypeReference = registry.GetTypeAnchor(ID{Name: name, URN: urn})
+		if udt.TypeReference == 0 {
```
benbellick: `TypeReference` == 0 doesn't necessarily mean that type reference wasn't set. It could be the case that a type anchor of `0` was used. We discourage the usage of 0 but don't disallow it.

PR #183, opened 2026-01-20
feat: add lambda expression support

### LINKED ISSUE #187: Implement lambdas
Now that https://github.com/substrait-io/substrait/pull/889 is merged in, lets implement lambdas!

Slimsammylim:
Adds lambda expression support:

- Represent lambdas natively with ability to build them with validation
- Consume lambdas from protobuf (validation not yet implemented)
- Parse `FuncType` from extension YAML (enables `functions_list.yaml`)

Closes #187

### DIFF
--- a/expr/lambda_test.go
+++ b/expr/lambda_test.go
@@ -284,7 +284,7 @@ func TestLambdaBuilder_TypeResolution(t *testing.T) {
 		name         string
 		paramTypes   []types.Type
 		fieldIndex   int32
-		expectedType string
+		expectedType types.Type
 	}{
 		{
 			name: "reference first param (i32)",
@@ -292,7 +292,7 @@ func TestLambdaBuilder_TypeResolution(t *testing.T) {
 				&types.Int32Type{Nullability: types.NullabilityRequired},
 			},
 			fieldIndex:   0,
-			expectedType: "i32",
+			expectedType: &types.Int32Type{Nullability: types.NullabilityRequired},
 		},
 		{
 			name: "reference second param (i64)",
@@ -301,7 +301,7 @@ func TestLambdaBuilder_TypeResolution(t *testing.T) {
 				&types.Int64Type{Nullability: types.NullabilityRequired},
 			},
 			fieldIndex:   1,
-			expectedType: "i64",
+			expectedType: &types.Int64Type{Nullability: types.NullabilityRequired},
 		},
 		{
 			name: "reference third param (string)",
@@ -311,7 +311,7 @@ func TestLambdaBuilder_TypeResolution(t *testing.T) {
 				&types.StringType{Nullability: types.NullabilityRequired},
 			},
 			fieldIndex:   2,
-			expectedType: "str",
+			expectedType: &types.StringType{Nullability: types.NullabilityRequired},
 		},
 		{
 			name: "reference float64 param",
@@ -319,7 +319,7 @@ func TestLambdaBuilder_TypeResolution(t *testing.T) {
 				&types.Float64Type{Nullability: types.NullabilityRequired},
 			},
 			fieldIndex:   0,
-			expectedType: "fp64",
+			expectedType: &types.Float64Type{Nullability: types.NullabilityRequired},
 		},
 	}
 
@@ -344,14 +344,16 @@ func TestLambdaBuilder_TypeResolution(t *testing.T) {
 			require.NotNil(t, lambdaType, "Lambda should have a resolved type")
 			funcType, ok := lambdaType.(*types.FuncType)
 			require.True(t, ok, "Lambda type should be FuncType")
-			require.Equal(t, tc.expectedType, funcType.ReturnType.ShortString(),
-				"Lambda return type should match referenced parameter type")
+			require.True(t, tc.expectedType.Equals(funcType.ReturnType),
+				"Lambda return type should match referenced parameter type: expected %s, got %s",
+				tc.expectedType.String(), funcType.ReturnType.String())
 
 			// Also verify body's type matches
 			bodyType := lambda.Body.GetType()
 			require.NotNil(t, bodyType, "Body should have a resolved type")
-			require.Equal(t, tc.expectedType, bodyType.ShortString(),
-				"Body type should match referenced parameter type")
+			require.True(t, tc.expectedType.Equals(bodyType),
+				"Body type should match referenced parameter type: expected %s, got %s",
+				tc.expectedType.String(), bodyType.String())
 
 		})
 	}


Slimsammylim:
@codex review

expr/lambda.go
```diff
@@ -0,0 +1,380 @@
...
+
+// resolveLambdaParamTypes walks the body expression and resolves types for any
+// FieldReferences that have LambdaParameterReference roots.
+// - stepsOut=0: resolves against params (this lambda's parameters)
+// - stepsOut>0: resolves against outerParams (outer lambda parameters)
+// For nested lambdas, recursively resolves with updated outer context.
+// Recurses into all descendants (e.g., Cast(FieldRef), ScalarFunction(Cast(FieldRef))).
+func resolveLambdaParamTypes(body Expression, params *types.StructType, outerParams []*types.StructType) Expression {
+	// Handle if body itself is a nested lambda - must check BEFORE Visit
+	// because Lambda.Visit would bypass our nested lambda context handling
```
Slimsammylim: Lambda.Visit(callback) internally calls callback(lambda.Body), which skips the current Lambda wrapper entirely. This loses the context we need to resolve stepsOut>0 references. Therefore, we need to handle if the body is a lambda first.

expr/lambda_test.go
```diff
@@ -0,0 +1,1714 @@
...
+		"Body references parameter %d but lambda only has %d parameters", fieldIndex, numParams)
+
+	// Verify the referenced parameter type matches what we expect
+	referencedParamType := params.GetTypes()[fieldIndex]
+	require.NotNil(t, referencedParamType.GetI32(),
+		"Referenced parameter %d should be i32 type", fieldIndex)
+
+	t.Logf("Successfully parsed lambda reference: %v", &exprProto)
+	t.Logf("Body references parameter %d (type: i32)", fieldIndex)
+}
```
benbellick:
```suggestion
```

Same with this one.

expr/lambda_test.go
```diff
@@ -0,0 +1,1714 @@
...
+	require.Equal(t, int32(2), literal.GetI32(), "Literal should be 2")
+
+	// Verify output type
+	outputType := scalarFunc.GetOutputType()
+	require.NotNil(t, outputType.GetI32(), "Output type should be i32")
+
+	t.Logf("Successfully parsed lambda with function: %v", &exprProto)
+	t.Logf("Function has %d arguments, first is param ref, second is literal %d",
+		len(scalarFunc.GetArguments()), literal.GetI32())
+}
```
benbellick: I think same as this one, too.
```suggestion
```

expr/lambda_test.go
```diff
@@ -0,0 +1,1714 @@
...
+
+	t.Logf("Lambda Go expression: %s", lambda.String())
+	t.Logf("Body references parameter %d via LambdaParameterReference", structFieldRef.Field)
+	t.Logf("Body type: %s, Lambda type: %s", bodyType.ShortString(), lambdaType.ShortString())
+}
+
+// TestLambdaWithFunctionExprFromProto converts a lambda with a scalar function
+// from protobuf to Go expression. Uses a minimal plan to set up extensions.
+// Lambda: (x: i32) -> multiply(x, 2)
+func TestLambdaWithFunctionExprFromProto(t *testing.T) {
```
benbellick: It seems a bit strange to me to have both `planJSON` and `lambdaJSON` in this plan. Is the intention to just have one full plan with the lambda inside of it?

Slimsammylim: This made it easier to get the lambda proto, but I see how this looks confusing. I will change it to one plan!

expr/builder.go
```diff
@@ -407,3 +507,56 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+	// 2. Get the target lambda parameters
+	targetParams := lpb.b.lambdaContext[len(lpb.b.lambdaContext)-1-int(lpb.stepsOut)]
+
+	// 3. Validate field index for StructFieldRef (most common case)
+	if structRef, ok := lpb.ref.(*StructFieldRef); ok {
+		if int(structRef.Field) >= len(targetParams.Types) {
+			return nil, fmt.Errorf("%w: lambda body references parameter %d but lambda only has %d parameters",
+				substraitgo.ErrInvalidExpr, structRef.Field, len(targetParams.Types))
+		}
+	}
```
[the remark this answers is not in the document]
benbellick: If I understand correctly, I think this would subsume all of the logic above here.

benbellick (changes requested):
I think that the logic of the PR is sound! In terms of the package code, I have just a couple comments about comments and about a small refactor. 

The core of the comments though are about the tests. I think there is opportunity to make them a bit simpler. I didn't get a chance to go through all of them by hand quite yet, but hopefully the comments I've given so far are useful. Thanks!

EDIT: Also FYI, looks like you are failing some of the CI checks now.

expr/builder.go
```diff
@@ -18,9 +25,9 @@ type Builder interface {
 
 // ExprBuilder is the parent context for all the expression builders.
 // It maintains a pointer to an extension registry and, optionally,
-// a pointer to a base input schema. This allows less verbose expression
-// building as it isn't necessary to pass these to every `New*` function
-// to construct the expressions.
+// a pointer to a base input schema and a slice of lambda parameters.
+// This allows less verbose expression building as it isn't necessary
+// to pass these to every `New*` function to construct the expressions.
```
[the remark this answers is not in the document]
Slimsammylim: That makes sense

expr/builder.go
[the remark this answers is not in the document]
Slimsammylim: I will add a helper. Since we need the `lpb`'s `ref`, I'll make it a func on`*lambdaParamRefBuilder`.

expr/lambda.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/substrait-io/substrait-go/v7/types"
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+)
+
+// Lambda represents a lambda expression with parameters and a body.
+// It is required that the parameters be a struct type with NULLABILITY_REQUIRED.
+type Lambda struct {
+	Parameters *types.StructType
+	Body       Expression
+}
```
benbellick: nit:
```suggestion
// Lambda represents a lambda expression with parameters and a body.
type Lambda struct {
	Parameters *types.StructType // The formal lambda parameters, required to have NULLABILITY_REQUIRED
	Body       Expression
}
```

expr/lambda.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (l *Lambda) isRootRef() {}
+
+func (l *Lambda) IsScalar() bool {
+	return l.Body.IsScalar()
+}
+
+func (l *Lambda) GetType() types.Type {
+	return l.Body.GetType()
```
benbellick: This one isn't quite right either. The "type" of a lambda is a distinct thing, which we call `Func`. If that type doesn't exist in types, you may need to add it.

expr/builder.go
```diff
@@ -407,3 +507,65 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+	ref      Reference
+	stepsOut uint32
+}
+
+func (lpb *lambdaParamRefBuilder) Build() (*FieldReference, error) {
+	// Validate stepsOut - check we have enough outer lambdas
+	if int(lpb.stepsOut) >= len(lpb.b.lambdaContext) {
+		return nil, fmt.Errorf("%w: stepsOut %d references non-existent outer lambda (only %d outer lambdas available)",
+			substraitgo.ErrInvalidExpr, lpb.stepsOut, len(lpb.b.lambdaContext))
+	}
```
benbellick: nit: I think this section should be part of the check of getLambdaParamType

expr/builder.go
```diff
@@ -407,3 +507,65 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+
+	// Get the target lambda parameters
```
benbellick:
```suggestion
```

expr/builder.go
```diff
@@ -407,3 +507,65 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+	targetParams, err := lpb.getLambdaParamType()
```
benbellick:
```suggestion
	targetParam, err := lpb.getLambdaParamType()
```

nit: Am I correct that this is only one parameter, not multiple?

expr/builder.go
```diff
@@ -407,3 +507,65 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+	}
+
+	// This should never happen - all Reference types should implement ReferenceSegment
+	return nil, fmt.Errorf("%w: reference type %T does not implement ReferenceSegment interface",
+		substraitgo.ErrInvalidExpr, lpb.ref)
+}
+
+// getLambdaParamType gets the target lambda parameters struct from the context stack and validates that the parameter exists.
+// Returns an error if the parameter does not exist.
+func (lpb *lambdaParamRefBuilder) getLambdaParamType() (*types.StructType, error) {
```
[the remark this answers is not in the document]
benbellick: I think this change plus [this comment](https://github.com/substrait-io/substrait-go/pull/183/changes#r2738951310) means that the builder for lambda param ref becomes as simple as: 
```golang
func (lpb *lambdaParamRefBuilder) Build() (*FieldReference, error) {
	resolvedType, err := lpb.getLambdaParamType()
	if err != nil {
		return nil, err
	}

	return &FieldReference{
		Root:      LambdaParameterReference{StepsOut: lpb.stepsOut},
		Reference: lpb.ref,
		knownType: resolvedType,
	}, nil
}
```

benbellick (commented):
Have to run out but just a few more comments. Thanks so much!

expr/lambda.go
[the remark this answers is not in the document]
Slimsammylim: I think it makes sense to keep it seperate. The next PR is going to add back a lot of code for validation in `ExprFromProto`.

expr/testdata/lambda/lambda-field-ref.json
benbellick: I know its a bit of a pain, but can we just have all of the JSON files only be plans? Just because it makes the test a bit simpler if we don't have to do any conditional handling.

types/types.go
```diff
@@ -1182,6 +1194,102 @@ func (*StructType) BaseString() string {
...
+
+func (*FuncType) isRootRef() {}
+
+func (f *FuncType) WithNullability(n Nullability) Type {
+	out := *f
+	out.Nullability = n
+	return &out
+}
+
+func (f *FuncType) GetParameters() []interface{} {
```
benbellick: This implementing a function from the `Type` interface, so both returning `[]interface{}` and appending the return type to the returned list is correct 👍

types/types.go
```diff
@@ -1182,6 +1194,102 @@ func (*StructType) BaseString() string {
...
+	ReturnType     Type
+}
+
+func (*FuncType) isRootRef() {}
+
+func (f *FuncType) WithNullability(n Nullability) Type {
+	out := *f
+	out.Nullability = n
+	return &out
+}
```
benbellick: I think that this is _not_ a deep copy, so the same `ParameterTypes` is used in the implicit parameter and the returned parameter. Is that safe to do in this context? It might be, I genuinely am unsure.

Slimsammylim: Hm, I thought this function was a bit strange too. This definition was mostly me just following how previous types have defined this function

expr/builder.go
```diff
@@ -407,3 +510,56 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+	// Validate stepsOut - check we have enough outer lambdas
+	if int(lpb.stepsOut) >= len(lpb.b.lambdaContext) {
+		return nil, fmt.Errorf("%w: stepsOut %d references non-existent outer lambda (only %d outer lambdas available)",
+			substraitgo.ErrInvalidExpr, lpb.stepsOut, len(lpb.b.lambdaContext))
+	}
+	targetParams := lpb.b.lambdaContext[len(lpb.b.lambdaContext)-1-int(lpb.stepsOut)]
+
+	// Validate that the parameter field index exists
+	if int(lpb.ref.Field) >= len(targetParams.Types) {
+		return nil, fmt.Errorf("%w: lambda body references parameter %d but lambda only has %d parameters",
```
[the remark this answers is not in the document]
benbellick: Otherwise, it could be ambiguous from the error message which one we are talking about. Let's also make sure some test captures that this info is present in the error message.

expr/builder.go
```diff
@@ -407,3 +510,56 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+			substraitgo.ErrInvalidExpr, lpb.ref.Field, len(targetParams.Types))
+	}
+
+	// Resolve type using GetType method
```
benbellick:
```suggestion
```

expr/builder.go
```diff
@@ -407,3 +510,56 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+// Returns an error if the parameter does not exist.
+func (lpb *lambdaParamRefBuilder) getLambdaParamType() (types.Type, error) {
+	// Validate stepsOut - check we have enough outer lambdas
+	if int(lpb.stepsOut) >= len(lpb.b.lambdaContext) {
+		return nil, fmt.Errorf("%w: stepsOut %d references non-existent outer lambda (only %d outer lambdas available)",
+			substraitgo.ErrInvalidExpr, lpb.stepsOut, len(lpb.b.lambdaContext))
+	}
+	targetParams := lpb.b.lambdaContext[len(lpb.b.lambdaContext)-1-int(lpb.stepsOut)]
+
+	// Validate that the parameter field index exists
```
benbellick: I think the meaning if this line is clear enough from the following error message on the next line, but I can go either way 🤷 
```suggestion
```

expr/builder.go
```diff
@@ -407,3 +510,56 @@ func (rb *fieldRefBuilder) BuildFuncArg() (types.FuncArg, error) {
...
+		Root:      LambdaParameterReference{StepsOut: lpb.stepsOut},
+		Reference: lpb.ref,
+		knownType: resolvedType,
+	}, nil
+}
+
+// getLambdaParamType gets the target lambda parameters struct from the context stack and validates that the parameter exists.
+// Returns an error if the parameter does not exist.
+func (lpb *lambdaParamRefBuilder) getLambdaParamType() (types.Type, error) {
+	// Validate stepsOut - check we have enough outer lambdas
```
benbellick:
```suggestion
```
I also think here it is clear enough from the error message, but feel free to take it or leave it.

types/func_type_test.go
benbellick: Can we actually drop this file and move it to `type_test.go`? We can add the roundtrip tests as just additional cases in:
https://github.com/substrait-io/substrait-go/blob/579861e290363a09304916963ddc61a8e7873698/types/types_test.go#L61

I think there is similarly already also a test for type short names that we can add one more case too. Maybe one for the long name though I haven't looked.

expr/builder.go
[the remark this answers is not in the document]
Slimsammylim: I added a better error message to include the steps out value. I also updated the builder error tests to search for this info in the error message.

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/stretchr/testify v1.10.0
 	github.com/substrait-io/substrait v0.78.1
```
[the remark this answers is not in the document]
benbellick: Okay... this is opening up a can of worms. If you do this, we will fail parsing these YAML files. To fix this, you then need to update this line: 
https://github.com/substrait-io/substrait-go/blob/579861e290363a09304916963ddc61a8e7873698/grammar/generate.go#L5
to v0.79.0

Then when you call `go generate`, it will update to use the new grammar which includes funcs. 

Then unfortunately, you'll need to update some more code.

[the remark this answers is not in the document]
benbellick: Unresolving this one, I think you still haven't called `go generate` to regenerate all of the code here

grammar/generate.go
vbarua: From in person discussion, we need to update these to point to the v0.79.0 tag as well.

expr/lambda_test.go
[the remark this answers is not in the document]
vbarua: +1 to roundtrip testing. As Ben says, they're not _as good_ as checking things individually like you've done, and can run into issues if plans don't roundtrip exactly the same (even if they do the same thing), but they're good enough and much easier to write and review.

expr/field_reference.go
```diff
@@ -29,6 +29,15 @@ type OuterReference uint32
 
 func (OuterReference) isRootRef() {}
 
+// LambdaParameterReference is a root reference type for accessing lambda parameters
+// within a lambda body expression. StepsOut indicates how many lambda boundaries
+// to traverse up (0 = current lambda, 1 = outer lambda, etc.)
+type LambdaParameterReference struct {
+	StepsOut uint32
+}
```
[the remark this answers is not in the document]
Slimsammylim: I agree that the struct approach is better. I think Ben's idea is good, and I can add it as a new issue (?)

expr/testdata/lambda/basic-lambda.json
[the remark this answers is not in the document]
Slimsammylim: I added new test cases using transform!

expr/lambda_test.go
```diff
@@ -0,0 +1,575 @@
...
+
+	lambda, err := b.Lambda(params,
+		b.LambdaParamRef(expr.StructFieldRef{Field: 5}, 0), // invalid - only 1 param (index 0)
+	).Build()
+
+	require.Error(t, err, "Build should fail for out-of-bounds field index")
+	require.Nil(t, lambda)
+	require.Contains(t, err.Error(), "trying to access parameter 5")
+	require.Contains(t, err.Error(), "only has 1 parameters")
+	t.Logf("Expected error: %v", err)
```
benbellick:
```suggestion
```

expr/lambda_test.go
```diff
@@ -0,0 +1,575 @@
...
+}
+
+// TestLambdaBuilder_TypeResolution verifies that GetType() returns the correct type
+// after building a lambda, including type resolution for LambdaParameterReferences.
+func TestLambdaBuilder_TypeResolution(t *testing.T) {
+	tests := []struct {
+		name         string
+		paramTypes   []types.Type
+		fieldIndex   int32
+		expectedType string
```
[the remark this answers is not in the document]
Slimsammylim: Fixed this!

PR #185, opened 2026-01-20
fix: resolve polymorphic types in struct, map, and UDT return types

### LINKED ISSUE #182: Polymorphic types (any, any1, any2, etc.) fail inside struct return types
When a function return type is `struct<any1>`, resolving the return type fails with:
```
error in struct field type: no matching any type found in function parameters
```

Other parameterized types like `list<any1>` work correctly - this issue is specific to structs.

## Reproduction

```yaml
urn: extension:test:any_bug
scalar_functions:
  - name: "identity"
    impls:
      - args:
          - name: value
            value: any1
        return: any1  # works

  - name: "wrap_in_list"
    impls:
      - args:
          - name: value
            value: any1
        return: list<any1>  # works

  - name: "wrap_in_struct"
    impls:
      - args:
          - name: value
            value: any1
        return: struct<any1>  # fails
```

### LINKED ISSUE #184: Polymorphic types (any, any1, any2, etc.) fail in user-defined type parameters
When a user-defined type has a polymorphic type parameter, resolving the return type fails with:
```
no matching any type found in function parameters
```

This is the same root cause as #182.

## Reproduction

```yaml
urn: extension:test:udt_any
types:
  - name: Wrapper
    structure:
      value: T
scalar_functions:
  - name: wrap
    impls:
      - args:
          - name: value
            value: any1
        return: Wrapper<any1>  # fails
```

benbellick:
Fixes the type resolution for polymorphic types (`any`, `any1`, `any2`, etc.) when used inside compound return types like `struct<any1>`, `map<any1, any2>`, and `u!Wrapper<any1>`.

Previously, `ReturnType` wasn't passing through `funcParams` and `argTypes` to nested types, so polymorphic types couldn't be resolved and would fail with "no matching any type found in function parameters".

Closes #182
Closes #184

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -508,14 +508,6 @@ type (
 		//
 		// funcParameters are the abstract parameter types from the function definition.
 		// argumentTypes are the concrete types passed to the function.
-		//
-		// For polymorphic types (any, any1, etc.): searches funcParameters to find itself,
-		// returns the corresponding concrete type from argumentTypes.
-		// For compound types (struct, list, map): recursively resolves nested types.
-		// For concrete types: returns itself unchanged.
-		//
-		// Note: The same type may appear in both parameter and return positions (e.g., any1
-		// as a parameter and nested inside struct<any1>). This interface covers both cases.
 		ReturnType(funcParameters []FuncDefArgType, argumentTypes []Type) (Type, error)
 
 		// WithParameters returns a new instance of this type with the given parameters.


types/parameterized_user_defined_type.go
```diff
@@ -26,6 +26,7 @@ func (d *DataTypeUDTParam) String() string {
 }
 
 func (d *DataTypeUDTParam) toTypeParam() (TypeParam, error) {
+	// TODO(#184): Pass funcParams and argTypes to support polymorphic types (any, any1, etc.)
```
benbellick: Wanted to keep the diff small and this one seemed a tiny bit more involved than the others.

benbellick: Actually went ahead and did this one, too

PR #192, opened 2026-02-02
chore: remove buf tooling for local protobuf generation

### LINKED ISSUE #193: Remove buf tooling for local protobuf generation
Let's remove `doc.go` and `buf.gen.yaml` because we do not need to generate protobuf locally. We can also remove the buf tooling.

Slimsammylim:
closes #193

This PR removes `doc.go` and `buf.gen.yaml` now that we do not need to generate protobuf files locally. I also updated the `README` to remove `buf` installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,4 +66,3 @@ considered stable.**
 
 Protobuf bindings come from [`substrait-protobuf`](https://github.com/substrait-io/substrait-protobuf). 
 
-Protobuf bindings come from [`substrait-protobuf`](https://github.com/substrait-io/substrait-protobuf). 


doc.go
```diff
@@ -1,9 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
-
-// Package substraitgo contains the experimental go bindings for substrait
-// (https://substrait.io).
-//
-// Current generated proto substrait version: v0.64.0
-package substraitgo
-
-//go:generate buf generate https://github.com/substrait-io/substrait.git#tag=v0.64.0
```
vbarua: I don't think we want to fully remove this file, because it's used to generate a (small) part of the package documentation: https://pkg.go.dev/github.com/substrait-io/substrait-go#pkg-overview

We take all the stuff about the protos out.

benbellick (approved):
LGTM!

PR #194, opened 2026-02-04
feat: update substrait-go to substrait v0.79.0

vbarua:


testcases/parser/visitor.go
```diff
@@ -427,7 +427,7 @@ func (v *TestCaseVisitor) VisitNullArg(ctx *baseparser.NullArgContext) interface
 
 func (v *TestCaseVisitor) VisitBooleanArg(ctx *baseparser.BooleanArgContext) interface{} {
 	value := strings.ToLower(ctx.BooleanLiteral().GetText()) == "true"
```
benbellick: This isn't this PR so no worries about fixing here, but seems a bit strange to treat everything that isn't "true" as false.

benbellick (approved):
LGTM, too!

PR #214, opened 2026-03-13
refactor: remove literalTypeInContext in TestCaseVisitor

vbarua:
The literalTypeInContext is difficult to reason about in terms of understanding
where it is set and reset.

It is used to smuggle type information through various tree nodes during literal
processing. Another way to accomplish this is to decompose node more explicitly
and not rely solely on the ANTLR generated visitation.

This allows us to replace a context variable with an explicit argument passed
through a series of functions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of table rows, column values, and literal lists using their declared types.
  * Correctly handles compact aggregate expressions, including column-count inference and invalid column indexes.
  * Improved validation for nested lists and incompatible or malformed elements.
  * Numeric and null literals now follow explicit element-type rules, while untyped numeric values remain strings.

### DIFF
--- a/testcases/parser/visitor.go
+++ b/testcases/parser/visitor.go
@@ -130,7 +130,7 @@ func (v *TestCaseVisitor) VisitSingleArgAggregateFuncCall(ctx *baseparser.Single
 }
 
 func (v *TestCaseVisitor) VisitCompactAggregateFuncCall(ctx *baseparser.CompactAggregateFuncCallContext) interface{} {
-	// Handles test of the form
+	// Handles tests of the form
 	//   ((20, 20), (-3, -3), (1, 1), (10,10), (5,5)) corr(col0::fp32, col1::fp32) = 1::fp64
 	//
 	// Types for values can be inferred from the type annotations on col* arguments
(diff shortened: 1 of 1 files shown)


testcases/parser/visitor.go
```diff
@@ -141,23 +130,43 @@ func (v *TestCaseVisitor) VisitSingleArgAggregateFuncCall(ctx *baseparser.Single
...
+
+	// Determine column count from the first row without processing literals.
+	allRows := ctx.TableRows().AllColumnValues()
+	var numberOfColumns int32
+	if len(allRows) > 0 {
+		numberOfColumns = int32(len(allRows[0].AllLiteral()))
+	}
+
+	// Build columnTypes array defaulting to nil (untyped columns stored as strings),
+	// then apply declared types from args.
```
vbarua: `nil` is used to signal that the data should be stored as string data. This is the _current_ behaviour which I've preserved, for now.

benbellick (approved):
Mostly LGTM. Just a couple comments about adding some comments to clarify why unreachable code is unreachable + a suggested small refactor to make the list processing code a bit easier to digest. Thanks!

benbellick:
@coderabbit review

Issue #217, opened 2026-03-23
Update go.mod module path from v7 to v8


Slimsammylim:
Lets update the go.mod module path from v7 to v8.

Issue #133, opened 2025-04-21
first major release version is always broken


vbarua:
The release script doesn't handle major versions well. Specifically, when it triggers a new major version release, it doesn't update the import paths so the first version of a major release always needs to be updated.

Amusingly, updating the import paths and then triggering a release via workflow doesn't work, because that results in the release of _another_ major version (because the import path change is breaking).

Example release with issue: https://github.com/substrait-io/substrait-go/tree/v4.0.0
PR to fix imports: https://github.com/substrait-io/substrait-go/pull/132

benbellick:
This happened again (didn't realize this was a problem 🤦) due to a release I kicked off last week without a module bump. @Slimsammylim has introduced a bump here: https://github.com/substrait-io/substrait-go/pull/218

PR #219, opened 2026-03-23
ci: add check to validate go.mod version matches git tags

benbellick:
Adds a preRelease hook to prevent releasing a major version without updating the module path, which breaks Go module resolution for consumers.

The script validates that the go.mod module path major version matches the release version computed by go-semantic-release. The `RELEASE_VERSION` env var is [set by go-semantic-release](https://github.com/Nightapes/go-semantic-release/blob/master/internal/hooks/hooks.go#L68) before running hooks:

```go
cmd.Env = append(cmd.Env, "RELEASE_VERSION="+h.version.Next.Version.String())
```

See the [hooks documentation](https://github.com/Nightapes/go-semantic-release#hooks) in the README. 

Note: the README example uses `- name: echo $RELEASE_VERSION` but the actual config expects a string array, not objects with a `name` field (https://github.com/Nightapes/go-semantic-release/issues/85).

Related to #217
Related to #133

---


.release.yml
```diff
@@ -5,6 +5,12 @@ release: 'github'
 github:
   repo: "substrait-go"
   user: "substrait-io"
+# preRelease hooks run after go-semantic-release computes the next version
+# but before the tag is created. RELEASE_VERSION env var contains the
```
vbarua, line 9: What sets `REALEASE_VERSION`?

vbarua (approved):

benbellick:
### Testing the script

#### Mismatched version (would block release)
```bash
$ RELEASE_VERSION=9.0.0 ./scripts/check-module-version.sh
ERROR: go.mod module path (v8) does not match RELEASE_VERSION (v9)

Module path: github.com/substrait-io/substrait-go/v8
RELEASE_VERSION: 9.0.0

To fix: update go.mod and all imports to use v9
See: https://go.dev/doc/modules/major-version

$ echo $?
1
```

#### Matching version (release proceeds)
```bash
$ RELEASE_VERSION=8.0.0 ./scripts/check-module-version.sh
OK: module version (v8) matches RELEASE_VERSION (v8)

$ echo $?
0
```

bvolpato (approved):
LGTM.

PR #221, opened 2026-04-09
fix(testcases): normalize decimal literal types to declared column type in aggregate args

benbellick:
When parsing aggregate test cases with typed decimal arguments (e.g. `1.00::dec<38,0>`), the decimal literal was created with natural precision/scale but never rescaled to the declared type. This caused type mismatches when building function invocations.

Fix: call `WithType` on the parsed decimal literal to rescale it to the declared `DecimalType` annotation.

### DIFF
--- a/testcases/parser/visitor.go
+++ b/testcases/parser/visitor.go
@@ -286,7 +286,9 @@ func (v *TestCaseVisitor) VisitTableRows(ctx *baseparser.TableRowsContext) inter
 func (v *TestCaseVisitor) VisitColumnValues(ctx *baseparser.ColumnValuesContext) interface{} {
 	values := make([]expr.Literal, 0, len(ctx.AllLiteral()))
 	for _, literalValue := range ctx.AllLiteral() {
-		values = append(values, v.Visit(literalValue).(expr.Literal))
+		if result := v.Visit(literalValue); result != nil {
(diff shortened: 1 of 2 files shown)


testcases/parser/visitor.go
```diff
@@ -792,7 +800,7 @@ func (v *TestCaseVisitor) VisitLiteral(ctx *baseparser.LiteralContext) interface
 			// in compactAggregateFuncCall context, the type is not set, full schema of table may not be available
 			return literal.NewString(ctx.NumericLiteral().GetText(), false)
 		}
-		value := v.getLiteralFromString(nil, ctx.NumericLiteral().GetText(), v.getLiteralTypeInContext())
+		value := v.getLiteralFromString(ctx, ctx.NumericLiteral().GetText(), v.getLiteralTypeInContext())
```
bvolpato, line 803: nit -- this now reports both the specific decimal failure and the generic `invalid numeric arg` for the same token. If easy, I'd keep just the specific one.

bvolpato (approved):
LGTM. Nice fix -- normalizing the aggregate decimal literals with `WithType` is the right place to do it, and the added tests are good. Thanks for the fix.

testcases/parser/visitor.go
benbellick: good catch!

PR #222, opened 2026-04-09
feat(parser): update to Substrait v0.87.0 with enum arg support

benbellick:
Upgrades the substrait dependency from v0.85.0 to v0.87.0 and adds parser support for the `enumArg` syntax introduced in that version (`IDENTIFIER::enum`, e.g. `SAMPLE::enum` in `std_dev`).

The interesting change is widening `CaseLiteral.Value` from `expr.Literal` to `types.FuncArg`. This is necessary because enum args don't map to an `Expression.Literal` in the substrait proto. `FunctionArgument` is a oneof of `{string enum, Type type, Expression value}`, making enum a sibling of a literal expression rather than a subtype. So `types.Enum` cannot implement `expr.Literal`, and the field needs to hold the broader `types.FuncArg` interface instead.

`std_dev` and `variance` test cases remain skipped because the v0.87.0 test files use an invalid single-column compact table format fixed in substrait v0.88.0 (substrait-io/substrait#1043).


bvolpato (approved):
LGTM -- nice writeup on the `CaseLiteral.Value` widening from `expr.Literal` to `types.FuncArg`, the reasoning around the proto oneof is spot on. Good that `extract.test` gets unblocked too.

One thing worth fixing before merge: in `testcases/parser/parse_test.go:436-438`, the new `if lit, ok := tc.AggregateArgs[0].Argument.Value.(expr.Literal); ok { assert.Equal(t, listType, lit.GetType()) }` silently no-ops if the cast fails. Previously that was an unconditional `assert.Equal(t, listType, tc.AggregateArgs[0].Argument.Value.GetType())`, so a regression would now pass the test instead of failing it. The value here is always a float32 list literal, so use `require.True(t, ok, ...)` like you already do at `parse_test.go:218-219`.

types/types.go
```diff
@@ -625,7 +625,7 @@ func (e *EnumType) MatchWithoutNullability(ot Type) bool {
 }
 
 func (e *EnumType) ShortString() string {
-	return "enum"
```
bvolpato, line 628: add a comment to this?

PR #228, opened 2026-04-14
feat(plan): add ExtensionTable method to Builder interface

benbellick:
Adds an ExtensionTable builder method so ExtensionTableReadRel can be constructed programmatically. Previously it was only obtainable via protobuf deserialization.


bvolpato (approved):
LGTM, thanks Ben!

PR #229, opened 2026-04-15
chore(deps): use tagged versions of dependencies

thaJeztah:
- relates to https://github.com/substrait-io/substrait-go/pull/139

The github.com/davecgh/go-spew and github.com/pmezard/go-difflib dependencies were updated to an untagged commit from their main branches in d202273a2169b71cca05f855a85c70ab2be69411, likely through an (indirect) dependency.

Both dependencies are either archived or dormant, and unlikely to do new releases.

This patch reverts both dependencies to their latest release to prevent forcing untagged releases on consumers of this module.

The this reverts the changes in the modules that were not yet released, butt those only impact CI and documentation:

- https://github.com/davecgh/go-spew/compare/v1.1.1...d8f796af33cc
- https://github.com/pmezard/go-difflib/compare/v1.0.0...5d4384ee4fb2


benbellick (approved):
I created https://github.com/substrait-io/substrait-go/issues/232 because we should be able to drop these dependencies. 

But for now, this works. Thanks!

thaJeztah:
Thanks!

oh! Yes, I recall I saw things in progress in testify, also to drop the old yaml dependency.

Mostly arrived in this repository because I went way too deep into cleaning up some of our transitive dependencies 😂

benbellick:
No worries, I appreciate it! It has also inspired me to setup depenadabot (https://github.com/substrait-io/substrait-go/pull/233) 😅

PR #230, opened 2026-04-15
chore(deps): remove use of deprecated gopkg.in/yaml/v3

thaJeztah:
The gopkg.in/yaml/v3 module has been archived and maintenance moved to a fork maintained by the YAML org.

This module already uses github.com/goccy/go-yaml as a dependency, which can be used as a replacement for these tests so that the direct dependency can be removed.


benbellick (approved):
LGTM! Thanks

PR #233, opened 2026-04-15
chore: add dependabot config

benbellick:
Adds weekly dependabot updates for Go modules and GitHub Actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly


thaJeztah:
Oh! Slight push-back from my side; or at least, it really depends on what this module is meant to be;

For "library" modules, it's good to deliberately keep dependencies boring and "slow";  Go modules are designed around  MVS (Minimal Version Selection); https://research.swtch.com/vgo-mvs

The idea is for a library module to specify the lowest possible versions of all dependencies (including Go); basically the minimum requirements.

With SemVer (assumed implemented correctly) that allows users of the module to pick any higher _compatible_ (same major version) of the dependency if they have a need to require a more minimal version.

Keeping the dependency versions low prevents an (indirect) dependency from forcing updates to consumers of the module, which sometimes can cause a ripple effect that's much larger than the set of dependencies defined in the library module.

For example, we maintain some small library modules in https://github.com/moby/sys, and we try to keep them as "boring" as possible; for some dependencies that means "the oldest tag that provides the functionality needed"; https://github.com/moby/sys/blob/bdb4ad64325dbd9abac686f89c91a8f0b8b52186/signal/go.mod#L3-L5

Nobody would use `golang.org/x/sys v0.1.0`, but "any version higher" can be used (but is not _forced_ to be used).

thaJeztah:
Oh, perhaps a good example of "bad"; this repository was a dependency we had for tests https://github.com/cloudfoundry/clock

The code literally hasn't changed in 9 or 10 Years, but it uses some test dependencies that get updated on a weekly cycle, and automatically released as a new version of the module, sometimes resulting in 1000s of lines of dependencies to be updated in our repository 🫠

benbellick:
@thaJeztah these are all great points :) Thanks for spelling that all out for me. Closing as a consequence!

PR #234, opened 2026-04-15
feat(testcases)!: model non-value results explicitly

benbellick:
SubstraitError held a free-form string that could be any value, but the Substrait test case spec only defines two non-value outcomes: `<!ERROR>` and `<!UNDEFINED>`. This replaces it with a sealed TestCaseResult interface and a NonValueOutcome enum, making the valid states explicit at the type level.

This changes `TestCase.Result` from `*CaseLiteral` to `TestCaseResult`. Callers that need concrete literal fields should type assert to `*CaseLiteral`.

I also added some docs :)

---


testcases/parser/nodes.go
```diff
@@ -20,17 +20,28 @@ const (
...
+	isTestCaseResult()
+}
+
+// CaseLiteral is a typed literal value used as a function argument or expected
+// result in a Substrait test case (e.g. "120::i8" or "NULL").
 type CaseLiteral struct {
-	Type           types.Type
-	ValueText      string
-	Value          types.FuncArg
-	SubstraitError *SubstraitError
```
benbellick, line 27: Notice that we are removing `SubstraitError` here. `CaseLiteral` is only for representing literals now.

testcases/parser/nodes.go
```diff
@@ -486,10 +494,27 @@ type CompactAggregateFuncCall struct {
...
+)
+
+func (e NonValueOutcome) String() string {
+	switch e {
+	case NonValueError:
+		return "<!ERROR>"
+	case NonValueUndefined:
+		return "<!UNDEFINED>"
+	default:
+		panic(fmt.Sprintf("unknown non-value outcome: %d", e))
```
benbellick, line 518: I am unsure if panicking is the right decision here. This would be indication of a user constructing an invalid `NonValueOutcome`.

testcases/parser/visitor.go
```diff
@@ -142,7 +142,7 @@ func (v *TestCaseVisitor) VisitSingleArgAggregateFuncCall(ctx *baseparser.Single
 	return &TestCase{
 		FuncName:      ctx.Identifier().GetText(),
 		AggregateArgs: []*AggregateArgument{{Argument: arg, ColumnType: arg.Type}},
-		Result:        &CaseLiteral{SubstraitError: &SubstraitError{Error: "uninitialized"}},
+		Result:        &CaseLiteral{},
```
kadinrabo: We should probably leave Result unset here because `VisitAggFuncTestCase` fills it immediately after visiting `AggFuncCall`. If not, `&CaseLiteral{}` looks like a real `NULL` literal if the overwrite ever stops happening

testcases/parser/visitor.go
```diff
@@ -849,17 +849,10 @@ func (v *TestCaseVisitor) VisitResult(ctx *baseparser.ResultContext) interface{}
...
-		err.Error = "UNDEFINED"
-	}
-	return &CaseLiteral{
-		Type:           nil,
-		ValueText:      ctx.GetText(),
-		SubstraitError: err,
+	if ctx.UndefineResult() != nil {
+		return NonValueUndefined
 	}
+	return NonValueError
```
kadinrabo, line 859: I know the grammar is closed over `ERROR`/`UNDEFINED` but an explicit check of `ctx.ErrorResult() != nil` before returning `NonValueError` would help make the invariant visible

kadinrabo (approved):
Core refactor looks clean left a few nits

testcases/parser/visitor.go
benbellick: this is a good point, thanks!

vbarua (approved):
This is much cleaner now 🧹

PR #235, opened 2026-04-21
fix(expr): initialize MapLiteral KeyValue entries before use

benbellick:
`MapLiteral.ToProtoLiteral` allocated `kv := make([]*proto.Expression_Literal_Map_KeyValue, len(t.Value))` (a slice of nil pointers) and then wrote to `kv[i].Key` / `kv[i].Value` without creating each element. Any non-empty map literal panicked with a nil pointer dereference when serialized.

### DIFF
--- a/expr/proto_literals_test.go
+++ b/expr/proto_literals_test.go
@@ -39,6 +39,30 @@ func TestToProtoLiteral(t *testing.T) {
 	}
 }
 
+func TestMapLiteralToProtoLiteral(t *testing.T) {
+	mapLit := NewNestedLiteral(MapLiteralValue{
+		{
+			Key:   NewPrimitiveLiteral("foo", false),
+			Value: NewPrimitiveLiteral[int32](1, false),
+		},
+		{
+			Key:   NewPrimitiveLiteral("bar", false),
+			Value: NewPrimitiveLiteral[int32](2, false),
+		},
+	}, false)
+
+	got := mapLit.ToProtoLiteral()
+
+	mapProto := got.GetMap()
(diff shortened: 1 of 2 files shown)


expr/proto_literals_test.go
```diff
@@ -39,6 +39,30 @@ func TestToProtoLiteral(t *testing.T) {
...
+		{
+			Key:   NewPrimitiveLiteral("bar", false),
+			Value: NewPrimitiveLiteral[int32](2, false),
+		},
+	}, false)
+
+	got := mapLit.ToProtoLiteral()
+
+	mapProto := got.GetMap()
+	assert.NotNil(t, mapProto, "expected Map literal type, got %T", got.LiteralType)
```
kadinrabo: Asserting the values would be good to verify the content survived serialization as well

expr/proto_literals_test.go
```diff
@@ -39,6 +39,30 @@ func TestToProtoLiteral(t *testing.T) {
 	}
 }
 
+func TestMapLiteralToProtoLiteral(t *testing.T) {
```
kadinrabo, line 42: Either style works but other tests used `cmp.Diff` for structural proto comparison if you wanted to keep it more consistent

kadinrabo (approved):

expr/proto_literals_test.go
benbellick: Good suggestion, I constructed the proto in place and am comparing use `cmp.Diff`. Thanks!

vbarua (approved):
Changes LGTM. Thanks for reviewing @kadinrabo

PR #241, opened 2026-05-06
fix(extensions): default ScalarFunctionImpl.Deterministic to true

### LINKED ISSUE #239: ScalarFunctionImpl.Deterministic defaults to false; spec says true
The [substrait spec](https://substrait.io/extensions/#scalar-function-properties) says `deterministic` defaults to `true` when omitted, but [`Deterministic bool` lacks the `default:` tag](https://github.com/substrait-io/substrait-go/blob/f2441bec8e196eef13667d2dd43141738eb685ac/extensions/simple_extension.go#L240) (compare to `Nullability` two lines down, which has `default:"MIRROR"`). On unmarshal, an absent yaml field hits Go's bool zero value, so every standard function reports `Deterministic() == false`.

```go
v, _ := DefaultCollection.GetScalarFunc(extensions.ID{
    URN:  "extension:io.substrait:functions_arithmetic",
    Name: "multiply:i64_i64",
})
fmt.Println(v.Deterministic()) // false — should be true
```

Suggested fix: `Deterministic bool ` + "`" + `yaml:",omitempty" default:"true"` + "`" + `.

SAY-5:
Fixes #239.

The [substrait spec](https://substrait.io/extensions/#scalar-function-properties) defines `deterministic` as defaulting to `true` when omitted, but the `Deterministic` field lacked a `default:` tag (compare to `Nullability` two lines down with `default:"MIRROR"`). On unmarshal an absent yaml field hits Go's bool zero value, so every standard scalar/aggregate/window function loaded from the default collection reported `Deterministic() == false`.

Repro:
```go
v, _ := extensions.GetDefaultCollectionWithNoError().GetScalarFunc(extensions.ID{
    URN:  "extension:io.substrait:functions_arithmetic",
    Name: "multiply:i64_i64",
})
fmt.Println(v.Deterministic()) // false on main, true with this patch
```

Test added asserts `Deterministic()` returns true alongside the existing `MirrorNullability` assertion in `TestCollection_GetAllScalarFunctions`.

### DIFF
--- a/extensions/simple_extension.go
+++ b/extensions/simple_extension.go
@@ -261,12 +261,8 @@ type ScalarFunction struct {
 	Metadata    map[string]any       `yaml:"metadata,omitempty"`
 }
 
-// UnmarshalYAML decodes a ScalarFunction and applies the substrait
-// spec default of `deterministic = true` to each impl, but only for
-// impls where the YAML omits the key. Doing this via a custom decode
-// (rather than a `default:"true"` struct tag) avoids clobbering an
-// explicit `deterministic: false`, which would otherwise be overwritten
-// because false is the zero value that creasty/defaults targets.
+// UnmarshalYAML decodes a ScalarFunction and applies the Substrait default
+// for omitted deterministic values.
 func (s *ScalarFunction) UnmarshalYAML(fn func(interface{}) error) error {
 	type rawImpl struct {
 		ScalarFunctionImpl `yaml:",inline"`
@@ -348,9 +344,8 @@ type AggregateFunction struct {
 	Metadata    map[string]any `yaml:"metadata,omitempty"`
 }
 
-// UnmarshalYAML mirrors ScalarFunction.UnmarshalYAML: it preserves
-// explicit `deterministic: false` on each impl while defaulting an
-// omitted key to true per the substrait spec.
+// UnmarshalYAML decodes an AggregateFunction and applies the Substrait default
+// for omitted deterministic values.
 func (s *AggregateFunction) UnmarshalYAML(fn func(interface{}) error) error {
 	type rawImpl struct {
 		AggregateFunctionImpl `yaml:",inline"`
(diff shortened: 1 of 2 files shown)


benbellick (commented):
Thanks for the contribution!

Can we add a test that validates YAML with `deterministic: false` preserves that value after unmarshalling/loading? Since `Deterministic` is a plain bool, I want to make sure applying `default:"true"` doesn’t treat explicit `false` the same as an omitted field.

SAY-5:
Good catch — I'll add a unit test that round-trips `deterministic: false` and confirms the value is preserved.

extensions/simple_extension.go
```diff
@@ -261,6 +261,42 @@ type ScalarFunction struct {
 	Metadata    map[string]any       `yaml:"metadata,omitempty"`
 }
 
+// UnmarshalYAML decodes a ScalarFunction and applies the substrait
+// spec default of `deterministic = true` to each impl, but only for
+// impls where the YAML omits the key. Doing this via a custom decode
+// (rather than a `default:"true"` struct tag) avoids clobbering an
+// explicit `deterministic: false`, which would otherwise be overwritten
+// because false is the zero value that creasty/defaults targets.
```
benbellick:
```suggestion
// UnmarshalYAML decodes a ScalarFunction and applies the Substrait default
// for omitted deterministic values.
```

extensions/simple_extension.go
```diff
@@ -362,6 +431,39 @@ type WindowFunction struct {
 	Metadata    map[string]any `yaml:"metadata,omitempty"`
 }
 
+// UnmarshalYAML mirrors ScalarFunction.UnmarshalYAML: it preserves
+// explicit `deterministic: false` on each impl while defaulting an
+// omitted key to true per the substrait spec.
```
benbellick:
```suggestion
// UnmarshalYAML decodes a WindowFunction and applies the Substrait default
// for omitted deterministic values.
```

extensions/simple_extension.go
```diff
@@ -312,6 +348,39 @@ type AggregateFunction struct {
...
```
benbellick:
```suggestion
// UnmarshalYAML decodes an AggregateFunction and applies the Substrait default
// for omitted deterministic values.
```

benbellick (approved):
I thought about suggesting to alter `bool` to be a ptr because it would simplify the unmarshalling logic, but that has the negative consequence of complicating the core library logic. 

Therefore, I think this change is a reasonable compromise. Thanks for your contribution!

PR #243, opened 2026-05-11
fix(testcases): reject non-nullable null literal types

benbellick:
Substrait test cases now require null literals to be explicitly typed as nullable. The parser was still accepting forms like `null::str` and normalizing them to `null::str?`, which hides invalid test cases instead of matching the upstream test case semantics.

This rejects explicitly typed null literals whose type is not nullable and updates the parser tests to use explicit nullable null types.

Related to substrait-io/substrait#988.

---


vbarua (commented):
Makes sense to bring this inline with the core spec. Could you add a test that captures the rejection case?

benbellick:
> Makes sense to bring this inline with the core spec. Could you add a test that captures the rejection case?

@vbarua [There's already one there](https://github.com/substrait-io/substrait-go/pull/243/files#diff-d18a3096856ca69e65e8993e53c49e14edab401eb545c3443f4f737faf5e572dR802) :)

testcases/parser/parse_test.go
```diff
@@ -799,6 +799,7 @@ func TestParseTestWithBadScalarTests(t *testing.T) {
 		{"add(123::fp32, 3.E.5::fp32) = 123::fp32", 17, "no viable alternative at input '3.E'"},
 		{"f1((1, 2, 3, 4)::i64) = 10::fp64", 0, "expected scalar testcase based on test file header, but got aggregate function testcase"},
 		{"add(4.53::dec<1, 0>, 0.25::dec<2, 2>) = 0.78::dec<5, 2>", 0, "Visit error at line 5: invalid argument number 4.53"},
+		{"f1(null::str) = null::str?", 0, "null literal must use a nullable type, got string"},
```
vbarua, line 802: Ah, missed this negative test case.

vbarua (approved):

PR #244, opened 2026-05-11
fix(types): preserve nested nullability in any binding

benbellick:
Function matching for constrained `any` parameters was ignoring nullability too broadly. Under the clarified Substrait semantics, MIRROR and DECLARED_OUTPUT strip only the outermost argument nullability before binding; nested nullability inside compound types still participates in structural matching.

This keeps cases like `list<i32>` and `list<i32?>` from binding to the same `list<any1>` parameter while still allowing outer nullable wrappers to match as intended. The added tests mirror the examples from substrait-io/substrait#943.

---

### DIFF
--- a/extensions/variants.go
+++ b/extensions/variants.go
@@ -681,33 +681,46 @@ func getLeafParameterizedParams(abstractTypes interface{}) []string {
 	panic("invalid non-leaf, non-parameterized type param")
 }
 
+func bindAnyType(anyType *types.AnyType, argType types.Type, bindings map[string]types.Type, ignoreNullability bool) error {
+	// Plain "any" is unconstrained - each occurrence can be a different type.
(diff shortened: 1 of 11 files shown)


dd-david-levin:
I opened #291 to cover a complementary return-substitution gap.

#244 correctly preserves/enforces nested nullability during matching, but `unwrapAnyTypeWithName` still returns the nullable concrete occurrence directly. As a result, `map(list<any1?>, list<any2?>) -> map<any1, any2?>` resolves as `map<string?, i64?>` rather than `map<string, i64?>`.

#291 normalizes a nullable `anyN?` source occurrence to the required base binding and lets the return occurrence reapply its own nullability. I also verified the #244 merge ref plus #291 with `go test ./...`; they pass together.

PR #247, opened 2026-05-21
feat(extensions): validate simple extension dependencies

benbellick:
This will be part of a multi-PR approach to handling cross-extension YAML dependencies. There is currently no implementation in this repo.

This PR contributes:
- parsing of the `dependencies` field in extension YAML files
- validation that dependency URNs are valid and already loaded
- [validation of dependency aliases per the spec](https://github.com/substrait-io/substrait/blob/21ab33c24057ccaf4b4101f9ceefdee14222f46d/text/simple_extensions_schema.yaml#L18)

Follow-up work will include parsing dependency-qualified type references such as `ext.u!point`, validating that used aliases map to declared dependencies, and resolving referenced types/type variations from dependency extension files.

---

### DIFF
--- a/extensions/extension_mgr.go
+++ b/extensions/extension_mgr.go
@@ -31,11 +31,18 @@ var (
 	}
 )
 
-var urnPattern = regexp.MustCompile(`^extension:[^:]+:[^:]+$`)
+var (
+	urnPattern             = regexp.MustCompile(`^extension:[^:]+:[^:]+$`)
+	dependencyAliasPattern = regexp.MustCompile(`^[A-Za-z_$][A-Za-z0-9_$]*$`)
+)
 
 // validateURN validates that a URN follows the format "extension:<owner>:<id>"
 func validateUrn(urn string) bool {
 	return urnPattern.MatchString(urn)
+}
+
+func validateDependencyAlias(alias string) bool {
+	return dependencyAliasPattern.MatchString(alias)
 }
 
 // GetDefaultCollectionWithNoError returns a Collection that is loaded with the default Substrait extension definitions.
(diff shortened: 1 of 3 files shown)


extensions/extension_mgr_test.go
```diff
@@ -595,6 +595,84 @@ scalar_functions:
 	assert.Contains(t, err.Error(), "invalid urn")
 }
 
+func TestLoadExtensionWithDependencies(t *testing.T) {
+	const dependency = `---
+urn: extension:test:dependency
+scalar_functions:
```
vbarua, line 597: meta: the empty `scalar_function` list here is not load-bearing right?

extensions/extension_mgr.go
```diff
@@ -31,13 +31,20 @@ var (
 	}
 )
 
-var urnPattern = regexp.MustCompile(`^extension:[^:]+:[^:]+$`)
+var (
+	urnPattern             = regexp.MustCompile(`^extension:[^:]+:[^:]+$`)
+	dependencyAliasPattern = regexp.MustCompile(`^[A-Za-z_$][A-Za-z0-9_$]*$`)
```
vbarua, line 35: This is based on https://github.com/substrait-io/substrait/blob/21ab33c24057ccaf4b4101f9ceefdee14222f46d/grammar/SubstraitLexer.g4#L119-L121, correct?

extensions/extension_mgr_test.go
```diff
@@ -595,6 +595,84 @@ scalar_functions:
...
+urn: extension:test:with_dependencies
+dependencies:
+  ext: extension:test:dependency
+  $ext: extension:test:dependency
+scalar_functions:
+`
+
+	var c extensions.Collection
+	require.NoError(t, c.Load("http://localhost/dependency.yaml", strings.NewReader(dependency)))
+	err := c.Load("http://localhost/test.yaml", strings.NewReader(extensionWithDependencies))
```
vbarua: At first I was confused about why were weren't using URNs here, but we have't fully gotten rid of URIs in this library yet have we?

vbarua (commented):
Parsin

extensions/extension_mgr_test.go
benbellick: This is a great point 🤦 

It turns out there is still some URI stuff lying around. I'm going to address that in a separate PR which we should merge in first. Then we can tackle this one. Thanks.

extensions/extension_mgr.go
benbellick: Yup, exactly

extensions/extension_mgr_test.go
benbellick: https://github.com/substrait-io/substrait-go/pull/250

extensions/extension_mgr_test.go
benbellick: Nope! I'll alter the test to have a fn there just to make it seem less suspect. Thanks

extensions/extension_mgr.go
```diff
@@ -217,6 +224,15 @@ func (c *Collection) Load(r io.Reader) error {
...
 	}
 
+	for alias, dependencyURN := range file.Dependencies {
+		if !validateDependencyAlias(alias) {
+			return fmt.Errorf("%w: invalid dependency alias %q", substraitgo.ErrInvalidSimpleExtention, alias)
+		}
+		if !c.URNLoaded(dependencyURN) {
+			return fmt.Errorf("%w: dependency urn %q for alias %q is not loaded", substraitgo.ErrNotFound, dependencyURN, alias)
+		}
+	}
```
benbellick, line 234: A previous iteration of this PR also checked that `dependencyURN` is a valid URN. But I realize now that such a check is redundant since only valid URNs would be loaded into the collection in the first place.

vbarua (approved):
LGTM. This feels like a pretty good check to have.

PR #249, opened 2026-05-21
fix(grammar): parse dependency-qualified UDT references

benbellick:
The extension docs allow YAML type expressions to reference user-defined types from dependencies using syntax like `ext.u!point`, and the examples already use that form. The type grammar only accepted unqualified `u!point`, so generated parsers rejected a syntax that the extension docs describe as valid.

This updates the type grammar to allow an optional dependency alias before `u!` and regenerates the Python parser used by the tests. The added grammar tests cover qualified UDTs directly, nested inside parameterized types, and as UDT parameters.

---

### DIFF
--- a/extensions/extension_mgr.go
+++ b/extensions/extension_mgr.go
@@ -31,11 +31,18 @@ var (
 	}
 )
 
-var urnPattern = regexp.MustCompile(`^extension:[^:]+:[^:]+$`)
+var (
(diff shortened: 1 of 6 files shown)


benbellick:
Closed because this PR was opened against the wrong repository. The upstream grammar change belongs in substrait-io/substrait.

PR #250, opened 2026-05-22
feat(extensions)!: drop loader URI tracking

benbellick:
Follow-up to #212. The protobuf API no longer has simple extension URI fields, but `Collection.Load` still accepted a source URI and the collection still kept a URI-to-URN map internally.

This removes that leftover loader URI state and makes loaded-extension identity fully URN-based. `Collection.Load` now reads the extension URN from the YAML file and duplicate detection happens only on that URN.

This is a breaking API cleanup because `Collection.Load` no longer takes a URI argument.

---

### DIFF
--- a/extensions/extension_mgr_test.go
+++ b/extensions/extension_mgr_test.go
@@ -88,11 +88,10 @@ aggregate_functions:
 `
 
 func TestLoadExtensionCollection(t *testing.T) {
-	const uri = "http://localhost/sample.yaml"
 	const urn = "extension:test:sample"
 
 	var c extensions.Collection
-	require.NoError(t, c.Load(uri, strings.NewReader(sampleYAML)))
+	require.NoError(t, c.Load(strings.NewReader(sampleYAML)))
 
 	t.Run("check types", func(t *testing.T) {
 		id := extensions.ID{URN: urn}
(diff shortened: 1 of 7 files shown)


extensions/extension_mgr_test.go
```diff
@@ -589,51 +585,31 @@ scalar_functions:
...
 
 	var c extensions.Collection
-	err := c.Load("http://localhost/test.yaml", strings.NewReader(extensionWithInvalidURN))
+	err := c.Load(strings.NewReader(extensionWithInvalidURN))
 
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "invalid urn")
 }
 
-func TestCannotLoadDuplicateURI(t *testing.T) {
```
vbarua, line 598: What happens now if you load the same extension twice?

vbarua (approved):
Changes make sense to me. Now that the URN is part of the extension content, we don't need to provide it like we did URIs.

extensions/extension_mgr_test.go
benbellick: [There is a test below checking the same thing for URNs
](https://github.com/substrait-io/substrait-go/pull/250/files#diff-1a55b0bd31f0a856c01e9482389b2491c0997476ccca2f5dda2222ba0e15bbcfR594)

PR #251, opened 2026-05-22
ci: disable scheduled releases

benbellick:
Stops the release workflow from running on the weekly schedule while keeping the configuration commented out so it can be restored easily later. Manual and PR-triggered runs remain available.

This is so we can batch lots of changes for the next release.


vbarua (approved):
LGTM. Let's break out the 🛠️

PR #252, opened 2026-05-22
feat(extensions): validate local UDT references

benbellick:
Before this, loading an extension file whose function referenced an undeclared UDT like `u!point` silently succeeded when failure is expected. 

The type was never added to the collection, so the function was left with a dangling reference to a type that doesn't exist anywhere.

This validates at load time that every UDT reference in a function signature resolves to a type declared in the same file's `types:` section, and errors otherwise.

References are pulled out by a new `types.ReferencedUserDefinedTypes` helper that sits next to the UDT type and walks args, returns, and aggregate/window intermediates, including UDTs nested in lists, maps, structs, function types, and type parameters.

The helper returns a `UserDefinedTypeReference` with a nil `DependencyAlias` for local refs. Dependency-qualified refs like `dep.u!name` are skipped for now. A later PR will set the alias and resolve those against a file's declared dependencies without touching this API.

---


types/parameterized_user_defined_type.go
```diff
@@ -231,3 +231,47 @@ func (m *ParameterizedUserDefinedType) WithParameters(params []interface{}) (Typ
...
+// ReferencedUserDefinedTypes returns every user-defined type referenced by t,
+// including those nested inside lists, maps, structs, function types, and
+// user-defined type parameters.
+func ReferencedUserDefinedTypes(t FuncDefArgType) []UserDefinedTypeReference {
+	var refs []UserDefinedTypeReference
+	collectUserDefinedTypes(t, &refs)
+	return refs
+}
+
+func collectUserDefinedTypes(t FuncDefArgType, refs *[]UserDefinedTypeReference) {
```
kadinrabo, line 253: I think this is missing a case for `OutputDerivation`. We probably need to recurse into `OutputDerivation.FinalType` before validating the refs.

extensions/validation.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+
+	substraitgo "github.com/substrait-io/substrait-go/v8"
+	"github.com/substrait-io/substrait-go/v8/types"
+)
+
+// validateUserDefinedTypeReferences ensures every user-defined type used in a
+// function signature is declared in that same file's types section. It does
+// not check type structures or type variation parents.
+func (s SimpleExtensionFile) validateUserDefinedTypeReferences() error {
```
kadinrabo: nit: `validateUserDefinedTypeReferences` and `functionTypes` use value receivers on `SimpleExtensionFile` which makes a copy. I think we can use pointer receivers

extensions/validation.go
```diff
@@ -0,0 +1,79 @@
...
+			if typ := arg.GetTypeExpression(); typ != nil {
+				out = append(out, typ)
+			}
+		}
+		if impl.Return != nil && impl.Return.ValueType != nil {
+			out = append(out, impl.Return.ValueType)
+		}
+	}
+
+	for _, fn := range s.ScalarFunctions {
```
kadinrabo, line 68: nit: the `AggregateFunctions` and `WindowFunctions` loops are almost identical, a shared helper would be nice

kadinrabo (approved):

types/parameterized_user_defined_type.go
benbellick: good catch, added + some tests!

extensions/extension_mgr_test.go
```diff
@@ -591,6 +591,144 @@ scalar_functions:
...
+        return: i32
+`
+
+	var c extensions.Collection
+	err := c.Load(strings.NewReader(extension))
+
+	assert.NoError(t, err)
+}
+
+func TestLoadExtensionRejectsUndeclaredTypeInWindowFunction(t *testing.T) {
```
vbarua: What do you think about grouping this next to `TestLoadExtensionWithUndeclaredUserDefinedType`, renaming that to `TestLoadExtensionRejectsUndeclaredTypeInScalarFunction`, and then adding an explicit `TestLoadExtensionRejectsUndeclaredTypeInAggregateFunction`. That way it's easy to verify that you're testing all three function types. It took me a bit to notice you didn't have an aggregate case because of the way these were laid out.

extensions/validation.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	// Local references are checked against this file's own declared types,
+	// before they are committed to the collection registry, so a failed load
+	// leaves the collection untouched.
+	//
+	// Dependency-qualified references (non-nil DependencyAlias) are skipped for
+	// now. A later PR will resolve them against the registry (c.GetType under
+	// the dependency's URN), since a dependency's types are already loaded by
+	// the time this file is.
```
vbarua, line 28: > since a dependency's types are already loaded by the time this file is.

Having seen this... we probably want to explicitly forbid dependency cycles in extensions.

vbarua (approved):
Looks good overall. One suggestion for your tests.

extensions/validation.go
benbellick: Agreed, opened a ticket: https://github.com/substrait-io/substrait/issues/1097

PR #256, opened 2026-06-04
feat: support new ComparisonJoinKey field in HashJoinRel/MergeJoinRel

nielspardon:
## Summary

The deprecated `left_keys`/`right_keys` fields of `HashJoinRel` and `MergeJoinRel` are being replaced upstream by `repeated ComparisonJoinKey keys = 8`. This PR migrates substrait-go to **consume both** the old and new proto fields and to **always produce** the new `keys` field, while **additionally mirroring** the deprecated fields when doing so is lossless (i.e. every key is a plain equality). This keeps it working against the current proto and ready once the deprecated fields are removed, without breaking consumers that haven't adopted `keys` yet — mirroring the approach used for the [URI → URN migration](https://substrait.io/spec/breaking_change_policy/#uri-to-urn-migration).

The fields were deprecated in Jan 2024 via https://github.com/substrait-io/substrait/pull/585 and thus deprecated in Substrait since v0.42.0.

## Changes

- Add a `ComparisonJoinKey` type modelling the proto message, including the `comparison` oneof (`SimpleComparison` / `CustomComparison`) and a `SimpleComparisonType` alias over the proto enum (consistent with how `WriteOp`/`OutputMode` are aliased).
- `HashJoinRel` / `MergeJoinRel` now store `keys []*ComparisonJoinKey` and expose `Keys()`. `LeftKeys()` / `RightKeys()` are kept as `// Deprecated:` derived accessors so existing consumers keep working.
- `RelFromProto` prefers `keys` and falls back to the deprecated fields (paired with a `SIMPLE_COMPARISON_TYPE_EQ` comparison) when `keys` is empty.
- `ToProto` always emits `keys`, and additionally emits the deprecated `left_keys`/`right_keys` when **every** key is a plain `SIMPLE_COMPARISON_TYPE_EQ`. `IS_NOT_DISTINCT_FROM`, `MIGHT_EQUAL`, and custom comparisons have no legacy encoding and an old consumer would silently read them as equality, so those joins emit only the modern `keys` field.
- Fix two pre-existing bugs in the `MergeJoin` read path: it constructed a `HashJoinRel` instead of a `MergeJoinRel`, and it reported `"error getting right input to HashJoin"` when its right input failed to decode.

## Tests

New `comparison_join_key_test.go` covering:
- lossless write policy: all-EQ joins mirror the deprecated fields, while non-EQ/custom joins emit only `keys`,
- deprecated accessor derivation from `keys` (both hash and merge),
- consume-legacy (old fields → EQ keys),
- precedence (`keys` wins when both old and new are set),
- full-fidelity round trip including a non-EQ simple comparison and a custom comparison function, with and without a post-join filter,
- error paths in `comparisonJoinKeysFromProto` / `joinKeyComparisonFromProto` (bad field refs, unset comparison, mismatched legacy lengths) exercised against both join types, plus a bad merge-join input.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plan/comparison_join_key_test.go
+++ b/plan/comparison_join_key_test.go
@@ -1,0 +1,354 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package plan
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/substrait-io/substrait-go/v8/expr"
+	"github.com/substrait-io/substrait-go/v8/extensions"
+	"github.com/substrait-io/substrait-go/v8/types"
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+// createJoinInput builds a named table read rel with three required int64 columns.
+func createJoinInput(name string) *NamedTableReadRel {
+	schema := types.NamedStruct{
+		Names: []string{"x", "y", "z"},
+		Struct: types.StructType{
+			Nullability: types.NullabilityRequired,
+			Types:       []types.Type{&types.Int64Type{}, &types.Int64Type{}, &types.Int64Type{}},
+		},
+	}
+	return &NamedTableReadRel{names: []string{name}, baseReadRel: baseReadRel{baseSchema: schema}}
+}
+
+func keyRef(t *testing.T, rel Rel, idx int32) *expr.FieldReference {
+	t.Helper()
+	base := rel.RecordType()
+	ref, err := expr.NewRootFieldRef(expr.NewStructFieldRef(idx), &base)
+	require.NoError(t, err)
+	return ref
+}
+
+func joinTestRegistry() expr.ExtensionRegistry {
+	return expr.NewEmptyExtensionRegistry(extensions.GetDefaultCollectionWithNoError())
+}
+
+// All joins below have a left and right input with three int64 columns each.
+func joinInputs() (Rel, Rel) {
+	return createJoinInput("L"), createJoinInput("R")
+}
+
+func eqKeys(t *testing.T, left, right Rel) []*ComparisonJoinKey {
+	return []*ComparisonJoinKey{
+		NewEqualityJoinKey(keyRef(t, left, 0), keyRef(t, right, 2)),
+		NewEqualityJoinKey(keyRef(t, left, 1), keyRef(t, right, 0)),
+	}
+}
+
+// Producing a join must only populate the new keys field, never the deprecated ones.
+func TestJoinProducesOnlyNewKeys(t *testing.T) {
+	left, right := joinInputs()
+	keys := eqKeys(t, left, right)
+
+	hash := (&HashJoinRel{left: left, right: right, joinType: HashMergeInner, keys: keys}).ToProto().GetHashJoin()
+	assert.Len(t, hash.GetKeys(), 2)
+	assert.Empty(t, hash.GetLeftKeys())
+	assert.Empty(t, hash.GetRightKeys())
+
+	merge := (&MergeJoinRel{left: left, right: right, joinType: HashMergeInner, keys: keys}).ToProto().GetMergeJoin()
+	assert.Len(t, merge.GetKeys(), 2)
+	assert.Empty(t, merge.GetLeftKeys())
+	assert.Empty(t, merge.GetRightKeys())
+}
+
+// The deprecated accessors derive their values from the keys.
+func TestJoinDeprecatedKeyAccessors(t *testing.T) {
+	left, right := joinInputs()
+	keys := eqKeys(t, left, right)
+	wantLeft := []*expr.FieldReference{keyRef(t, left, 0), keyRef(t, left, 1)}
+	wantRight := []*expr.FieldReference{keyRef(t, right, 2), keyRef(t, right, 0)}
+
+	hash := &HashJoinRel{left: left, right: right, joinType: HashMergeInner, keys: keys}
+	assert.Equal(t, wantLeft, hash.LeftKeys())
+	assert.Equal(t, wantRight, hash.RightKeys())
+
+	merge := &MergeJoinRel{left: left, right: right, joinType: HashMergeInner, keys: keys}
+	assert.Equal(t, wantLeft, merge.LeftKeys())
+	assert.Equal(t, wantRight, merge.RightKeys())
+}
+
+// A plan from a legacy producer (only deprecated fields set) is consumed and
+// mapped to keys with EQ comparisons.
+func TestJoinConsumesLegacyKeys(t *testing.T) {
+	left, right := joinInputs()
+	reg := joinTestRegistry()
+
+	for _, tc := range []struct {
+		name  string
+		build func(hj *proto.HashJoinRel) *proto.Rel
+		keys  func(rel Rel) []*ComparisonJoinKey
+	}{
+		{
+			name:  "hash",
+			build: func(hj *proto.HashJoinRel) *proto.Rel { return &proto.Rel{RelType: &proto.Rel_HashJoin{HashJoin: hj}} },
+			keys:  func(rel Rel) []*ComparisonJoinKey { return rel.(*HashJoinRel).Keys() },
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			legacy := tc.build(&proto.HashJoinRel{
+				Common: &proto.RelCommon{},
+				Left:   left.ToProto(),
+				Right:  right.ToProto(),
+				Type:   proto.HashJoinRel_JOIN_TYPE_INNER,
+				LeftKeys: []*proto.Expression_FieldReference{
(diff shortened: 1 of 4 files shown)


benbellick:
I think this PR will break things if a plan produced by this change contains e.g. `HashJoinRel` and the consumer hasn't been updated to a version that accepts the new `keys` field.

A safer policy here would be:

- Read both formats.
  - Prefer new `keys` when present.
  - Fall back to deprecated `left_keys` / `right_keys` when `keys` is empty.
  - Interpret legacy fields as `SIMPLE_COMPARISON_TYPE_EQ`.
- Write new `keys` always.
- Also write deprecated `left_keys` / `right_keys` when every key is exactly `SIMPLE_COMPARISON_TYPE_EQ`.
- Do not write deprecated fields for `IS_NOT_DISTINCT_FROM`, `MIGHT_EQUAL`, or custom comparisons, since the old fields cannot represent those semantics and an old consumer may silently interpret them as equality.

That keeps `keys` as the canonical modern representation while preserving compatibility for plain equality joins, which are the only joins the deprecated fields can represent safely.

This behavior of accepting either old or new but writing both when it is lossless is similar to how we handled the URI -> URN [migration](https://substrait.io/spec/breaking_change_policy/#uri-to-urn-migration).

nielspardon:
> That keeps `keys` as the canonical modern representation while preserving compatibility for plain equality joins, which are the only joins the deprecated fields can represent safely.
> 
> This behavior of accepting either old or new but writing both when it is lossless is similar to how we handled the URI -> URN [migration](https://substrait.io/spec/breaking_change_policy/#uri-to-urn-migration).

Done. I asked Claude to update the PR accordingly. Let me know if this is better. I'm also updating the Java PR in parallel along the same lines.

plan/comparison_join_key_test.go
```diff
@@ -0,0 +1,393 @@
...
+	left, right := joinInputs()
+	reg := joinTestRegistry()
+
+	for _, tc := range []struct {
+		name  string
+		build func(hj *proto.HashJoinRel) *proto.Rel
+		keys  func(rel Rel) []*ComparisonJoinKey
+	}{
+		{
+			name:  "hash",
```
benbellick: was there also supposed to be a merge join test here, too?

plan/comparison_join_key_test.go
```diff
@@ -0,0 +1,393 @@
...
+			build: func(hj *proto.HashJoinRel) *proto.Rel { return &proto.Rel{RelType: &proto.Rel_HashJoin{HashJoin: hj}} },
+			keys:  func(rel Rel) []*ComparisonJoinKey { return rel.(*HashJoinRel).Keys() },
```
benbellick: these two lines of tests seem a bit weird to me. Maybe this test would be better suited to being two tests, one for hash and one for merge, rather than making something table driven here.

plan/relations.go
```diff
@@ -1820,17 +1820,200 @@ const (
...
+	return out
+}
+
+// legacyJoinKeysToProto returns the deprecated left_keys/right_keys
+// representation of the given join keys, but only when every key is a plain
+// SIMPLE_COMPARISON_TYPE_EQ comparison. Those are the only joins the deprecated
+// fields can express; IS_NOT_DISTINCT_FROM, MIGHT_EQUAL and custom comparisons
+// have no legacy encoding and an old consumer would silently treat them as
+// equality, so for those we emit nothing and rely on the modern keys field.
+func legacyJoinKeysToProto(keys []*ComparisonJoinKey) (leftKeys, rightKeys []*proto.Expression_FieldReference) {
```
benbellick: Thinking out loud here, I wonder if it would be better if this also returned a `bool` or something which indicated whether or not we could make the conversion in a lossless way?

benbellick: Upon further thought, never mind! It doesn't really matter either way since there is no need to distinguish between "no keys" vs "lossy keys", as both will result in the same proto.

plan/relations.go
```diff
@@ -1820,17 +1820,200 @@ const (
...
+	}
+}
+
+// comparisonJoinKeysFromProto builds the join keys for a hash/merge join,
+// preferring the keys field. The deprecated leftKeys/rightKeys are only used
+// when keys is empty, in which case they are paired with an EQ comparison.
+func comparisonJoinKeysFromProto(
+	keys []*proto.ComparisonJoinKey,
+	leftKeys, rightKeys []*proto.Expression_FieldReference,
+	leftBase, rightBase *types.RecordType,
```
benbellick: nit: I think this would be a bit clearer.
```suggestion
	leftSchema, rightSchema *types.RecordType,
```

plan/relations.go
```diff
@@ -1862,22 +2054,15 @@ func (hr *HashJoinRel) SetAdvancedExtension(advExtension *extensions.AdvancedExt
...
-	keysLeft := make([]*proto.Expression_FieldReference, len(hr.leftKeys))
-	for i, k := range hr.leftKeys {
-		keysLeft[i] = k.ToProtoFieldRef()
-	}
-	keysRight := make([]*proto.Expression_FieldReference, len(hr.rightKeys))
-	for i, k := range hr.rightKeys {
-		keysRight[i] = k.ToProtoFieldRef()
-	}
-
+	leftKeys, rightKeys := legacyJoinKeysToProto(hr.keys)
```
benbellick: What do you think about instead doing
```suggestion
	leftKeys, rightKeys, ok := tryEqualityJoinKeysToLegacyProto(hr.keys)
	if ok {
	  ...
	}
```

I think that makes the intent a bit clearer. Not a blocker though.

plan/relations.go
```diff
@@ -1820,17 +1820,200 @@ const (
...
+// fields can express; IS_NOT_DISTINCT_FROM, MIGHT_EQUAL and custom comparisons
+// have no legacy encoding and an old consumer would silently treat them as
+// equality, so for those we emit nothing and rely on the modern keys field.
+func legacyJoinKeysToProto(keys []*ComparisonJoinKey) (leftKeys, rightKeys []*proto.Expression_FieldReference) {
+	for _, k := range keys {
+		simple, ok := k.comparison.(SimpleComparison)
+		if !ok || simple.Type != SimpleComparisonTypeEq {
+			return nil, nil
+		}
+	}
```
benbellick, line 1927:
```suggestion
	for _, k := range keys {
		switch simple := k.comparison.(type) {
		case SimpleComparison:
			if simple.Type != SimpleComparisonTypeEq {
				return nil, nil
			}
		case *SimpleComparison:
			if simple == nil || simple.Type != SimpleComparisonTypeEq {
				return nil, nil
			}
		default:
			return nil, nil
		}
	}
```

I think we need to match on the ptr case as well.

plan/comparison_join_key_test.go
```diff
@@ -0,0 +1,393 @@
...
+		Right:  &proto.Rel{}, // nil RelType -> RelFromProto errors
+	}}}
+
+	_, err := RelFromProto(rel, joinTestRegistry())
+	require.ErrorContains(t, err, "right input to MergeJoinRel")
+}
+
+// Exercises the join accessors and the ComparisonJoinKey getters, and covers
+// the post-join-filter branch of ToProto via a round trip.
+func TestJoinAccessorsAndPostJoinFilter(t *testing.T) {
```
benbellick: This test seems unrelated to the change here IMO, which is about the new comparison key and handling the migration gracefully. Can we remove it? It isn't obvious to me exactly what is being tested.

plan/comparison_join_key_test.go
```diff
@@ -0,0 +1,393 @@
...
+		})
+		t.Run("merge/"+tc.name, func(t *testing.T) {
+			_, err := RelFromProto(tc.fields.mergeProto(left, right), reg)
+			require.Error(t, err)
+		})
+	}
+}
+
+// A merge join whose right input fails to decode surfaces an error.
+func TestMergeJoinBadRightInput(t *testing.T) {
```
benbellick: I also feel like this test is outside of the scope of this PR. I'd be okay with removing this one too.

benbellick (approved):
left some small comments but otherwise looks good!

plan/comparison_join_key_test.go
nielspardon: there's now one test for hash and one for merge join as you suggested in the other comment

plan/comparison_join_key_test.go
nielspardon: there's now one test for hash and one for merge join as you suggested

plan/relations.go
nielspardon: done

plan/relations.go
nielspardon: done

plan/relations.go
nielspardon: done

plan/comparison_join_key_test.go
nielspardon: removed

plan/comparison_join_key_test.go
nielspardon: removed

benbellick:
I pushed one small fix to add an extra test for the new cases, but if all passes I'll merge. Hope you don't mind :)

PR #258, opened 2026-06-09
refactor(extensions)!: use distinct extension ID types

benbellick:
Extension IDs share the same URN/name shape, but function, type, and type variation IDs live in different namespaces. Modeling those namespaces with distinct Go types keeps callers from accidentally passing a function ID where a type ID is required while preserving the existing external shape.

---


extensions/extension_mgr.go
```diff
@@ -91,28 +92,25 @@ func loadExtensionFile(collection *Collection, substraitFS embed.FS, ent fs.DirE
...
+// FunctionID is the unique identifier for a Substrait extension function.
+type FunctionID = types.ExtensionFunctionID
+type TypeID = types.ExtensionTypeID
+type TypeVariationID = types.ExtensionTypeVariationID
 
 type Collection struct {
 	urnSet map[string]struct{}
 
-	simpleNameMap map[ID]string
+	simpleNameMap map[FunctionID]string
```
benbellick, line 97: I am not convinced we should have this `simpleNameMap` at all by the way. I am exploring what it would like like to remove it in another PR. But in theory, the below `scalarMap`, `aggregateMap`, and `windowMap` should be sufficient.

extensions/extension_mgr.go
```diff
@@ -91,28 +92,25 @@ func loadExtensionFile(collection *Collection, substraitFS embed.FS, ent fs.DirE
 	return nil
 }
 
-// ID is the unique identifier for a substrait object
-type ID struct {
-	URN string
-	// Name of the object. For functions, a simple name may be used for lookups,
-	// but as a unique identifier the compound name will be used
```
kadinrabo, line 98: nit: I think this comment was useful, maybe move it to `ExtensionFunctionID` so it's clear that the simple vs compound naming is specific to functions

extensions/extension_mgr.go
```diff
@@ -91,28 +92,25 @@ func loadExtensionFile(collection *Collection, substraitFS embed.FS, ent fs.DirE
...
-	Name string
-}
+// FunctionID is the unique identifier for a Substrait extension function.
+type FunctionID = types.ExtensionFunctionID
+type TypeID = types.ExtensionTypeID
+type TypeVariationID = types.ExtensionTypeVariationID
```
kadinrabo: nit: comments for `TypeID` and `TypeVariationID` might be useful like `FunctionID` has

kadinrabo (approved):
Overall looks good I'd just make the PR title more explicit this is a breaking change for consumers

extensions/extension_mgr.go
benbellick: Ah yeah I should add that back. I had ulterior motives, which was [this](https://github.com/substrait-io/substrait-go/pull/264/files) PR 😅 

But I should leave the comment in and only remove it in the other PR. Thanks!

extensions/extension_mgr.go
```diff
@@ -91,28 +92,29 @@ func loadExtensionFile(collection *Collection, substraitFS embed.FS, ent fs.DirE
...
 
-// ID is the unique identifier for a substrait object
-type ID struct {
-	URN string
-	// Name of the object. For functions, a simple name may be used for lookups,
-	// but as a unique identifier the compound name will be used
-	Name string
-}
+// FunctionID is the unique identifier for a Substrait extension function.
+type FunctionID = types.ExtensionFunctionID
```
vbarua: What's the benefit to aliasing `types.ExtensionFunctionID` to `extension.FunctionID` here?

I do like the `FunctionID` name more, as it reads better and is more succint in the code. Would there be a problem renaming `types.ExtensionFunctionID` to `types.FunctionID` and using it directly here?

types/extension_id.go
```diff
@@ -0,0 +1,20 @@
+package types
```
vbarua: Thinking about it, it's a little weird (to me) to have the Extension IDs types defined in the types package. The rest of the files in this package seem to deal with modelling Substrait types, but here we now have code related to functions as well.

A better place for these might be in the `extensions` packages. ExtensionIDs are definitionally associated with extensions, and the URN values in them only make sense in the context of extensions as well.

vbarua (commented):
Thanks for the review @kadinrabo

Expanding `extensions.ID` to `extension.FunctionID`, `extensions.TypeID`, etc makes sense to me given the number of entities that we have in Substrait. It's basically no extra cost given that we already require the use of the `extensions.ID` wrappers.

I did leave on pair of comments to the effect of "Should we move `extension_id.go` to the `extensions` package. Let me know what you think.

extensions/extension_mgr.go
benbellick: Yup that is totally fair. Good suggestion!

types/extension_id.go
benbellick: I agree with that suggestion as well. Let me try and move it around.

extensions/extension_id.go
```diff
@@ -0,0 +1,22 @@
...
+	// Name identifies the object within the extension file.
+	Name string
+}
+
+// TypeID is the unique identifier for a Substrait extension type.
+type TypeID ExtensionID
+
+// FunctionID is the unique identifier for a Substrait extension function.
+// For lookups, Name may be the simple function name. As a unique identifier,
+// Name is the compound function name.
```
benbellick, line 18: This is preserving current behavior. However, I would like to alter this to only refer to unique IDs. Working on this here https://github.com/substrait-io/substrait-go/pull/264

vbarua (approved):

PR #260, opened 2026-06-09
feat(literal)!: add NewMap, NewEmptyMap, NewEmptyList constructors

### LINKED ISSUE #236: literal: add NewMap helper
The `literal` package provides constructors for most types (`NewInt32`, `NewString`, `NewList`, etc.), but there is no `NewMap`. Callers have to hand-build an `expr.MapLiteralValue` and pass it through the generic `expr.NewLiteral`.

Let's add a `NewMap(entries expr.MapLiteralValue, nullable bool) (expr.Literal, error)` mirroring [`NewList`](https://github.com/substrait-io/substrait-go/blob/356db3548f6b6a858c49e634a635e21aede75f08/literal/utils.go#L404).

More broadly, the placement of the existing map/list constructors is inconsistent. [`expr.NewEmptyMapLiteral`](https://github.com/substrait-io/substrait-go/blob/356db3548f6b6a858c49e634a635e21aede75f08/expr/literals.go#L854) and [`expr.NewEmptyListLiteral`](https://github.com/substrait-io/substrait-go/blob/356db3548f6b6a858c49e634a635e21aede75f08/expr/literals.go#L866) live in `expr`, while [`literal.NewList`](https://github.com/substrait-io/substrait-go/blob/356db3548f6b6a858c49e634a635e21aede75f08/literal/utils.go#L404) lives in `literal`. All four of `NewMap`, `NewList`, `NewEmptyMap`, and `NewEmptyList` arguably belong in `literal`, which is where callers already look for literal constructors.

kadinrabo:
- Add `NewMap`, `NewEmptyMap`, and `NewEmptyList` literal constructors.
- Validate list elements and map entries using exact Substrait type equality.
- Preserve contextual element types when parsing collection literals.

Closes #236

BREAKING CHANGE: `NewList` now rejects elements with differing Substrait types, including nullability, parameters, variations, and nested types.


literal/utils.go
benbellick, line 1: We discussed offline, what you are doing in this PR is reasonable as it is consistent with the status quo. However, reading this has made me realize that there is a bit of a confusing precedent set in the codebase right now which I think we should clean up. 

Tried to capture the problem here: https://github.com/substrait-io/substrait-go/issues/266

literal/utils.go
```diff
@@ -401,28 +400,87 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
 	return expr.NewLiteral[expr.ListLiteralValue](elements, nullable)
 }
 
+// NewMap creates a map literal from the given key/value entries. As in NewList,
+// all keys must share one base type and all values one base type; null literals
+// are allowed in either position, and type parameters (such as varchar length)
+// and nullability may differ. The map's declared key/value types are taken from
+// the first entry. For an empty map use NewEmptyMap, which takes the key and
+// value types explicitly.
+func NewMap(entries expr.MapLiteralValue, nullable bool) (expr.Literal, error) {
```
benbellick, line 436: I think this current impl actually allows construction of invalid map literals:
```go
mapLiteral, err := literal.NewMap(expr.MapLiteralValue{
	{
		Key:   expr.NewNullLiteral(&types.StringType{Nullability: types.NullabilityNullable}),
		Value: literal.NewString("a", false),
	},
	{
		Key:   literal.NewInt8(1, false),
		Value: literal.NewString("b", false),
	},
}, false)
```

This succeeds and produces a map declared as `map<string?, string>`, even though the second key is an `i8` literal.

literal/utils.go
```diff
@@ -401,28 +400,87 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
+// is compatible. Null literals are allowed in any position, and type parameters
+// (such as varchar length) and nullability may differ within one base type; a
+// later cast unifies those. Base types are compared by short name rather than
+// by the Go wrapper, which is too coarse: distinct types such as decimal and
+// varchar share one wrapper. The comparison is shallow and matches the original
+// wrapper check's depth: it does not distinguish the element types of nested
+// lists, maps, or structs, nor different user-defined types (whose short name is
+// empty), so a column mixing, for example, list<i8> and list<i32> is treated as
+// uniform.
+func firstMismatch(column []expr.Literal) int {
```
benbellick: this fn may be a bit more explicit if we instead had it return `(int, bool)`. That way we don't have to depend on the `-1` value to have significance.

literal/utils.go
```diff
@@ -401,28 +400,87 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
+// marked nullable or not.
+func NewEmptyMap(keyType, valueType types.Type, nullable bool) expr.Literal {
+	return expr.NewEmptyMapLiteral(keyType, valueType, nullable)
+}
+
+// NewEmptyList creates an empty list literal of the given element type, marked
+// nullable or not.
+func NewEmptyList(elementType types.Type, nullable bool) expr.Literal {
+	return expr.NewEmptyListLiteral(elementType, nullable)
+}
```
benbellick: I believe both of these accept `nil` as the first argument. Only later if you try and turn them `ToProto` do they actually error. Instead, we should probably check that here if we are considering these the "nice" cstrs that check for validity (though I think we will reconsider that approach in https://github.com/substrait-io/substrait-go/issues/266).

literal/utils_test.go
```diff
@@ -959,6 +969,128 @@ func TestNewList(t *testing.T) {
...
+	nullableMapType := &types.MapType{Key: int8Type, Value: stringType, Nullability: types.NullabilityNullable}
+	// differing varchar lengths keep the first entry's parameters
+	varCharValMapType := &types.MapType{Key: int8Type, Value: varChar1Type, Nullability: types.NullabilityRequired}
+	varCharKeyMapType := &types.MapType{Key: varChar1Type, Value: stringType, Nullability: types.NullabilityRequired}
+	// the key/value types come from the first entry, so a leading null literal
+	// keeps that column nullable
+	nullValueMapType := &types.MapType{Key: int8Type, Value: nullStringType, Nullability: types.NullabilityRequired}
+	nullKeyMapType := &types.MapType{Key: nullInt8Type, Value: stringType, Nullability: types.NullabilityRequired}
+
+	tests := []struct {
```
benbellick: IMO, these tests are a bit easier to follow if we instead break them into a group of tests we expect to pass and another group of tests we expect to error. The current test struct tries to capture both.

literal/utils_test.go
```diff
@@ -959,6 +969,128 @@ func TestNewList(t *testing.T) {
...
+		{"crossFamilyKeys", expr.MapLiteralValue{{Key: dec1, Value: strVal1}, {Key: vc1, Value: strVal2}}, false, false, nil, assert.Error},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := NewMap(tt.entries, tt.nullable)
+			if !tt.wantErr(t, err, fmt.Sprintf("NewMap(%v)", tt.entries)) {
+				return
+			}
+			if tt.expSuccess {
+				want, err := expr.NewLiteral[expr.MapLiteralValue](tt.entries, tt.nullable)
```
benbellick: This is basically delegating to the underlying constructor. I think it would be more useful to construct the desired `want` values by hand and compare against those.

literal/utils_test.go
```diff
@@ -959,6 +969,128 @@ func TestNewList(t *testing.T) {
...
+}
+
+func TestNewEmptyMap(t *testing.T) {
+	tests := []struct {
+		name      string
+		keyType   types.Type
+		valueType types.Type
+		nullable  bool
+	}{
+		{"i8ToString", &types.Int8Type{Nullability: types.NullabilityRequired}, &types.StringType{Nullability: types.NullabilityRequired}, false},
```
benbellick: IMO since `NewEmptyMap` is such a shallow function impl, just one of these cases is sufficient.

literal/utils_test.go
```diff
@@ -959,6 +969,128 @@ func TestNewList(t *testing.T) {
...
+				wantNullability = types.NullabilityNullable
+			}
+			wantType := &types.MapType{Nullability: wantNullability, Key: tt.keyType, Value: tt.valueType}
+			assert.Equalf(t, wantType, got.GetType(), "NewEmptyMap(%v, %v)", tt.keyType, tt.valueType)
+			assert.Equalf(t, expr.NewEmptyMapLiteral(tt.keyType, tt.valueType, tt.nullable), got, "NewEmptyMap(%v, %v)", tt.keyType, tt.valueType)
+		})
+	}
+}
+
+func TestNewEmptyList(t *testing.T) {
```
benbellick: Same here

literal/utils_test.go
```diff
@@ -959,6 +969,128 @@ func TestNewList(t *testing.T) {
...
+			if tt.nullable {
+				wantNullability = types.NullabilityNullable
+			}
+			wantType := &types.ListType{Nullability: wantNullability, Type: tt.elementType}
+			assert.Equalf(t, wantType, got.GetType(), "NewEmptyList(%v)", tt.elementType)
+			assert.Equalf(t, expr.NewEmptyListLiteral(tt.elementType, tt.nullable), got, "NewEmptyList(%v)", tt.elementType)
+		})
+	}
+}
+
```
benbellick, line 1080: However, we probably want to check the `nil` argument cases I mentioend above for these empty X constructors.

literal/utils.go
kadinrabo: Agreed. I kept this PR scoped to matching the current model

literal/utils.go
kadinrabo: Sounds good!

literal/utils_test.go
kadinrabo: Can do 👍

literal/utils_test.go
kadinrabo: Reduced to one case

literal/utils_test.go
kadinrabo: Good catch, add some nil key/value type coverage

literal/utils.go
kadinrabo: I'll compare literal base types directly to avoid this

literal/utils.go
kadinrabo: I'll make them reject nil type arguments immediately for now

literal/utils_test.go
kadinrabo: That's fair I'll build the expected map literals directly

literal/utils.go
```diff
@@ -401,28 +400,89 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
+// literal is compatible. Null literals are allowed in any position when their
+// typed base matches the anchor. Type parameters (such as varchar length) and
+// nullability may differ within one base type; a later cast unifies those. Base
+// types are compared by short name rather than by the Go wrapper, which is too
+// coarse: distinct types such as decimal and varchar share one wrapper. The
+// comparison is shallow and matches the original wrapper check's depth: it does
+// not distinguish the element types of nested lists, maps, or structs, nor
+// different user-defined types (whose short name is empty), so a column mixing,
+// for example, list<i8> and list<i32> is treated as uniform.
+func firstMismatch(column []expr.Literal) (int, bool) {
```
benbellick: You are only comparing type short names here, which means that nested types will incorrectly match. For example, I believe the following does not error, but should:
```go
      innerI8, err := NewList([]expr.Literal{NewInt8(1, false)}, false)
      if err != nil {
         t.Fatal(err)
      }
      innerI32, err := NewList([]expr.Literal{NewInt32(2, false)}, false)
      if err != nil {
         t.Fatal(err)
      }

      NewList([]expr.Literal{innerI8, innerI32}, false)
```

literal/utils.go
```diff
@@ -401,28 +400,89 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
 	}
 	return expr.NewLiteral[expr.ListLiteralValue](elements, nullable)
 }
 
+// NewMap creates a map literal from the given key/value entries. As in NewList,
+// all keys must share one base type and all values one base type; null literals
+// are allowed in either position, and type parameters (such as varchar length)
+// and nullability may differ. The map's declared key/value types are taken from
+// the first entry. For an empty map use NewEmptyMap, which takes the key and
+// value types explicitly.
```
benbellick: Hmm this comment makes it seem like the behavior I noticed above is intentional. But in substrait `list<i32>` is not the same as `list<i64>`, so I think we would be better of treating those as distinct.

benbellick (changes requested):
Thanks! I think we may need to resolve the issue of comparing base types before we can merge. Please let me know if I said something unclear.

literal/utils.go
```diff
@@ -401,28 +400,153 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
-		return nil, fmt.Errorf("empty list literal")
-	}
-	anchorType := reflect.TypeOf(elements[0])
-	for i, e := range elements {
-		currentType := reflect.TypeOf(e)
-		if currentType != anchorType {
-			nullLiteralType := reflect.TypeOf((*expr.NullLiteral)(nil))
-			if currentType == nullLiteralType {
-				continue
+// firstMismatch returns the index of the first nil literal or literal whose base
```
benbellick: Perhaps there is some complexity here that I am not quite understanding, but it seems like a lot of this code is devoted to fairly complex recursive calls into type constructors. 

Is there any reason why we can't use standard [type comparison directly](https://github.com/substrait-io/substrait-go/blob/4a512e6f8750ec91fe5d8e8afd4d67cf9f58a8f2/types/types.go#L450)? 

Then for e.g. List, it becomes as simple as iterating over all of the literals and enforcing that they all have the same type (without any reference to this notion of "base type"). Is there a reason that wouldn't be sufficient?

benbellick: As a concrete example, the current PR accepts:

```text
[2::i8, 3::i8?, null::i8?]::list<i8>
```

The element type is `i8` because it is taken from the first element, even though the list contains a null. Reordering the same elements can instead produce `list<i8?>`.

If the goal is to support coercion from `i8` to `i8?`, that could be useful, but the rules become complex quickly, particularly for nested types. I think it would be safer to require exact type equality for now (`i8 != i8?`) and consider coercion separately later.

benbellick (changes requested):

literal/utils.go
```diff
@@ -401,28 +401,121 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
-				continue
+// firstMismatch returns the index of the first nil literal or literal whose type
+// differs from the column's anchor (the first literal), or false if every
+// literal has the same type.
+func firstMismatch(column []expr.Literal) (int, bool) {
+	if column[0] == nil {
+		return 0, true
+	}
+	anchor := column[0].GetType()
+	if !validType(anchor) {
```
benbellick: IMO this is too defensive. At this point, we need to be able to assume that the types are valid. And regardless, there is a lot more that goes into validating a type then what exists here. 

I think we should assume the types are valid and then from there we can work on https://github.com/substrait-io/substrait-go/issues/266 in a separate PR to make sure this is definitely the case.

literal/utils.go
benbellick:
```suggestion
func hasTypeMistmatch(literals []expr.Literal) (int, bool) {
```
I think this is a slightly better name bc it makes it clear that we are comparing types. Also this list of types doesn't really relate to columns.

literal/utils.go
```diff
@@ -401,28 +401,121 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
+		if !validType(current) || !anchor.Equals(current) {
+			return i + 1, true
+		}
+	}
+	return 0, false
+}
+
+// validType checks nested type components before calling Type.Equals, whose
+// implementations assume that collection and user-defined types are complete.
+func validType(t types.Type) bool {
```
benbellick: So as mentioned above, I think we should drop `validType` and `isNil`.

literal/utils.go
benbellick: With that assumption, the code becomes much simpler:
```go
func hasTypeMismatch(literals []expr.Literal) (int, bool) {
	if literals[0] == nil {
		return 0, true
	}

	expectedType := literals[0].GetType()
	for i, literal := range literals[1:] {
		if literal == nil || !expectedType.Equals(literal.GetType()) {
			return i + 1, true
		}
	}
	return 0, false
}
```

literal/utils_test.go
```diff
@@ -937,11 +980,27 @@ func TestNewList(t *testing.T) {
 		wantErr    assert.ErrorAssertionFunc
 	}{
```
benbellick, line 939: IMO, this is a bit too many tests. I can understand the urge to add a lot of tests, but it makes it harder for the reviewer to isolated where the new change is being tested. I think adding a single test for each of:
- nil element present in list
- a list of params with diff types
- a list of params with the same type but differing elements
- a list of params with same base type but differing nested type (e.g. `list<i8>` and `list<i32>`)

is perfectly sufficient. No need to test the inner details of type equality by throwing UDTs into the mix.

literal/utils_test.go
```diff
@@ -959,6 +1018,141 @@ func TestNewList(t *testing.T) {
...
+		require.Error(t, err)
+		assert.Nil(t, got)
+	})
+
+	t.Run("nilValueType", func(t *testing.T) {
+		got, err := NewEmptyMap(keyType, nil, false)
+		require.Error(t, err)
+		assert.Nil(t, got)
+	})
+}
```
benbellick: This feels redundant with the above list

literal/utils_test.go
```diff
@@ -959,6 +1018,141 @@ func TestNewList(t *testing.T) {
...
+		},
+	}
+	assert.Equal(t, want, got)
+
+	t.Run("nilElementType", func(t *testing.T) {
+		got, err := NewEmptyList(nil, false)
+		require.Error(t, err)
+		assert.Nil(t, got)
+	})
+}
```
benbellick: this does too

testcases/parser/visitor.go
```diff
@@ -820,8 +868,12 @@ func (v *TestCaseVisitor) VisitLiteral(ctx *baseparser.LiteralContext) interface
...
 	if ctx.StringLiteral() != nil {
 		valueStr := getRawStringFromStringLiteral(ctx.GetText())
-		value := literal.NewString(valueStr, false)
-		return value
+		if literalType := v.getLiteralTypeInContext(); literalType != nil {
+			if value := v.getLiteralFromString(ctx, valueStr, literalType); value != nil {
+				return value
+			}
+		}
+		return literal.NewString(valueStr, false)
```
benbellick, line 880: As an aside, it would be helpful to understand why we had to have code changes here. I would expect that the parser code is fairly independent? 

I haven't looked to closely yet, but just something that caught my eye.

benbellick (commented):
Sorry I had to run before I could finish but left a few more comments. Thanks!

literal/utils_test.go
```diff
@@ -959,6 +962,91 @@ func TestNewList(t *testing.T) {
...
+
+	int8Type := &types.Int8Type{Nullability: types.NullabilityRequired}
+	nullInt8Type := &types.Int8Type{Nullability: types.NullabilityNullable}
+	stringType := &types.StringType{Nullability: types.NullabilityRequired}
+	nullStringType := &types.StringType{Nullability: types.NullabilityNullable}
+	mapType := &types.MapType{Key: int8Type, Value: stringType, Nullability: types.NullabilityRequired}
+	nullableMapType := &types.MapType{Key: int8Type, Value: stringType, Nullability: types.NullabilityNullable}
+	nullableEntriesMapType := &types.MapType{Key: nullInt8Type, Value: nullStringType, Nullability: types.NullabilityRequired}
+
+	successTests := []struct {
```
benbellick, line 1022: This is technically inconsistent with how lists are tested above, but this test is wayy better IMO. I'll open an issue for making the list test separate out success vs failures cases like this one, so leave it just like this. Thanks!

literal/utils_test.go
```diff
@@ -959,6 +962,91 @@ func TestNewList(t *testing.T) {
...
+	vc4, _ := NewVarChar("abcd", false)
+	dec1, _ := NewDecimalFromString("1.0", false)
+	i8ListVal, _ := NewList([]expr.Literal{i8Key1}, false)
+	i32ListVal, _ := NewList([]expr.Literal{i32Val1}, false)
+
+	int8Type := &types.Int8Type{Nullability: types.NullabilityRequired}
+	nullInt8Type := &types.Int8Type{Nullability: types.NullabilityNullable}
+	stringType := &types.StringType{Nullability: types.NullabilityRequired}
+	nullStringType := &types.StringType{Nullability: types.NullabilityNullable}
+	mapType := &types.MapType{Key: int8Type, Value: stringType, Nullability: types.NullabilityRequired}
```
benbellick:
```suggestion
	mapI8StrType := &types.MapType{Key: int8Type, Value: stringType, Nullability: types.NullabilityRequired}
```
better for the name to be a bit more meaningful

literal/utils_test.go
```diff
@@ -959,6 +962,91 @@ func TestNewList(t *testing.T) {
...
+	nullableMapType := &types.MapType{Key: int8Type, Value: stringType, Nullability: types.NullabilityNullable}
```
benbellick: and here
```suggestion
	nullableMapI8StrType := &types.MapType{Key: int8Type, Value: stringType, Nullability: types.NullabilityNullable}
```

literal/utils.go
```diff
@@ -401,28 +400,75 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
+	}
+	if i, ok := hasTypeMismatch(values); ok {
+		return nil, fmt.Errorf("value %d of map literal has invalid or different type", i)
+	}
+	return expr.NewLiteral[expr.MapLiteralValue](entries, nullable)
+}
+
+// NewEmptyMap creates an empty map literal of the given key and value types,
+// marked nullable or not.
+func NewEmptyMap(keyType, valueType types.Type, nullable bool) (expr.Literal, error) {
```
benbellick, line 456: minor nit: there is no test that calls `NewEmptyMap`

literal/utils.go
```diff
@@ -401,28 +400,75 @@ func NewPrecisionTimestampTzFromString(precision types.TimePrecision, value stri
...
+	}
+	if valueType == nil {
+		return nil, fmt.Errorf("empty map literal value type cannot be nil")
+	}
+	return expr.NewEmptyMapLiteral(keyType, valueType, nullable), nil
+}
+
+// NewEmptyList creates an empty list literal of the given element type, marked
+// nullable or not.
+func NewEmptyList(elementType types.Type, nullable bool) (expr.Literal, error) {
```
benbellick, line 468: or any that calls `NewEmptyList`. 

Just a simple test of the happy path is good enough for me.

testcases/parser/parse_test.go
```diff
@@ -375,6 +375,23 @@ some_func([[1, null, 2], [3, 4]]::list<list<i32?>>) = true::bool
...
+func TestParseNullablePrimitiveListLiterals(t *testing.T) {
+	header := makeHeader("v1.0", "/extensions/functions_list.yaml")
+	tests := `# nullable primitive lists
+some_func([true, null]::list<bool?>) = true::bool
+some_func(['1991-01-01', null]::list<date?>) = true::bool
+some_func(['13:01:01.234', null]::list<time?>) = true::bool
+some_func(['1991-01-01T01:02:03.456', null]::list<timestamp?>) = true::bool
+some_func(['1991-01-01T01:02:03.456+00:00', null]::list<timestamp_tz?>) = true::bool
+some_func(['P10Y5M', null]::list<interval_year?>) = true::bool
+some_func(['P10DT5H6M7S', null]::list<interval_day?>) = true::bool
```
benbellick, line 389:
```suggestion
some_func(['P10DT5H6M7S', null]::list<interval_day?>) = true::bool
some_func(['a', null]::list<string?>) = true::bool
some_func(['a', null]::list<binary?>) = true::bool
```
adding these test would be good to show string and binary type.

benbellick (approved):
few more comments but then otherwise looks good!

PR #264, opened 2026-06-09
feat(extensions)!: require function variant names

benbellick:
`NewScalarFunc`, `NewAggregateFunc`, `NewWindowFunc`, collection lookups, and plan builder function refs now require function IDs to name a registered function variant.

Previously, these APIs accepted an `extensions.FunctionID` whose `Name` could be only the simple function name. The constructor then tried to infer the registered variant from the argument types.

```go
reg := expr.NewEmptyExtensionRegistry(extensions.GetDefaultCollectionWithNoError())

// Previously accepted because NewScalarFunc resolved "add" to a concrete
// variant such as "add:i32_i32" from the argument types.
id := extensions.FunctionID{
    URN:  extensions.SubstraitDefaultURNPrefix + "functions_arithmetic",
    Name: "add",
}

_, err := expr.NewScalarFunc(reg, id, nil,
    expr.NewPrimitiveLiteral(int32(1), false),
    expr.NewPrimitiveLiteral(int32(2), false),
)
```

That behavior made ID-based APIs do two jobs. They accepted an ID, but they also performed simple-name overload resolution. This made a convenience behavior permeate the core function structs and lookup paths, even though an ID should identify an already-resolved extension object.

With this PR, `FunctionID` uses `Signature` instead of `Name`, and callers must provide the full registered variant signature.

```go
reg := expr.NewEmptyExtensionRegistry(extensions.GetDefaultCollectionWithNoError())

// Required now. The ID names the exact registered function variant.
id := extensions.FunctionID{
    URN:       extensions.SubstraitDefaultURNPrefix + "functions_arithmetic",
    Signature: "add:i32_i32",
}

_, err := expr.NewScalarFunc(reg, id, nil,
    expr.NewPrimitiveLiteral(int32(1), false),
    expr.NewPrimitiveLiteral(int32(2), false),
)
```

The same rule applies to aggregate and window functions, collection lookups, and plan builder function refs. This PR also adds `Collection.IsRegisteredFunction` so callers can check whether an ID names a registered scalar, aggregate, or window function variant before asking a plan builder for a function ref.

```go
builder := plan.NewBuilderDefault()

// Rejected now because "equal" is only the simple function name.
_, err := builder.GetFunctionRef(extensions.FunctionID{
    URN:       extensions.SubstraitDefaultURNPrefix + "functions_comparison",
    Signature: "equal",
})

// Accepted because "equal:any_any" is the registered variant signature.
ref, err := builder.GetFunctionRef(extensions.FunctionID{
    URN:       extensions.SubstraitDefaultURNPrefix + "functions_comparison",
    Signature: "equal:any_any",
})
```

This is a breaking change for callers that relied on the previous implicit resolution behavior.

A simple-name convenience API is still useful, but it should be explicit and opt in. This PR adds resolver helpers for that path:

```go
// ResolveScalarFunctionVariant resolves a simple scalar function name and arguments
// to the ID of a registered function variant.
func ResolveScalarFunctionVariant(
    reg expr.ExtensionRegistry,
    urn string,
    name string,
    args ...types.FuncArg,
) (extensions.FunctionID, error)

// ResolveAggregateFunctionVariant resolves a simple aggregate function name and arguments
// to the ID of a registered function variant.
func ResolveAggregateFunctionVariant(
    reg expr.ExtensionRegistry,
    urn string,
    name string,
    args ...types.FuncArg,
) (extensions.FunctionID, error)

// ResolveWindowFunctionVariant resolves a simple window function name and arguments
// to the ID of a registered function variant.
func ResolveWindowFunctionVariant(
    reg expr.ExtensionRegistry,
    urn string,
    name string,
    args ...types.FuncArg,
) (extensions.FunctionID, error)
```

Callers that want convenience resolution can opt into it first, then pass the resolved ID to `NewScalarFunc`, `NewAggregateFunc`, or `NewWindowFunc`.

```go
id, err := expr.ResolveScalarFunctionVariant(reg,
    extensions.SubstraitDefaultURNPrefix+"functions_arithmetic",
    "add",
    expr.NewPrimitiveLiteral(int32(1), false),
    expr.NewPrimitiveLiteral(int32(2), false),
)
if err != nil {
    return err
}

_, err = expr.NewScalarFunc(reg, id, nil,
    expr.NewPrimitiveLiteral(int32(1), false),
    expr.NewPrimitiveLiteral(int32(2), false),
)
```

This keeps exact ID construction separate from convenience resolution and avoids maintaining simple-name aliases in `Collection`. It also gives the resolver room to be smarter than the previous constructor fallback. For example, if a function had multiple variants and the best match for `i32, i32` was a generic variant like `add:any_any`, the old constructor fallback would still synthesize `add:i32_i32` and fail unless that exact variant existed.

---

## Summary by CodeRabbit

- **New Features**
  - Function identifiers now include complete signatures, improving support for overloaded scalar, aggregate, and window functions.
  - Added clearer function resolution for typed and generic arguments.
  - Function lookup now provides explicit handling for unregistered functions.

- **Bug Fixes**
  - Improved resolution of function variants across expressions, query planning, window functions, and lambda expressions.
  - Updated function references to consistently use signature-qualified names, reducing ambiguity between variants.

### DIFF
--- a/extensions/extension_mgr.go
+++ b/extensions/extension_mgr.go
@@ -130,7 +130,7 @@ var void = struct{}{}
 type variants interface {
 	*ScalarFunctionVariant | *AggregateFunctionVariant | *WindowFunctionVariant
 	Name() string
-	Signature() string
+	CompoundName() string
 }
 
 type extFn[T variants] interface {
@@ -139,7 +139,7 @@ type extFn[T variants] interface {
 
 func addToMaps[T variants](id FunctionID, fn extFn[T], m map[FunctionID]T) {
 	for _, v := range fn.GetVariants(id.URN) {
-		id.Signature = v.Signature()
+		id.Signature = v.CompoundName()
 		m[id] = v
 	}
 }
--- a/extensions/extension_mgr.go
+++ b/extensions/extension_mgr.go
@@ -130,7 +130,7 @@ var void = struct{}{}
 type variants interface {
 	*ScalarFunctionVariant | *AggregateFunctionVariant | *WindowFunctionVariant
 	Name() string
-	CompoundName() string
+	Signature() string
 }
 
 type extFn[T variants] interface {
@@ -139,7 +139,7 @@ type extFn[T variants] interface {
 
 func addToMaps[T variants](id FunctionID, fn extFn[T], m map[FunctionID]T) {
 	for _, v := range fn.GetVariants(id.URN) {
-		id.Signature = v.CompoundName()
+		id.Signature = v.Signature()
 		m[id] = v
 	}
 }
--- a/extensions/extension_mgr.go
+++ b/extensions/extension_mgr.go
@@ -94,15 +94,12 @@ func loadExtensionFile(collection *Collection, substraitFS embed.FS, ent fs.DirE
 // ID is the unique identifier for a substrait object
 type ID struct {
 	URN string
-	// Name of the object. For functions, a simple name may be used for lookups,
-	// but as a unique identifier the compound name will be used
+	// Name of the object.
 	Name string
 }
 
 type Collection struct {
 	urnSet map[string]struct{}
-
-	simpleNameMap map[ID]string
 
 	scalarMap        map[ID]*ScalarFunctionVariant
 	aggregateMap     map[ID]*AggregateFunctionVariant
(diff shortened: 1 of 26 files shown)


extensions/extension_mgr.go
```diff
@@ -284,13 +253,7 @@ func (c *Collection) Load(r io.Reader) error {
...
-	}
-
-	// add simple name aliases
-	for k, v := range simpleNames {
-		id.Name = k
-		c.simpleNameMap[id] = v
+		addToMaps[*WindowFunctionVariant](id, &wf, c.windowMap)
 	}
 
 	return nil
```
kadinrabo: Lookups are exact now but `GetFunctionRef` still takes whatever string you give it with no checking. So you can still build a plan with a bare name that won't read back. We might want to address that before opening it up

https://github.com/substrait-io/substrait-go/blob/259cc55e10227fbd84f76d7570f057c30243efc2/plan/builders.go#L230-L232

benbellick: excellent catch

kadinrabo: Sorry didn't mean to comment on that many lines

benbellick:
@coderabbitai review

PR #265, opened 2026-06-10
feat(literal): add NewEmptyMap and NewEmptyList constructors

kadinrabo:
follow-up to #236. adds NewEmptyMap and NewEmptyList to literal so all four collection constructors sit together, where callers already look. they wrap the existing expr.NewEmptyMapLiteral / NewEmptyListLiteral, which stay put (literal already wraps expr this way for the primitive constructors).

didn't move the expr versions: the dependency only runs literal -> expr, so relocating the bodies means literal reaching into expr's types. happy to remove them under a breaking change later if you'd rather they only live in one place.


kadinrabo:
folding this into #260 to keep #236 to a single PR. closing here.

PR #271, opened 2026-07-11
fix(plan): derive right join record types

bvolpato:
## Summary

Implement `RecordType()` semantics for right semi, right anti, and right single joins. This addresses the right-join portion of #210 without expanding into extension or write relations.

## Changes

- Return only right input schema for `RIGHT_SEMI` and `RIGHT_ANTI`.
- Return nullable left fields followed by unchanged right fields for `RIGHT_SINGLE`, matching right outer semantics.
- Enable schema validation for these now-supported join types.
- Add table-driven regression coverage.

## Repro

Calling `RecordType()` for these join types previously panicked. Right semi and anti now return right schema; right single returns combined schema with only left fields made nullable.

## Testing

- `go test ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6 run`

## Did this cause any problems?

No. Revert this commit to restore previous behavior.

### DIFF
--- a/plan/plan.go
+++ b/plan/plan.go
@@ -304,16 +304,11 @@ func validateRootNamesForSchema(recordType types.RecordType, names []string) err
 // have incomplete implementations that panic or guess.
 // TODO(#210): remove this once RecordType() is fixed for all relation types.
 func isRecordTypeSupported(rel Rel) bool {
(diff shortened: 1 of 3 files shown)


benbellick (approved):
💯

PR #273, opened 2026-07-13
fix: persist extension registry in SubqueryConverters for FromProto

Broshen:
When deserializing a plan from protobuf via FromProto, or when building a plan via PlanWithBindings, subquery expressions could not be resolved because the SubqueryConverter was never registered on the extension registry. This caused subquery expression conversion to silently fail or panic. This PR fixes this so that protobuf plans with Subquery Expressions can also be deserialized correctly.

### DIFF
--- a/plan/builders.go
+++ b/plan/builders.go
@@ -784,10 +784,12 @@ func (b *builder) PlanWithBindings(root Rel, rootNames []string, expectedTypeURL
 		relations[i].rel = o
 	}
 
+	reg := b.reg
+	reg.SetSubqueryConverter(&ExpressionConverter{ExtensionRegistry: reg})
 	return &Plan{
 		version:           &CurrentVersion,
 		extensions:        b.extSet,
-		reg:               b.reg,
+		reg:               reg,
 		expectedTypeURLs:  expectedTypeURLs,
 		relations:         relations,
 		parameterBindings: bindings,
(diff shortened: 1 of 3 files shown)


plan/builders.go
```diff
@@ -784,10 +784,12 @@ func (b *builder) PlanWithBindings(root Rel, rootNames []string, expectedTypeURL
 		relations[i].rel = o
 	}
 
+	reg := b.reg
+	reg.SetSubqueryConverter(&ExpressionConverter{ExtensionRegistry: reg})
```
benbellick, line 788: I tested locally and the currently added tests pass without this change. Would you mind adding a test that exercises this codepath?

plan/plan_test.go
```diff
@@ -132,6 +132,109 @@ func TestRejectsMismatchedRootNames(t *testing.T) {
 	assert.ErrorContains(t, err, "1 output name(s) but the output schema requires 2")
 }
 
+func TestFromProtoWithSubqueries(t *testing.T) {
```
benbellick, line 235: nit: I think this test might be a bit clearer if you did it table-driven style and extracted the proto builder to a named thing outside of the function. This is a minor style thing though, up to you if you agree :)

benbellick (commented):
Thanks! I think it looks good as long as we add a test proving that the `builders.go` change fixes a bug.

plan/builders.go
Broshen: `TestBuilderPlanRegistryWithSubqueries` should cover this now

benbellick (approved):
LGTM, thanks!

PR #276, opened 2026-07-20
fix: decode schemas in extension relations

Broshen:
RelFromProto always wrapped extension relation details (ExtensionSingleRel, ExtensionMultiRel, ExtensionLeafRel) in UndecodedExtension, which returns only the input schema. This caused panics when an emit mapping referenced columns added by the extension itself, and gave callers no way to teach the deserializer about their custom extension schemas.

This PR implements an ExtensionRelDecoder interface, that callers can provide to decode their extension relation details into a schema. It adds this interface to ExtensionRegistry, and RelFromProto uses this to decode extension relation types, falling back to UndecodedExtension when the decoder returns (nil, nil).

### DIFF
--- a/expr/expressions_test.go
+++ b/expr/expressions_test.go
@@ -729,4 +729,3 @@ type stubRelDecoder struct{}
 func (s *stubRelDecoder) DecodeExtensionRel(_ *anypb.Any) (any, error) {
 	return nil, errors.New("stub")
 }
-
--- a/expr/expressions_test.go
+++ b/expr/expressions_test.go
@@ -699,27 +699,26 @@ func TestSubqueryExpressionRoundtrip(t *testing.T) {
 }
 
 func TestExtensionRegistrySetAndGetDecoder(t *testing.T) {
+	const typeURL = "type.googleapis.com/test.Ext"
 	c := ext.GetDefaultCollectionWithNoError()
 
 	t.Run("nil before set", func(t *testing.T) {
 		reg := expr.NewEmptyExtensionRegistry(c)
-		require.Nil(t, reg.ExtensionRelDecoderFor())
+		require.Nil(t, reg.ExtensionRelDecoderFor(typeURL))
 	})
 
 	t.Run("returns decoder after set", func(t *testing.T) {
 		reg := expr.NewEmptyExtensionRegistry(c)
 		dec := &stubRelDecoder{}
-		reg.SetExtensionRelDecoder(dec)
-		require.Equal(t, dec, reg.ExtensionRelDecoderFor())
+		require.NoError(t, reg.SetExtensionRelDecoder(typeURL, dec))
+		require.Equal(t, dec, reg.ExtensionRelDecoderFor(typeURL))
 	})
 
-	t.Run("overwrite replaces decoder", func(t *testing.T) {
+	t.Run("duplicate typeURL returns error", func(t *testing.T) {
 		reg := expr.NewEmptyExtensionRegistry(c)
-		first := &stubRelDecoder{}
-		second := &stubRelDecoder{}
-		reg.SetExtensionRelDecoder(first)
-		reg.SetExtensionRelDecoder(second)
-		require.Equal(t, second, reg.ExtensionRelDecoderFor())
+		require.NoError(t, reg.SetExtensionRelDecoder(typeURL, &stubRelDecoder{}))
+		err := reg.SetExtensionRelDecoder(typeURL, &stubRelDecoder{})
+		require.ErrorContains(t, err, typeURL)
 	})
 }
 
--- a/expr/expressions_test.go
+++ b/expr/expressions_test.go
@@ -5,6 +5,7 @@ package expr_test
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"sort"
@@ -14,6 +15,7 @@ import (
 	"github.com/goccy/go-yaml"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	substraitgo "github.com/substrait-io/substrait-go/v8"
 	"github.com/substrait-io/substrait-go/v8/expr"
 	ext "github.com/substrait-io/substrait-go/v8/extensions"
 	"github.com/substrait-io/substrait-go/v8/plan"
@@ -22,6 +24,7 @@ import (
 	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
 	"google.golang.org/protobuf/encoding/protojson"
 	pb "google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 )
 
 const sampleYAML = `---
@@ -694,4 +697,70 @@ func TestSubqueryExpressionRoundtrip(t *testing.T) {
 			// comment in plan/subquery.go
 		})
 	}
+}
+
+func TestExtensionRegistrySetAndGetDecoder(t *testing.T) {
+	c := ext.GetDefaultCollectionWithNoError()
+
+	t.Run("nil before set", func(t *testing.T) {
+		reg := expr.NewEmptyExtensionRegistry(c)
+		require.Nil(t, reg.ExtensionRelDecoderFor())
+	})
+
+	t.Run("returns decoder after set", func(t *testing.T) {
+		reg := expr.NewEmptyExtensionRegistry(c)
+		dec := &stubRelDecoder{}
+		reg.SetExtensionRelDecoder(dec)
+		require.Equal(t, dec, reg.ExtensionRelDecoderFor())
+	})
+
+	t.Run("overwrite replaces decoder", func(t *testing.T) {
+		reg := expr.NewEmptyExtensionRegistry(c)
+		first := &stubRelDecoder{}
+		second := &stubRelDecoder{}
+		reg.SetExtensionRelDecoder(first)
+		reg.SetExtensionRelDecoder(second)
+		require.Equal(t, second, reg.ExtensionRelDecoderFor())
+	})
+}
+
+// stubRelDecoder is a minimal ExtensionRelDecoder that always returns an error.
+type stubRelDecoder struct{}
+
+func (s *stubRelDecoder) DecodeExtensionRel(_ *anypb.Any) (any, error) {
+	return nil, errors.New("stub")
+}
+
+func TestStructFieldRefGetTypeBounds(t *testing.T) {
+	st := &types.StructType{
+		Types: []types.Type{
+			&types.Int64Type{Nullability: types.NullabilityRequired},
+			&types.Int32Type{Nullability: types.NullabilityRequired},
+		},
+	}
+
+	t.Run("valid index returns type", func(t *testing.T) {
+		ref := &expr.StructFieldRef{Field: 0}
+		got, err := ref.GetType(st)
+		require.NoError(t, err)
+		assert.IsType(t, &types.Int64Type{}, got)
+	})
+
+	t.Run("negative index returns error", func(t *testing.T) {
+		ref := &expr.StructFieldRef{Field: -1}
+		_, err := ref.GetType(st)
+		require.ErrorIs(t, err, substraitgo.ErrInvalidType)
+	})
+
+	t.Run("index equal to len returns error", func(t *testing.T) {
+		ref := &expr.StructFieldRef{Field: int32(len(st.Types))}
+		_, err := ref.GetType(st)
+		require.ErrorIs(t, err, substraitgo.ErrInvalidType)
+	})
+
+	t.Run("index beyond len returns error", func(t *testing.T) {
+		ref := &expr.StructFieldRef{Field: int32(len(st.Types) + 1)}
+		_, err := ref.GetType(st)
+		require.ErrorIs(t, err, substraitgo.ErrInvalidType)
+	})
 }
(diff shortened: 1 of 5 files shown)


benbellick:
Hoping to review this soon, thanks! This PR made me realize that the current behavior is to silently accept unrecognized protobufs, which is a bit strange IMO. I have opened a ticket (https://github.com/substrait-io/substrait-go/issues/278) to work on this in a separate PR, but no need to address it here, just sharing for context. 

Also, there are some CI checks failing. Thanks!

Broshen:
updated!

expr/utils.go
```diff
@@ -7,8 +7,24 @@ import (
...
 )
 
+// ExtensionRelDecoder decodes an extension relation's Any detail into an
+// ExtensionRelDefinition that knows its output schema and expressions.
+// It is called by RelFromProto instead of falling back to UndecodedExtension
+// when a decoder is registered for the detail's type URL.
+//
+// The Rel type is intentionally left as an interface{} here to avoid an import
+// cycle between the expr and plan packages; the plan package casts it to plan.Rel
+// via the ExtensionRelDecoderFunc type alias defined there.
```
benbellick: I don't see `ExtensionRelDecoderFunc` mentioned anywhere else.

expr/field_reference.go
```diff
@@ -192,7 +192,7 @@ func (r *StructFieldRef) GetType(parentType types.Type) (types.Type, error) {
 		return nil, substraitgo.ErrInvalidType
 	}
 
-	if len(st.Types) < int(r.Field) {
+	if r.Field < 0 || int(r.Field) >= len(st.Types) {
```
benbellick: Is this a separate bugfix? If so, I think it would be easier to review as a separate PR to make the diff smaller and make the bug fixing commit easier to identify in git history.

plan/plan_test.go
```diff
@@ -120,15 +122,21 @@ func TestPlanRoundTripWithExtensions(t *testing.T) {
 
 func TestPlanRoundTripWithSubqueries(t *testing.T) {
```
benbellick, line 123: Why does this test need to change? Seems unrelated to the stated goal of the PR.

expr/expressions_test.go
```diff
@@ -695,3 +698,69 @@ func TestSubqueryExpressionRoundtrip(t *testing.T) {
...
+		_, err := ref.GetType(st)
+		require.ErrorIs(t, err, substraitgo.ErrInvalidType)
+	})
+
+	t.Run("index beyond len returns error", func(t *testing.T) {
+		ref := &expr.StructFieldRef{Field: int32(len(st.Types) + 1)}
+		_, err := ref.GetType(st)
+		require.ErrorIs(t, err, substraitgo.ErrInvalidType)
+	})
+}
```
benbellick: I think these may be changes for the other bug fix you mentioned? I would suggest these also become part of a different PR for ease of reviewing. Thanks!

plan/plan_test.go
benbellick, line 1: I haven't gotten a chance to go through this file entirely yet but i think there is opportunity to have a fewer tests that are simpler and clearer. I want to flush the comments that I have so far now, but hopefully I'll have a chance to give more substantive feedback on them later today.

benbellick (changes requested):
Thanks! I didn't fully finish the review but hopefully will later today. Left a few comments, please let me know if anything is unclear :)

expr/utils.go
Broshen: stagnant comment, updated

expr/field_reference.go
Broshen: moved to https://github.com/Broshen/substrait-go/pull/3, on top of this branch

plan/plan_test.go
Broshen: i was trying to test round tripping with multiple columns & remappings instead of just emit:direct & i touched this test by accident. undid the changes

expr/expressions_test.go
Broshen: moved to https://github.com/Broshen/substrait-go/pull/3

plan/plan_test.go
Broshen: restructured the test a bit to be shorter & clearer

plan/plan.go
benbellick, line 1: I think that `FromProto` would still fail to handle extensions since it makes its own extension registry. We should add a way to provide the extension decoder info. This would serve as a good full integration test as well (and maybe we find some of the others are redundant with that). 
https://github.com/substrait-io/substrait-go/blob/4a512e6f8750ec91fe5d8e8afd4d67cf9f58a8f2/plan/plan.go#L228-L260


I'm wondering if an even better solution is to make the decoder part of the `extensions.Collection` struct, which is already passed into `FromProto`, but I think that would be a larger refactor and may not be worth it for this PR.

expr/utils.go
```diff
@@ -18,6 +35,10 @@ type ExtensionRegistry struct {
 	// subqueryConverter is injected by the plan package to handle subquery expressions
 	// TODO: We may want to consider refactoring to make a cleaner interface here
 	subqueryConverter
+
+	// extensionRelDecoder is optionally set to decode extension relation details
+	// into typed ExtensionRelDefinitions during RelFromProto.
+	extensionRelDecoder ExtensionRelDecoder
```
benbellick:
```suggestion
	extensionRelDecoders map[String]ExtensionRelDecoder
```
I'm not sure, but I'm wondering if it makes more sense to store a single decoder per type URL? That may simplify the code in the actual extension decoders and we can have the plan decoder dynamically dispatch each individual decoder based off of type url. This makes it a bit clearer for `substrait-go` to handle some extensions but not others. What do you think?

benbellick (commented):
🤦 sorry I had two more comments I meant to leave before, but they got lost between switching GH accounts. Thanks!

expr/utils.go
```diff
@@ -3,12 +3,29 @@
...
+// It is called by RelFromProto for extension rels whose type URL has a
+// registered decoder; unregistered type URLs fall back to UndecodedExtension.
+//
+// The return type is intentionally left as any to avoid an import cycle between
+// the expr and plan packages; callers cast it to the appropriate concrete type.
+type ExtensionRelDecoder interface {
+	// DecodeExtensionRel decodes detail into a plan.ExtensionRelDefinition.
+	// The return type is any because ExtensionRelDefinition is defined in the plan
+	// package, and importing it here would create a cycle with the expr package.
+	DecodeExtensionRel(detail *anypb.Any) (any, error)
```
benbellick, line 26: Kind of gross, but I think to get things working here, we should actually do:
```suggestion
	DecodeExtensionRel(detail *anypb.Any, reg expr.ExtensionRegistry,) (any, error)
```

The reason is that we need access to the ExtensionRegistry inside of `ExtensionRel`'s if they at all use extension stuff, e.g. a scalar function inside of an expression in that proto. 

That being said, I think its okay to leave this out for now if it isn't necessary for you. The "grossness" there is that `ExtensionRelDecoder` is itself attached to `ExtensionRegistry`, so there is some weird circular stuff which is confusing.

benbellick: Here is an example of this kind of problem:

```json
   {
     "extensionUrns": [
       {
         "extensionUrnAnchor": 1,
         "urn": "extension:io.substrait:functions_arithmetic"
       }
     ],
     "extensions": [
       {
         "extensionFunction": {
           "extensionUrnReference": 1,
           "functionAnchor": 42,
           "name": "add:i64_i64"
         }
       }
     ],
     "relations": [
       {
         "root": {
           "input": {
             "extensionSingle": {
               "input": "...",
               "detail": {
                 "@type": "type.googleapis.com/example.ProjectSet",
                 "expressions": [
                   {
                     "scalarFunction": {
                       "functionReference": 42, // WHAT IS THIS? need the extension registry
                       "arguments": ["..."]
                     }
                   }
                 ]
               }
             }
           }
         }
       }
     ]
   }
```

plan/plan.go
benbellick: I also am unsure if an `ExtensionRegistry` is the right place to put the decoders, since `ExtensionRegistry` is more like the "stateful thing which carries mappings between anchors and types / functions". The better place for it would either be on the `extensions.Collection` (which is like shared extension info), or even better, on a dedicated decoder struct, that we then use like `myCustomDecoder.FromProto(...)`. 

However, I think what you have here for now is fine... I know @kadinrabo is doing some work to separate out the decoders in this library. When that happens, it will probably include refactoring this stuff too, so we can take care of that then. 


Sorry, this is a thought dump just to say that we should leave this as is, but change it later :)

plan/plan_test.go
```diff
@@ -419,3 +421,359 @@ func TestFromProtoRightSemiJoinRootNames(t *testing.T) {
...
+
+	out, err := RelFromProto(rel, reg)
+	require.NoError(t, err)
+
+	// psStyleDecoder resolves types from the SRF expressions: i64 passthrough + i32 SRF output.
+	got := out.RecordType()
+	require.Equal(t, int32(2), got.FieldCount())
+	assert.IsType(t, &types.Int64Type{}, got.GetFieldRef(0))
+	assert.IsType(t, &types.Int32Type{}, got.GetFieldRef(1))
+}
```
benbellick: IMO the below tests sufficiently test the new code. I don't see the value of including "ProjectSetStyle" decoders without an explanation of what that is. Furthermore, I don't quite see what it is about that decoder that is ProjectSetStyle. Please let me know if I am missing something? 
```suggestion
```

benbellick: And it means we could shorten these tests quite a bit.

plan/plan_test.go
```diff
@@ -419,3 +421,359 @@ func TestFromProtoRightSemiJoinRootNames(t *testing.T) {
...
+	makePlan := func(extRel *proto.Rel, names []string) *proto.Plan {
+		return &proto.Plan{Relations: []*proto.PlanRel{{
+			RelType: &proto.PlanRel_Root{Root: &proto.RelRoot{Names: names, Input: extRel}},
+		}}}
+	}
+
+	cases := []struct {
+		name                 string
+		plan                 *proto.Plan
+		panicsWithoutDecoder bool
```
benbellick: nit: I think this test is a bit hard to read, and it is really because of this `panicsWithoutDecoder` (because we are testing multiple things at once). But we can fix this in a later PR, it doesn't have to be here.

benbellick: Upon further thought, I think this is a bit confusing now seeing the quite similar `TestExtensionRelDecoder` below. Can we instead have simply:
- a full-plan success test for the thre relation variants
- a single relation-level error test? 

I think this would be a bit easier to understand for readers and would result in less setup and branching. Feel free to do things a different way if you prefer.

expr/utils.go
Broshen: i think at this point it'd be up to the `ExtensionRelDecoder` implementation to have a registry internally and handle references. i'd rather just leave this out and get this merged.

plan/plan_test.go
Broshen: i was trying to illustrate a practical example usecase of an extension registry - representing unnset/project set relations as a extensionsinglerel, and using a decoder to correctly marshal/unmarshal it from protobuf, but ok, removed

plan/plan_test.go
Broshen: TestFromProtoWithDecoder now just has 1 success test per variant + a seperate error test

bvolpato (commented):
[P2] Validate decoded extension roots — `plan/plan.go:763`

Non-blocking: this new decoder path can produce a known schema, but `isRecordTypeSupported` still returns false for every extension relation, so `Relation.FromProto` skips `validateRootNamesForSchema`. A leaf extension decoded to three fields with a root containing one name is accepted by `FromProtoWithDecoder`, whereas an equivalent supported relation is rejected. Distinguish decoded definitions from `UndecodedExtension` here so malformed root names fail validation while genuinely unknown extensions remain exempt.

### What This PR Does

Adds per-type-URL extension-relation decoders, wires them through `FromProtoWithDecoder`, and applies decoded definitions to single-, multi-, and leaf-extension relations. Decoded schemas are available for emit mapping and round-tripping; root-name validation still does not consume them.

### Validation

Reviewed current head `bb07fa7` against `main` `c5dd519`, linked issue #278, all review threads, changed files, call sites, and tests. `go test -race ./expr ./plan` passed. Standalone reproduction confirmed a decoded three-field extension root with one name is accepted. All seven PR checks pass; Codecov reports 94.59% patch coverage with two changed lines uncovered.

### Recommendation

COMMENT

expr/utils.go
benbellick: Yup, totally reasonable 🙂 

I'll open an issue for it.

plan/plan_test.go
```diff
@@ -419,3 +421,269 @@ func TestFromProtoRightSemiJoinRootNames(t *testing.T) {
...
+	assert.True(t, isRecordTypeSupported(&ExtensionLeafRel{definition: decoded}))
+	assert.True(t, isRecordTypeSupported(&ExtensionMultiRel{definition: decoded}))
+
+	assert.False(t, isRecordTypeSupported(&ExtensionSingleRel{definition: undecoded}))
+	assert.False(t, isRecordTypeSupported(&ExtensionLeafRel{definition: undecoded}))
+	assert.False(t, isRecordTypeSupported(&ExtensionMultiRel{definition: undecoded}))
+}
+
+// customExtDef is a test ExtensionRelDefinition that claims a fixed output schema.
+type customExtDef struct {
```
benbellick, line 524: This is a great test example IMO 🚀

benbellick (approved):
nice!

expr/utils.go
benbellick: https://github.com/substrait-io/substrait-go/issues/285

Issue #280, opened 2026-08-04
Migrate substrait-protobuf out of core


kadinrabo:
## Problem

substrait-go provides Go types for describing query plans and serializes those plans to protobuf.

Some consumers (e.g. iceberg-go) only need the former to build plans and inspect types. They don't need substrait-protobuf in their dependency tree, but they're [transitively forced to have it](https://github.com/apache/iceberg-go/blob/211799b2864689719f8423e511785ea0c8db5b12/go.mod#L232) because:

1. the generated protobuf types are part of substrait-go's public API
2. `go.mod` requires substrait-protobuf
3. minimum version selection requires it for all downstream consumers

This turns a simple consumer bump of substrait-go into a larger coordinated change across substrait-go, iceberg-go and anything else in the chain.

The coupling cannot be trivially cleaned up because it's woven into the type declarations:

```go
type Nullability = proto.Type_Nullability
```
from [types/types.go:21](https://github.com/substrait-io/substrait-go/blob/main/types/types.go#L21).

That's a direct alias to the protobuf enum, and there are many more of these cases. https://github.com/substrait-io/substrait-go/issues/161 is a similar case where function and type anchors (a serialization concern) live in the domain model.

## Proposed Solution

Create a new module `substrait-go/codec` for substrait-protobuf code and move it out of core. Consumers who want the wire format import `substrait-go/codec` on purpose.

Steps:
- [ ] 1. create a new module `substrait-go/codec`, with its own `go.mod`, as the one place protobuf is allowed to live
- [ ] 2. give substrait-go its own versions of the public types that are currently protobuf types, and split any test that mixes domain and wire assertions
- [ ] 3. copy the proto conversions into `substrait-go/codec`, keeping the originals, with a [golden test file](https://en.wikipedia.org/wiki/Characterization_test) capturing the exact bytes the original produces, so the check outlives the original
- [ ] 4. delete the originals and drop the protobuf methods from the interfaces working top-down so each deletion's callers are already gone (major release)

Steps 2 and 3 repeat per slice of substrait functionality with lower-level slices first.

## Done when

A consumer only importing the domain packages runs this check:

```
go mod graph | grep substrait-protobuf
```

and it comes back empty.

kadinrabo:
Mapped out the rest of this below. The codec module isn't blocking any of the type work and the two can go in parallel. Only step 3 needs both.

```mermaid
flowchart TB
  subgraph S2["Step 2: replace the aliases"]
    direction TB
[...]

PR #282, opened 2026-08-06
feat(type)!: define Nullability instead of aliasing the protobuf enum

kadinrabo:
`types.Nullability` was `= proto.Type_Nullability`, so the generated enum was substrait-go's public API. It becomes substrait-go's own `int32` with the same constant names and the same wire numbers, so most consumers still compile.

`go mod graph` doesn't change here. The core still imports substrait-protobuf.

BREAKING CHANGE: `types.Nullability` is no longer an alias for `proto.Type_Nullability`. Code that passes one where the other is expected needs a cast.

Part of #280.

### DIFF
--- a/extensions/simple_extension_test.go
+++ b/extensions/simple_extension_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/substrait-io/substrait-go/v8/extensions"
 	"github.com/substrait-io/substrait-go/v8/types"
-	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
 )
 
 func TestUnmarshalSimpleExtension(t *testing.T) {
@@ -76,7 +75,7 @@ scalar_functions:
 	typ, err := arg1.Value.ValueType.ReturnType(nil, nil)
 	assert.NoError(t, err)
 	assert.IsType(t, &types.UserDefinedType{}, typ)
-	assert.Equal(t, proto.Type_NULLABILITY_REQUIRED, typ.GetNullability(), "expected NULLABILITY_REQUIRED")
+	assert.Equal(t, types.NullabilityRequired, typ.GetNullability(), "expected NULLABILITY_REQUIRED")
 
 	assert.Equal(t, "scalar2", f.ScalarFunctions[1].Name)
 	assert.IsType(t, extensions.ValueArg{}, f.ScalarFunctions[1].Impls[0].Args[0])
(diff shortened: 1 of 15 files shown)


extensions/simple_extension_test.go
```diff
@@ -11,7 +11,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/substrait-io/substrait-go/v8/extensions"
 	"github.com/substrait-io/substrait-go/v8/types"
-	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
```
benbellick, line 14: 💯

types/nullability_test.go
```diff
@@ -0,0 +1,29 @@
...
+func TestNullabilityString(t *testing.T) {
+	for _, td := range []struct {
+		n        types.Nullability
+		expected string
+	}{
+		{types.NullabilityUnspecified, "NULLABILITY_UNSPECIFIED"},
+		{types.NullabilityNullable, "NULLABILITY_NULLABLE"},
+		{types.NullabilityRequired, "NULLABILITY_REQUIRED"},
+		// unmodelled value
+		{types.Nullability(7), "7"},
```
benbellick:
```suggestion
```
nit: I don't think it matters that much to test this "bad" path.

types/types.go
```diff
@@ -18,14 +18,30 @@ import (
 
 type Version = proto.Version
 
-type Nullability = proto.Type_Nullability
+// Nullability mirrors the Type.Nullability enum in the Substrait spec. substrait-go declares it
+// rather than aliasing the generated enum so that serialization can move out of the core.
```
benbellick:
```suggestion
// Nullability indicates whether values of a Substrait type may be null.
```
No need to explain why. When the migration is done, we may forget to update this comment.

types/types.go
```diff
@@ -18,14 +18,30 @@ import (
...
 const (
-	NullabilityUnspecified = proto.Type_NULLABILITY_UNSPECIFIED
-	NullabilityNullable    = proto.Type_NULLABILITY_NULLABLE
-	NullabilityRequired    = proto.Type_NULLABILITY_REQUIRED
+	NullabilityUnspecified Nullability = 0
+	NullabilityNullable    Nullability = 1
+	NullabilityRequired    Nullability = 2
 )
 
+// String reports the spec name. Unknown values print as the number.
```
benbellick:
```suggestion
// String returns the protobuf enum name for the nullability value.
```

benbellick (approved):

PR #283, opened 2026-08-06
feat(codec): add a new codec module and wire it into ci

kadinrabo:
The protobuf conversions need somewhere to live that isn't the core's module, so this adds `substrait-go/codec` with its own `go.mod`. It is currently empty, but will later be iteratively populated with all of the protobuf-related code. 

GH Actions support is added, and coverage reports for both modules are uploaded together. 

Part of #280.


benbellick (approved):
perfect!

PR #288, opened 2026-08-12
docs: add CONTRIBUTING.md

nielspardon:
This repository had no contributor documentation at all, and the README carries no development or contributing guidance.

That matters most for the CLA. Substrait requires every contributor to sign the [Contributor License Agreement](https://cla-assistant.io/substrait-io/substrait), and CLA Assistant enforces this as a `license/cla` status check on pull requests here, but nothing in the repository mentioned it — so a contributor's first contact with the requirement is the check appearing on their PR, which is the problem described in substrait-io/substrait#630.

The new file documents:

- **The CLA**, using the wording carried by the other repositories in the organization.
- **The specification as source of truth**, and where to take a question when the spec is unclear rather than settling it in this tree.
- **Commit conventions** — conventional commits with an optional package scope and `!` for breaking changes. Worth being explicit here because `go-semantic-release` derives the version and release notes from these messages, so the type and the `!` marker directly determine what gets released.
- **Building, testing and linting**, plus the SPDX header convention on Go files.

Two notes on accuracy, since I checked rather than assumed:

- The SPDX header is described as a convention rather than a rule, because nothing in CI enforces it (58 of 99 Go files carry one, though 11 of the 12 most recently added files do, so it is clearly current practice).
- The linting section points at `golangci-lint` 2.x directly and does **not** recommend the configured `pre-commit` hook, because that hook pins golangci-lint `v1.52.2` while `.golangci.yml` is a `version: "2"` configuration that 1.x cannot parse. That mismatch looks like a genuine bug, but fixing it is out of scope for a docs change — happy to send a separate PR bumping the hook if you'd like.

Companion to substrait-io/substrait#1170, which fixes the specification repository's website and `CONTRIBUTING.md`.

See substrait-io/substrait#630

🤖 Generated with AI

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,75 @@
+# Contributing to Substrait Go
+
+Welcome! This page provides some orientation and recommendations on how to get
+the best results when engaging with the community.
+
+## Contributor License Agreement
+
+Substrait requires all contributors to sign the
+[Contributor License Agreement (CLA)](https://cla-assistant.io/substrait-io/substrait)
+before their contributions can be merged. A GitHub app checks this on every pull
+request and guides new contributors through signing it.
+
+## The specification is the source of truth
+
+Substrait Go is an implementation of the
+[Substrait specification](https://substrait.io/); it does not define Substrait
+semantics. Review behavioral changes against the spec text and the `.proto`
+comments in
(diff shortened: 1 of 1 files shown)


benbellick (approved):
nice! 🚀

PR #290, opened 2026-08-12
fix(type): keep the type variation reference in WithNullability

### LINKED ISSUE #289: WithNullability drops the type variation reference on four types
Four of the fifteen `WithNullability` implementations lose the type variation reference, so it silently becomes 0.

```go
c := types.NewIntervalCompoundType().WithPrecision(types.PrecisionSeconds).WithTypeVariationRef(7)
fmt.Println(c.WithNullability(types.NullabilityRequired).GetTypeVariationReference())
// want 7, got 0
```

[`IntervalCompoundType`](https://github.com/substrait-io/substrait-go/blob/985d911da1f72c4e66ea0cd2e1470105f897bc83/types/interval_compound_type.go#L10-L36) names two of its three fields in the returned literal, and [`PrecisionTimestampType`](https://github.com/substrait-io/substrait-go/blob/985d911da1f72c4e66ea0cd2e1470105f897bc83/types/precison_timestamp_types.go#L70-L103) and [`PrecisionTimeType`](https://github.com/substrait-io/substrait-go/blob/985d911da1f72c4e66ea0cd2e1470105f897bc83/types/precison_timestamp_types.go#L227-L258) do the same. [`PrecisionTimestampTzType`](https://github.com/substrait-io/substrait-go/blob/985d911da1f72c4e66ea0cd2e1470105f897bc83/types/precison_timestamp_types.go#L191-L195) calls `PrecisionTimestampType`'s version, so it loses the reference too.

Happy to send the fix, it's three field copies plus a table test that fails all four types.

kadinrabo:
Closes #289.

### DIFF
--- a/types/interval_compound_type.go
+++ b/types/interval_compound_type.go
@@ -30,8 +30,9 @@ func (m IntervalCompoundType) GetPrecisionProtoVal() int32 {
 func (IntervalCompoundType) isRootRef() {}
 func (m IntervalCompoundType) WithNullability(n Nullability) Type {
 	return IntervalCompoundType{
-		precision:   m.precision,
-		nullability: n,
+		precision:        m.precision,
+		typeVariationRef: m.typeVariationRef,
+		nullability:      n,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


bvolpato (approved):
LGTM

PR #291, opened 2026-08-12
fix(types): normalize nullable any return bindings

dd-david-levin:
## Summary

Normalize a nullable `anyN?` occurrence to the required base binding when deriving polymorphic return types, then let the return occurrence reapply its declared nullability.

This implements the return-substitution side of the semantics documented in substrait-io/substrait#960. In particular:

```text
list<string?> against list<any1?> binds any1 = string
list<i64?> against list<any2?> binds any2 = i64
map<any1, any2?> resolves to map<string, i64?>
```

Without this change, the map key incorrectly remains nullable and resolves as `map<string?, i64?>`.

## Details

- Normalize nullable `anyN?` source occurrences in `unwrapAnyTypeWithName`.
- Preserve the return occurrence's declaration, so `anyN` remains required and `anyN?` is nullable.
- Add direct required/nullable substitution tests and a composite map regression test.
- Make `IntervalDayType.WithNullability` honor the `Type` interface's copy contract. This prevents return derivation from mutating an input argument while normalizing its binding.

This is complementary to draft #244. That PR addresses broader argument matching and constrained-binding consistency; this PR is narrowly about extracting the correct binding for return-type substitution.

## Testing

```text
go test ./...
git diff --check
```

Both pass.

---

This PR was developed with AI assistance. All changes have been reviewed, and I take full responsibility for this contribution.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of nullable generic types, ensuring required and nullable values are resolved correctly in nested type expressions.
  - Prevented type resolution from unintentionally changing the nullability of input arguments.
  - Updated interval day type nullability changes to preserve the original type while returning an independent copy.

- **Tests**
  - Added coverage for nullable generic type resolution and non-mutating type behavior.

### DIFF
--- a/types/any_type.go
+++ b/types/any_type.go
@@ -51,13 +51,22 @@ func (m *AnyType) ShortString() string {
 }
 
 // unwrapAnyTypeWithName searches for AnyType in p with the specified name,
-// and if found, returns argType.  If p is a composite type,
-// recursively unwraps p to search for AnyType in p's parameters.
-// Returns nil Type if AnyType was not found.
+// and if found, returns the concrete type bound by that occurrence. A nullable
+// occurrence binds the required base type; ReturnType reapplies the return
+// occurrence's declared nullability. If p is a composite type, the search
+// recurses through its parameters. Returns nil Type if AnyType was not found.
 func unwrapAnyTypeWithName(name string, p FuncDefArgType, argType Type) (Type, error) {
 	switch arg := p.(type) {
 	case *AnyType:
 		if arg.Name == name {
+			// A nullable any occurrence represents a nullable use of the required
+			// base type. Strip the occurrence's nullability from the binding here;
+			// ReturnType reapplies the return occurrence's declared nullability.
+			// This lets a signature relate any1 and any1? across structures, for
+			// example j(any1, list<any1?>) -> any1.
+			if arg.Nullability == NullabilityNullable {
+				return argType.WithNullability(NullabilityRequired), nil
+			}
 			return argType, nil
 		}
 	case *ParameterizedListType:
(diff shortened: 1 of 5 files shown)


benbellick:
@coderabbit review

types/any_type.go
```diff
@@ -51,13 +51,22 @@ func (m *AnyType) ShortString() string {
...
 func unwrapAnyTypeWithName(name string, p FuncDefArgType, argType Type) (Type, error) {
 	switch arg := p.(type) {
 	case *AnyType:
 		if arg.Name == name {
+			// A nullable any occurrence represents a nullable use of the required
+			// base type. Strip the occurrence's nullability from the binding here;
+			// ReturnType reapplies the return occurrence's declared nullability.
+			// This lets a signature relate any1 and any1? across structures, for
+			// example j(any1, list<any1?>) -> any1.
+			if arg.Nullability == NullabilityNullable {
```
kadinrabo: Traced this through `ReturnType` and the `MatchWithNullability` gate and the normalization looks right: a nullable occurrence only ever binds a nullable arg, so stripping to the required base here is what lets a required return occurrence resolve (the map-key case). Nice fix, and the `WithNullability` copy change is the right call, `IntervalDayType` was the only impl still mutating its receiver.

One thing for the next reader: this strip is deliberately scoped to nullable occurrences, leaving the required-return-vs-nullable-arg case to the [#943 TODO](https://github.com/substrait-io/substrait/issues/943) in `ReturnType`. The two branches read as asymmetric at a glance, so a one-line note here might stop someone from collapsing them into one and quietly stepping on #943.

PR #293, opened 2026-08-19
refactor: invert control flow of *Remap builder functions

vbarua:
The *Remap builder functions are deprecated. However, they are currently used to implement the non-*Remap functions, which makes them harder to remove.

This PR inverts the control flow so that the *Remap functions are implemented using non-deprecated functions, which will make future removal easier.

All existing tests for the builder functions pass with their new implementation. This PR is purely a mechanical change.

### DIFF
--- a/plan/builders.go
+++ b/plan/builders.go
@@ -335,20 +335,13 @@ func (b *builder) AggregateExprs(input Rel, measures []AggRelMeasure, groups ...
 	return arb.Build()
 }
 
-func (b *builder) CreateTableAsSelectRemap(input Rel, remap []int32, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error) {
+func (b *builder) CreateTableAsSelect(input Rel, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error) {
 	if input == nil {
 		return nil, errNilInputRel
 	}
 
-	nOutput := input.RecordType().FieldCount()
-	for _, idx := range remap {
-		if idx < 0 || idx >= nOutput {
-			return nil, errOutputMappingOutOfRange
-		}
-	}
-
 	return &NamedTableWriteRel{
-		RelCommon:   RelCommon{mapping: remap},
+		RelCommon:   RelCommon{mapping: nil},
 		names:       tableName,
 		tableSchema: schema,
 		op:          WriteOpCTAS,
@@ -357,8 +350,16 @@ func (b *builder) CreateTableAsSelectRemap(input Rel, remap []int32, tableName [
 	}, nil
 }
 
-func (b *builder) CreateTableAsSelect(input Rel, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error) {
-	return b.CreateTableAsSelectRemap(input, nil, tableName, schema)
+func (b *builder) CreateTableAsSelectRemap(input Rel, remap []int32, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error) {
+	ctas, err := b.CreateTableAsSelect(input, tableName, schema)
+	if err != nil {
+		return nil, err
+	}
+	rel, err := ctas.Remap(remap...)
+	if err != nil {
+		return nil, err
+	}
+	return rel.(*NamedTableWriteRel), nil
 }
 
 func (b *builder) CrossRemap(left, right Rel, remap []int32) (*CrossRel, error) {
--- a/plan/builders.go
+++ b/plan/builders.go
@@ -101,11 +101,10 @@ type Builder interface {
 	Filter(input Rel, condition expr.Expression) (*FilterRel, error)
 	// Deprecated: Use JoinAndFilter(...).Remap() instead.
 	JoinAndFilterRemap(left, right Rel, condition, postJoinFilter expr.Expression, joinType JoinType, remap []int32) (*JoinRel, error)
-	// Deprecated: Use Fetch(...).Remap() instead.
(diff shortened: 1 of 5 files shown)


plan/builders.go
```diff
@@ -272,32 +272,32 @@ func (b *builder) AggregateFn(nameSpace, key string, opts []*types.FunctionOptio
 }
 
 func (b *builder) Project(input Rel, exprs ...expr.Expression) (*ProjectRel, error) {
-	return b.ProjectRemap(input, nil, exprs...)
-}
-
-func (b *builder) ProjectRemap(input Rel, remap []int32, exprs ...expr.Expression) (*ProjectRel, error) {
```
vbarua, line 278: For example, here Project was implemented using ProjectRemap.

What I've done here is renamed ProjectRemap to Project, and deleted the remapping code in the function to minimize diff noise with this change.

plan/builders.go
```diff
@@ -272,32 +272,32 @@ func (b *builder) AggregateFn(nameSpace, key string, opts []*types.FunctionOptio
...
+	project, err := b.Project(input, exprs...)
+	if err != nil {
+		return nil, err
+	}
+	rel, err := project.Remap(remap...)
+	if err != nil {
+		return nil, err
+	}
+	return rel.(*ProjectRel), err
+}
```
vbarua, line 298: Here, we implement ProjectRemap in terms of Project() and Remap(), which is what the deprecation message for ProjectRemap recommends users do.

This pattern is what I've followed for all the other relations.

plan/builders.go
```diff
@@ -272,32 +272,32 @@ func (b *builder) AggregateFn(nameSpace, key string, opts []*types.FunctionOptio
...
-	noutput := input.RecordType().FieldCount() + int32(len(exprs))
-	for _, idx := range remap {
-		if idx < 0 || idx >= noutput {
-			return nil, errOutputMappingOutOfRange
-		}
-	}
-
 	return &ProjectRel{
-		RelCommon: RelCommon{mapping: remap},
+		RelCommon: RelCommon{mapping: nil},
```
vbarua, line 282: Explicitly set a RelCommon with `mapping: nil`, which is how we signal the `direct` output mode in the library.

vbarua:
@kadinrabo can you take a look at this when you have a chance.

plan/builders.go
kadinrabo: I think there's a mismatch in the meaning of `nil` because the RemapHelper treats it like "emit no columns" but the builders want it to mean direct

https://github.com/substrait-io/substrait-go/blob/b922a54264fa6a54ca8c5d50022ff30ea4b06858/plan/relations.go#L177-L180

vbarua: Digging into this a bit.

On the RelCommon struct itself
https://github.com/substrait-io/substrait-go/blob/b922a54264fa6a54ca8c5d50022ff30ea4b06858/plan/common.go#L128-L132
there's only a mapping field. When looking at RelCommon, a `nil` mapping means that no mapping is set.

The original Project builder made use of this
```go
func (b *builder) Project(input Rel, exprs ...expr.Expression) (*ProjectRel, error) {
	return b.ProjectRemap(input, nil, exprs...)
}
```
passing in `nil` as the value of the `remap` arg to `ProjectRemap`. Setting `nil` explicitly here matches that behaviour.

However, as you point out the updated version of ProjectRemap can be passed a `nil` for remap, and the `Remap` function were using now treats that as being equivalent to mapping with no fields.

Cursedly, the Remap function itself is defined as
https://github.com/substrait-io/substrait-go/blob/b922a54264fa6a54ca8c5d50022ff30ea4b06858/plan/plan.go#L393-L401

If you call `.Remap()` with no args, `mapping` is `nil`, meaning that inside of the code we can't tell whether you want direct output mode, or an empty mapping. If we take `nil` to mean direct output mode, there's no way to construct a relation with an empty mapping.

What we _could_ do, is add a a guard to all the *Remap functions that short-circuits if `remap` is `nil` and avoids calling the `.Remap` helper in that case, preserving the current behaviour. That said, we are also about to delete all of the *Remap helpers in an immediate follow-up

vbarua: Something I will do actually is update the Remap method docs to make this clearer actually.

kadinrabo: That's fair, just calling it out

kadinrabo (approved):
Looks good to me given we'll remove the remap helpers in https://github.com/substrait-io/substrait-go/pull/294/

plan/builders.go
vbarua: > That's fair, just calling it out

To be clear, it's a good call out. It was good to double check this and clarify it in the docs.

benbellick: We should not rely on `nil` having semantic meaning IMO. But that is for a separate PR 😅 The documentation is sufficient for now!

plan/builders.go
```diff
@@ -101,11 +101,10 @@ type Builder interface {
 	Filter(input Rel, condition expr.Expression) (*FilterRel, error)
 	// Deprecated: Use JoinAndFilter(...).Remap() instead.
 	JoinAndFilterRemap(left, right Rel, condition, postJoinFilter expr.Expression, joinType JoinType, remap []int32) (*JoinRel, error)
-	// Deprecated: Use Fetch(...).Remap() instead.
-	JoinAndFilter(left, right Rel, condition, postJoinFilter expr.Expression, joinType JoinType) (*JoinRel, error)
```
vbarua, line 105: I wrote a test for this and decided to un-deprecate it (the comment is obviously on the wrong method). I moved it to after Join for better grouping (IMO).

benbellick (approved):

PR #295, opened 2026-08-20
feat(plan): support expression encoding for FetchRel offset and count

vbarua:
FetchRel now stores offset and count as *expr.Expression rather than int64,
enabling both the literal and expression variants of:
* count_mode: count/count_expr 
* offset_mode: offset/offset_expr

to be processed.

When reading from a FetchRel from a proto, the library promotes count and offset
to expression equivalents. When outputting protobufs, the library outputs
count_expr and offset_expr when they are set.

## Summary by CodeRabbit

* **New Features**
  * Fetch operations now support optional expression-based offset and count values.
  * Omitting the offset starts from the beginning; omitting the count returns all records.
  * Fetch plans can now preserve and process non-literal offset and count expressions.

* **Bug Fixes**
  * Improved compatibility when reading legacy fetch plans, including offset-only and count-only configurations.

### DIFF
--- a/plan/builders.go
+++ b/plan/builders.go
@@ -74,10 +74,11 @@ type Builder interface {
 	AggregateExprs(input Rel, measures []AggRelMeasure, groups ...[]expr.Expression) (*AggregateRel, error)
 	CreateTableAsSelect(input Rel, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error)
 	Cross(left, right Rel) (*CrossRel, error)
-	// Fetch constructs a fetch relation providing an offset (skipping some number of
-	// rows) and a count (restricting output to a maximum number of rows).  If count
-	// is FETCH_COUNT_ALL_RECORDS (-1) all records will be returned.
-	Fetch(input Rel, offset, count int64) (*FetchRel, error)
+	// Fetch constructs a Fetch relation that skips offset rows and returns at most count rows from the input.
+	//
+	// offset must evaluate to a non-negative integer or Substrait null. Pass a Go nil for no offset.
+	// count must evaluate to a non-negative integer or Substrait null. Pass a Go nil to return all records.
+	Fetch(input Rel, offset, count *expr.Expression) (*FetchRel, error)
 	Filter(input Rel, condition expr.Expression) (*FilterRel, error)
 	Join(left, right Rel, condition expr.Expression, joinType JoinType) (*JoinRel, error)
 	JoinAndFilter(left, right Rel, condition, postJoinFilter expr.Expression, joinType JoinType) (*JoinRel, error)
@@ -312,7 +313,7 @@ func (b *builder) Cross(left, right Rel) (*CrossRel, error) {
 	}, nil
 }
 
-func (b *builder) Fetch(input Rel, offset, count int64) (*FetchRel, error) {
+func (b *builder) Fetch(input Rel, offset, count *expr.Expression) (*FetchRel, error) {
 	if input == nil {
 		return nil, errNilInputRel
 	}
--- a/plan/plan.go
+++ b/plan/plan.go
@@ -549,18 +549,34 @@ func RelFromProto(rel *proto.Rel, reg expr.ExtensionRegistry) (Rel, error) {
 			return nil, fmt.Errorf("error getting input to FetchRel: %w", err)
 		}
 
-		var offset int64
-		if off, ok := rel.Fetch.OffsetMode.(*proto.FetchRel_Offset); ok {
-			offset = off.Offset
-		} else {
-			return nil, fmt.Errorf("%w: missing required Offset field for Fetch Relation", substraitgo.ErrInvalidRel)
+		base := input.RecordType()
+
+		var offset *expr.Expression = nil
+		switch om := rel.Fetch.OffsetMode.(type) {
+		case *proto.FetchRel_Offset:
+			e := expr.Expression(expr.NewPrimitiveLiteral(om.Offset, false))
+			offset = &e
+		case *proto.FetchRel_OffsetExpr:
+			e, exprErr := expr.ExprFromProto(om.OffsetExpr, &base, reg)
+			if exprErr != nil {
+				return nil, fmt.Errorf("error getting offset expression for FetchRel: %w", exprErr)
+			}
+			offset = &e
 		}
 
-		var count int64
-		if cnt, ok := rel.Fetch.CountMode.(*proto.FetchRel_Count); ok {
-			count = cnt.Count
-		} else {
-			return nil, fmt.Errorf("%w: missing required Count field for Fetch Relation", substraitgo.ErrInvalidRel)
+		var count *expr.Expression = nil
+		switch cm := rel.Fetch.CountMode.(type) {
+		case *proto.FetchRel_Count:
+			if cm.Count != FETCH_COUNT_ALL_RECORDS {
+				e := expr.Expression(expr.NewPrimitiveLiteral(cm.Count, false))
+				count = &e
+			}
+		case *proto.FetchRel_CountExpr:
+			e, exprErr := expr.ExprFromProto(cm.CountExpr, &base, reg)
+			if exprErr != nil {
+				return nil, fmt.Errorf("error getting count expression for FetchRel: %w", exprErr)
+			}
+			count = &e
 		}
 
 		out := &FetchRel{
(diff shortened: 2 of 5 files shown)


plan/builders.go
```diff
@@ -74,10 +74,15 @@ type Builder interface {
...
-	// Fetch constructs a fetch relation providing an offset (skipping some number of
-	// rows) and a count (restricting output to a maximum number of rows).  If count
-	// is FETCH_COUNT_ALL_RECORDS (-1) all records will be returned.
-	Fetch(input Rel, offset, count int64) (*FetchRel, error)
+	// Fetch constructs a Fetch relation that skips offset rows and returns at most count rows from the input.
+	//
+	// offset must evaluate to a non-negative integer or Substrait null. Pass a Go nil for no offset.
+	// count must evaluate to a non-negative integer or Substrait null. Pass a Go nil to return all records.
+	Fetch(input Rel, offset, count *expr.Expression) (*FetchRel, error)
+	// Limit constructs a Fetch relation that returns at most count rows from the input with no offset.
```
benbellick: IIUC this PR is really doing two things:
1. alters the `Fetch` API to allow expressions for `offset` and `count`, and
2. adds nice helpers for special cases of `Fetch` allowing for `Limit` and `Offset`. 

Both of those are good ideas, but we should either update the PR title to reflect both of those things or separate them into two PRs. I am fine either way, though it may make it easier to review it if its split.

benbellick: ah i see now its a bit easier to follow one commit at a time.

plan/builders.go
```diff
@@ -312,7 +313,7 @@ func (b *builder) Cross(left, right Rel) (*CrossRel, error) {
 	}, nil
 }
 
-func (b *builder) Fetch(input Rel, offset, count int64) (*FetchRel, error) {
+func (b *builder) Fetch(input Rel, offset, count *expr.Expression) (*FetchRel, error) {
```
benbellick: It would be nice to add a type check here to make sure `offset` + `count` are inded non-negative integers like I am trying to do here https://github.com/substrait-io/substrait-java/pull/1074

But that may require access to an extension registry and I'm not sure we have access to that here without some kind of refactor.

plan/plan.go
```diff
@@ -541,18 +541,34 @@ func RelFromProto(rel *proto.Rel, reg expr.ExtensionRegistry) (Rel, error) {
...
-		if cnt, ok := rel.Fetch.CountMode.(*proto.FetchRel_Count); ok {
-			count = cnt.Count
-		} else {
-			return nil, fmt.Errorf("%w: missing required Count field for Fetch Relation", substraitgo.ErrInvalidRel)
+		base := input.RecordType()
+
+		var offset *expr.Expression = nil
+		switch om := rel.Fetch.OffsetMode.(type) {
+		case *proto.FetchRel_Offset:
+			e := expr.Expression(expr.NewPrimitiveLiteral(om.Offset, false))
```
benbellick: Ah but at the very least here we could check the type I think?

plan/relations.go
```diff
@@ -1124,14 +1125,27 @@ func (f *FetchRel) Copy(newInputs ...Rel) (Rel, error) {
...
 	}
-	return f.Copy(newInputs...)
+	if f.count != nil {
+		newCount, err := rewrite(*f.count)
+		if err != nil {
+			return nil, err
+		}
+		fetch.count = &newCount
+	}
+	return &fetch, nil
```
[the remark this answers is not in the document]
benbellick: I think this is legit: 
https://github.com/substrait-io/substrait-go/blob/c2b82be366bd8114c21a46101a668542f5899d53/plan/plan.go#L423-L425

we should either update the comment or update the code to conform to that comment.

benbellick (approved):
Generally looks good to me but we should address the coderabbit comment and improve test coverage before merging. Thanks!

plan/builders.go
vbarua: I opted to remove the Limit and Offset APIs for now.

plan/builders.go
vbarua: We can do a shallow-ish check using GetType. I say shallow because for things like FieldRefs it uses the provided type, and doesn't actually verify that the provided type is correct, but that's a separate problem if the plan provided and derived types don't line up.

plan/plan.go
vbarua: I've added a type check to the builder, that way we can centralize the check and not have it duplicated everywhere.

plan/builders.go
```diff
@@ -312,11 +313,29 @@ func (b *builder) Cross(left, right Rel) (*CrossRel, error) {
...
+	}
+	return false
+}
+
+func (b *builder) Fetch(input Rel, offset, count *expr.Expression) (*FetchRel, error) {
 	if input == nil {
 		return nil, errNilInputRel
 	}
 
+	if offset != nil && !isIntegerType((*offset).GetType()) {
```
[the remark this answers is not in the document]
vbarua: I don't think this is a problem in practice?

plan/relations.go
```diff
@@ -1065,17 +1065,22 @@ type FetchRel struct {
...
 }
-func (f *FetchRel) Input() Rel    { return f.input }
-func (f *FetchRel) Offset() int64 { return f.offset }
-func (f *FetchRel) Count() int64  { return f.count }
+func (f *FetchRel) Input() Rel { return f.input }
+
+// Offset returns nil if no offset is set, otherwise returns the offset expression
+func (f *FetchRel) Offset() expr.Expression { return f.offset }
+
+// Count returns nil if no count is set, otherwise returns the offset expression
```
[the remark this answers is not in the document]
benbellick: yes this should be fixed.

plan/builders.go
```diff
@@ -312,11 +313,29 @@ func (b *builder) Cross(left, right Rel) (*CrossRel, error) {
...
 	if input == nil {
 		return nil, errNilInputRel
 	}
 
+	if offset != nil && !isIntegerType(offset.GetType()) {
+		return nil, fmt.Errorf("%w: offset for Fetch Relation must yield an integer type, not %s",
+			substraitgo.ErrInvalidArg, offset.GetType())
+	}
+
+	if count != nil && !isIntegerType(count.GetType()) {
```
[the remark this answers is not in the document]
benbellick: I think this is excessive and not such a big deal.

plan/builders.go
```diff
@@ -74,10 +74,11 @@ type Builder interface {
...
 	CreateTableAsSelect(input Rel, tableName []string, schema types.NamedStruct) (*NamedTableWriteRel, error)
 	Cross(left, right Rel) (*CrossRel, error)
-	// Fetch constructs a fetch relation providing an offset (skipping some number of
-	// rows) and a count (restricting output to a maximum number of rows).  If count
-	// is FETCH_COUNT_ALL_RECORDS (-1) all records will be returned.
-	Fetch(input Rel, offset, count int64) (*FetchRel, error)
+	// Fetch constructs a Fetch relation that skips offset rows and returns at most count rows from the input.
+	//
+	// offset must evaluate to a non-negative integer or Substrait null. Pass a Go nil for no offset.
+	// count must evaluate to a non-negative integer or Substrait null. Pass a Go nil to return all records.
```
benbellick:
```suggestion
	// Fetch constructs a Fetch relation that skips offset rows and returns at most count rows from the input.
	//
	// offset must evaluate to a non-negative integer or Substrait null. Pass nil for no offset.
	// A Substrait null offset is treated as 0.
	// count must evaluate to a non-negative integer or Substrait null. Pass nil to return all records.
	// A Substrait null count returns all records.
```
nit: no need to say "go" nil rather than just nil + good to document the behavior of null values here.

plan/builders.go
```diff
@@ -312,11 +313,29 @@ func (b *builder) Cross(left, right Rel) (*CrossRel, error) {
 	}, nil
 }
 
-func (b *builder) Fetch(input Rel, offset, count int64) (*FetchRel, error) {
+func isIntegerType(t types.Type) bool {
```
benbellick, line 318: nice

plan/plan_builder_test.go
```diff
@@ -562,19 +564,54 @@ func TestFetchRelErrors(t *testing.T) {
...
 
-	f, err = b.Fetch(scan, 0, 0)
+	f, err = b.Fetch(scan, zeroExpr, zeroExpr)
 	assert.NoError(t, err)
 	_, err = f.Remap(2)
 	assert.ErrorIs(t, err, substraitgo.ErrInvalidRel)
 	assert.ErrorContains(t, err, "output mapping index out of range")
 }
 
+func TestFetchRelTypeErrors(t *testing.T) {
```
benbellick, line 580: nice

benbellick (approved):

PR #296, opened 2026-08-20
fix: correct StructFieldRef.GetType bounds check to reject negative and equal-to-len indices

Broshen:
This PR fixes a bug where StructFieldRef.GetType panics instead of returning ErrInvalidType due to an off-by-one comparison, and also for negative indices


vbarua (approved):
I love a no-brainer improvement PR. Thanks for the clear test, and clear improvement.

PR #297, opened 2026-08-20
feat(type)!: define Version instead of aliasing the protobuf message

kadinrabo:
`types.Version` was an alias for the generated message, so anything depending on substrait-go picked up substrait-protobuf through it. Same move #282 made for `Nullability`, per the plan in #280. A message alias breaks harder than an enum one: the type changes, it loses the `proto.Message` method set, and `String()` reports `"0.29.0 (substrait-go)"` rather than prototext.

It also fixes a bug the alias was hiding. `Plan.ToProto()` handed back `&plan.CurrentVersion` itself, so writing to the returned version rewrote the package global. `VersionToProto` copies now, pinned by `TestPlanToProtoCopiesTheVersion`.


kadinrabo:
Closing, moving this to a fork branch for now.

PR #300, opened 2026-08-21
feat(type)!: make Version a domain struct

kadinrabo:
`types.Version` was `= proto.Version`, so the generated message was substrait-go's public API. It becomes substrait-go's own struct with the same field names and the same wire numbers, so consumers still compile.

BREAKING CHANGE: `types.Version` is no longer an alias for `proto.Version`.

Part of #280.

### DIFF
--- a/expr/expression.go
+++ b/expr/expression.go
@@ -1700,7 +1700,7 @@ func ExtendedFromProto(ex *proto.ExtendedExpression, c *extensions.Collection) (
 	}
 
 	return &Extended{
-		Version:          ex.Version,
+		Version:          types.VersionFromProto(ex.Version),
 		Extensions:       extSet,
 		ReferredExpr:     refs,
 		BaseSchema:       base,
@@ -1718,7 +1718,7 @@ func (ex *Extended) ToProto() *proto.ExtendedExpression {
 	}
 
 	return &proto.ExtendedExpression{
-		Version:            ex.Version,
+		Version:            types.VersionToProto(ex.Version),
 		ExtensionUrns:      urns,
 		Extensions:         decls,
 		BaseSchema:         ex.BaseSchema.ToProto(),
--- a/expr/expression.go
+++ b/expr/expression.go
@@ -1658,7 +1658,7 @@ func (er *ExpressionReference) GetExpr() Expression            { return er.expr 
 func (er *ExpressionReference) GetMeasure() *AggregateFunction { return er.measure }
 
 type Extended struct {
-	Version          *types.Version
+	Version          types.Version
 	Extensions       extensions.Set
 	ReferredExpr     []ExpressionReference
 	BaseSchema       types.NamedStruct
@@ -1667,6 +1667,9 @@ type Extended struct {
 
 	reg ExtensionRegistry
 }
+
+// HasVersion reports whether the expression declared a version (false when Version is types.UnsetVersion).
+func (ex *Extended) HasVersion() bool { return ex.Version != types.UnsetVersion }
 
 func ExtendedFromProto(ex *proto.ExtendedExpression, c *extensions.Collection) (*Extended, error) {
 	extSet, err := extensions.GetExtensionSet(ex, c)
@@ -1699,8 +1702,12 @@ func ExtendedFromProto(ex *proto.ExtendedExpression, c *extensions.Collection) (
 		}
 	}
 
+	version := types.UnsetVersion
+	if ex.Version != nil {
+		version = *types.VersionFromProto(ex.Version)
+	}
 	return &Extended{
-		Version:          types.VersionFromProto(ex.Version),
+		Version:          version,
 		Extensions:       extSet,
 		ReferredExpr:     refs,
 		BaseSchema:       base,
@@ -1718,7 +1725,7 @@ func (ex *Extended) ToProto() *proto.ExtendedExpression {
 	}
 
 	return &proto.ExtendedExpression{
-		Version:            types.VersionToProto(ex.Version),
+		Version:            types.VersionToProto(&ex.Version),
 		ExtensionUrns:      urns,
 		Extensions:         decls,
 		BaseSchema:         ex.BaseSchema.ToProto(),
--- a/expr/expression.go
+++ b/expr/expression.go
@@ -1668,9 +1668,6 @@ type Extended struct {
 	reg ExtensionRegistry
 }
 
-// HasVersion reports whether the expression declared a version (false when Version is types.UnsetVersion).
-func (ex *Extended) HasVersion() bool { return ex.Version != types.UnsetVersion }
-
 func ExtendedFromProto(ex *proto.ExtendedExpression, c *extensions.Collection) (*Extended, error) {
 	extSet, err := extensions.GetExtensionSet(ex, c)
 	if err != nil {
@@ -1702,12 +1699,8 @@ func ExtendedFromProto(ex *proto.ExtendedExpression, c *extensions.Collection) (
 		}
 	}
 
-	version := types.UnsetVersion
-	if ex.Version != nil {
-		version = *types.VersionFromProto(ex.Version)
-	}
 	return &Extended{
-		Version:          version,
+		Version:          types.VersionFromProto(ex.Version),
 		Extensions:       extSet,
 		ReferredExpr:     refs,
 		BaseSchema:       base,
@@ -1725,7 +1718,7 @@ func (ex *Extended) ToProto() *proto.ExtendedExpression {
 	}
 
 	return &proto.ExtendedExpression{
-		Version:            types.VersionToProto(&ex.Version),
+		Version:            types.VersionToProto(ex.Version),
 		ExtensionUrns:      urns,
 		Extensions:         decls,
 		BaseSchema:         ex.BaseSchema.ToProto(),
--- a/plan/plan.go
+++ b/plan/plan.go
@@ -143,7 +143,7 @@ type AdvancedExtension interface {
 // Plan describes a set of operations to complete. For
 // compactness, identifiers are normalized at the plan level.
 type Plan struct {
-	version           *types.Version
+	version           types.Version
 	extensions        extensions.Set
 	expectedTypeURLs  []string
 	advExtension      *extensions.AdvancedExtension
@@ -153,8 +153,11 @@ type Plan struct {
 	reg expr.ExtensionRegistry
 }
 
-// Version returns the substrait version of the plan, or nil if the plan declares none.
-func (p *Plan) Version() *types.Version { return p.version }
+// Version returns the plan's version, or types.UnsetVersion if it declared none.
+func (p *Plan) Version() types.Version { return p.version }
+
+// HasVersion reports whether the plan declared a version (false when Version is types.UnsetVersion).
+func (p *Plan) HasVersion() bool { return p.version != types.UnsetVersion }
 
 // ExtensionRegistry returns the set of registered extensions for this plan
 // that it may depend on.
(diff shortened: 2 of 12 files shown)


types/types.go
```diff
@@ -16,7 +16,99 @@ import (
 	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
 )
 
-type Version = proto.Version
+// Version is declared here rather than aliased to the protobuf message so serialization can move
+// out of the core. Getters are nil receiver safe: Plan.Version() can hand out a nil *Version.
```
vbarua: I would avoid comments like this.
```
// Version is declared here rather than aliased to the protobuf message so serialization can move
// out of the core
```

Issues and PRs are were we can discuss _why_ we're making changes. The code should capture what it does.

This comment, for example, will be irrelevant once we finish our planned changes.

types/types.go
```diff
@@ -16,7 +16,99 @@ import (
...
+type Version struct {
+	MajorNumber uint32
+	MinorNumber uint32
+	PatchNumber uint32
+	GitHash     string
+	Producer    string
+}
```
vbarua, line 26: So right now we still have a Version interface in `plan.go`
https://github.com/substrait-io/substrait-go/blob/47adfb6b5a744400cb572363ceb9beb220a0eae9/plan/plan.go#L138-L145

Did you consider replacing that interface with this struct directly? Version conceptually doesn't feel like something we need to let end-users define custom behaviour beyond "You can set the Version fields".

types/version_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/substrait-io/substrait-go/v8/types"
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+// Plan.Version() can hand out a nil *Version through an interface, so the getters must be nil safe.
```
vbarua: > Plan.Version() can hand out a nil *Version

`version` is actually a required field in the protobufs. Maybe we can take advantage of this and enforce that it should be non-null.
https://github.com/substrait-io/substrait/blob/60925234a0a2c02a41400df349b0bad1e674581b/proto/substrait/plan.proto#L30-L32

types/types.go
kadinrabo: I didn't but that seems appropriate, took a pass at that

types/types.go
kadinrabo: Yeah that will be quickly inaccurate I'll clean that up

types/version_test.go
kadinrabo: We should, got that started in #303

types/types.go
```diff
@@ -16,7 +16,61 @@ import (
...
+// String reports a readable version like "0.29.0+abc123 (producer)".
+func (v *Version) String() string {
+	if v == nil {
+		return "<nil>"
+	}
+
+	var b strings.Builder
+	fmt.Fprintf(&b, "%d.%d.%d", v.MajorNumber, v.MinorNumber, v.PatchNumber)
+	if v.GitHash != "" {
+		// semver build metadata: the hash pins a commit between releases
```
vbarua: I don't think this comment adds that much. For one, what does `semver build metadata` mean? If I look at the code, it's pretty straightforward to see that this just appends the GitHash to the version string, which is pretty self-explanatory.

plan/plan.go
```diff
@@ -135,15 +135,6 @@ func (r *Relation) ToProto() *proto.PlanRel {
...
 }
 
-type Version interface {
-	GetGitHash() string
-	GetMajorNumber() uint32
-	GetMinorNumber() uint32
-	GetPatchNumber() uint32
-	GetProducer() string
-	fmt.Stringer
-}
```
vbarua, line 145: meta: I was going to advocate that we move the struct definition here because `plan.Version` makes more sense (to me) than `types.Version`.

HOWEVER, because the `expr` package also uses Version, that would introduce a dependency cycle between `plan` and `expr`.

plan/plan.go
```diff
@@ -162,8 +153,8 @@ type Plan struct {
 	reg expr.ExtensionRegistry
 }
 
-// Version returns the substrait version of the plan.
-func (p *Plan) Version() Version { return p.version }
+// Version returns the substrait version of the plan, or nil if the plan declares none.
+func (p *Plan) Version() *types.Version { return p.version }
```
vbarua: In the future, when we enforce that Version is set in plans, we'll want to update this to return `types.Version` directly to indicate that.

Given that, would it make sense to just do that now, and then we process a proto Plan without set a default version like `0.0.0 (UNSET)` to help signal to users that their plans don't have version set correctly today?

You have a test that indicates that we want to roundtrip unset to unset, but is that worth retaining?

plan/version_test.go
```diff
@@ -0,0 +1,43 @@
...
+// A plan can hold no version, in which case Version() returns nil.
+func TestPlanWithoutAVersion(t *testing.T) {
+	p, err := plan.FromProto(&substraitproto.Plan{}, extensions.GetDefaultCollectionWithNoError())
+	require.NoError(t, err)
+
+	assert.Nil(t, p.Version())
+
+	roundTrip, err := p.ToProto()
+	require.NoError(t, err)
+	assert.Nil(t, roundTrip.Version, "an absent version must not come back as an empty message")
```
vbarua: See above comment, I think we should just set a default version with a big UNSET in it to make it clear something is up in the output.

plan/version_test.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// ToProto used to hand out &CurrentVersion itself. It now encodes a copy, so editing the returned
+// message no longer rewrites the version every later plan reports.
```
vbarua: > ToProto used to hand out &CurrentVersion itself.

Why do I care what this used to outside of this PR? In the code, what matters is what it does today.

vbarua (commented):
Left some more comments.

The main question though is: Should we make the domain Version allowed to be nullable, or should we make the domain Version be required and create a (default) version if it's not set in the protos.

plan/plan.go
kadinrabo: Doesn't seem too complicated to address here or too hard to verify when we move to the codec, pushed 👍

types/types.go
```diff
@@ -16,7 +16,58 @@ import (
...
+type Version struct {
+	MajorNumber uint32
+	MinorNumber uint32
+	PatchNumber uint32
+	GitHash     string
+	Producer    string
+}
+
+// UnsetVersion marks a version-less plan or expression; it renders as "0.0.0 (UNSET)".
+var UnsetVersion = Version{Producer: "UNSET"}
```
vbarua: I would make this internal only to avoid exposing it to users. We will eventually want to delete this.

expr/expression.go
```diff
@@ -1668,6 +1668,9 @@ type Extended struct {
 	reg ExtensionRegistry
 }
 
+// HasVersion reports whether the expression declared a version (false when Version is types.UnsetVersion).
+func (ex *Extended) HasVersion() bool { return ex.Version != types.UnsetVersion }
```
vbarua: IMO we should avoid exposing this methods. Users can check the version directly if they care about it, but we will remove this entirely at some point when we enforce that proto plans have versions set on them correctly.

plan/plan.go
```diff
@@ -162,8 +153,11 @@ type Plan struct {
 	reg expr.ExtensionRegistry
 }
 
-// Version returns the substrait version of the plan.
-func (p *Plan) Version() Version { return p.version }
+// Version returns the plan's version, or types.UnsetVersion if it declared none.
```
vbarua:
```suggestion
// Version returns the plans version
```

types.UnsetVersion is a valid version. In the library we shouldn't need to distinguish between the 2, that's purely at the protobuf consumer level.

types/version_test.go
```diff
@@ -0,0 +1,104 @@
...
+	} {
+		t.Run(td.name, func(t *testing.T) {
+			assert.Equal(t, td.expected, td.version.String())
+		})
+	}
+}
+
+// Pin the hand written struct to the generated descriptor: the field set and wire numbers, which
+// a spec change can move without breaking the build.
+func TestVersionMatchesDescriptor(t *testing.T) {
```
vbarua, line 50: If I understand what this is doing, it's checking that the `Version` protobuf message only has fields we expect, to guard against new field additions? that seems reasonable, but I might clarify either the document or string name to indicate what this checks.

>// Pin the hand written struct to the generated descriptor: the field set and wire numbers, which
// a spec change can move without breaking the build.

Didn't really tell me that.

types/version_test.go
```diff
@@ -0,0 +1,104 @@
...
+	for i := 0; i < fields.Len(); i++ {
+		f := fields.Get(i)
+		ourField, ok := ours[f.Name()]
+		require.Truef(t, ok, "no types.Version field covers %s", f.Name())
+		assert.EqualValues(t, ourField.number, f.Number(), "wire number for %s", f.Name())
+	}
+}
+
+// Both directions carry every field and leave an absent version absent. A non-hash git value
+// ("HEAD") shows the hash is copied as given, not validated.
```
vbarua: The test name alone tells me what this is doing, and the comment doesn't add much clarity.

types/version_test.go
```diff
@@ -0,0 +1,104 @@
...
+func TestVersionRoundTrip(t *testing.T) {
+	for _, td := range []struct {
+		name string
+		wire *proto.Version
```
vbarua: minor: wire -> protoVersion

types/types.go
```diff
@@ -16,7 +16,58 @@ import (
...
+		MajorNumber: v.MajorNumber,
+		MinorNumber: v.MinorNumber,
+		PatchNumber: v.PatchNumber,
+		GitHash:     v.GitHash,
+		Producer:    v.Producer,
+	}
+}
+
+// VersionToProto encodes a version; a nil pointer stays absent.
+func VersionToProto(v *Version) *proto.Version {
```
vbarua: Could this just be `func VersionToProto(v Version)` now? `Version` should never be nil on the Go side?

types/types.go
```diff
@@ -16,7 +16,58 @@ import (
...
+		b.WriteString(" (" + v.Producer + ")")
+	}
+	return b.String()
+}
+
+// VersionFromProto converts a protobuf version message to the domain Version.
+func VersionFromProto(v *proto.Version) *Version {
+	if v == nil {
+		return nil
+	}
```
vbarua, line 49: Similarly, could this be `VersionFromProto(v proto.Version) Version` and the only place we need to check for the `nil` proto.Version is in the consumer?

plan/plan.go
kadinrabo: Applied 👍

types/version_test.go
kadinrabo: Sounds good 👍

types/version_test.go
kadinrabo: Removed 👍

types/types.go
kadinrabo: Yes it wouldn't be nil that would work

types/types.go
kadinrabo: Tried that out and got:

copylocks: VersionFromProto passes lock by value: github.com/substrait-io/substrait-protobuf/go/substraitpb.Version contains google.golang.org/protobuf/runtime/protoimpl.MessageState contains sync.Mutex (govet)

https://github.com/kadinrabo/substrait-go/actions/runs/32860172888

types/version_test.go
```diff
@@ -0,0 +1,98 @@
...
+	} {
+		t.Run(td.name, func(t *testing.T) {
+			assert.Equal(t, td.expected, td.version.String())
+		})
+	}
+}
+
+// Guard against spec drift: fail if the Version proto message gains, drops, or renumbers a field
+// that the hand-written types.Version does not mirror.
+func TestVersionMatchesDescriptor(t *testing.T) {
```
vbarua, line 50: Great comment on this. I know exactly what this is for now (and how to fix it if it fails).

vbarua (approved):
Changes look great, thanks for working with me on this.

PR #301, opened 2026-08-21
build: bump go.mod to v9

vbarua:
BREAKING CHANGE: import path changes to substrait-go/v9


vbarua:
We are still actively working to extract out substrait-protobuf from substrait-go as per https://github.com/substrait-io/substrait-go/issues/280. We don't plan to cut an official v9 release until then.

However, in the interim it would still be beneficial to be able to publish new releases with our ongoing changes to:
1. Make sure downstreams can integrate with them.
2. Allow any users waiting on a feature they integrated to access them earlier.

What we can do for this is bump the module to v9 and start tagging pre-release alpha versions as per: https://go.dev/doc/modules/version-numbers#pre-release

benbellick (approved):

PR #305, opened 2026-08-25
feat(type)!: make CastFailBehavior a domain enum

kadinrabo:
`types.CastFailBehavior` was `= proto.Expression_Cast_FailureBehavior`, so the generated enum was substrait-go's public API. It becomes substrait-go's own `int32` with the same wire numbers.

BREAKING CHANGE: `types.CastFailBehavior` is no longer an alias for `proto.Expression_Cast_FailureBehavior`, and the `Behavior*` constants are renamed to `CastFailBehavior*`.

Part of #280.

### DIFF
--- a/types/cast_fail_behavior_test.go
+++ b/types/cast_fail_behavior_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/substrait-io/substrait-go/v9/types"
-	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
 )
 
 func TestCastFailBehaviorString(t *testing.T) {
@@ -26,4 +25,3 @@ func TestCastFailBehaviorString(t *testing.T) {
 		})
 	}
 }
-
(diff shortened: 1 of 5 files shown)


types/cast_fail_behavior_test.go
```diff
@@ -0,0 +1,46 @@
...
+		{types.CastFailBehaviorThrowException, proto.Expression_Cast_FAILURE_BEHAVIOR_THROW_EXCEPTION},
+	}
+	// fail if the spec adds a value we don't mirror
+	assert.Equal(t, proto.Expression_Cast_FAILURE_BEHAVIOR_UNSPECIFIED.Descriptor().Values().Len(), len(cases),
+		"a proto cast failure behavior value is not covered")
+	for _, td := range cases {
+		assert.EqualValues(t, td.pb, td.domain, "wire number for %s", td.pb)
+		assert.Equal(t, td.pb.String(), td.domain.String(), "enum name for %s", td.pb)
+	}
+}
```
benbellick:
```suggestion
```
nit: this isn't an unreasonable test but I don't think there is precedent for this kind of proto coverage in other places in the codebase and IMO its fine to just catch these as they come up.

benbellick (approved):

PR #306, opened 2026-08-25
feat(extensions): expose function variant return types

### LINKED ISSUE #299: Expose unresolved return type on ScalarFunctionVariant
[`ScalarFunctionVariant`](https://github.com/substrait-io/substrait-go/blob/47adfb6b5a744400cb572363ceb9beb220a0eae9/extensions/variants.go#L337-L377) exposes unresolved input definitions through `Args()`, but not the corresponding unresolved output definition stored in [`ScalarFunctionImpl.Return`](https://github.com/substrait-io/substrait-go/blob/47adfb6b5a744400cb572363ceb9beb220a0eae9/extensions/simple_extension.go#L235-L243). The variant API should expose both for consistency.

benbellick:
Closes #299

---

### DIFF
--- a/extensions/variants.go
+++ b/extensions/variants.go
@@ -352,6 +352,15 @@ func (s *ScalarFunctionVariant) SessionDependent() bool           { return s.imp
 func (s *ScalarFunctionVariant) Nullability() NullabilityHandling { return s.impl.Nullability }
 func (s *ScalarFunctionVariant) URN() string                      { return s.urn }
 func (s *ScalarFunctionVariant) Metadata() map[string]any         { return s.metadata }
+
+// ReturnType returns the unresolved return type expression, or nil when the
+// variant does not define one.
+func (s *ScalarFunctionVariant) ReturnType() types.FuncDefArgType {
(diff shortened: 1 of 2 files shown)


extensions/variants.go
```diff
@@ -352,6 +352,15 @@ func (s *ScalarFunctionVariant) SessionDependent() bool           { return s.imp
...
 func (s *ScalarFunctionVariant) Metadata() map[string]any         { return s.metadata }
+
+// ReturnType returns the unresolved return type expression, or nil when the
+// variant does not define one.
+func (s *ScalarFunctionVariant) ReturnType() types.FuncDefArgType {
+	if s.impl.Return == nil {
+		return nil
+	}
+	return s.impl.Return.ValueType
+}
```
vbarua, line 363: Would it make sense to add this for aggregate and window functions as well?

benbellick: yup, it would :) 

done!

vbarua (approved):
Makes sense to me that we should be able to retrieve these easily.

PR #308, opened 2026-08-28
feat(type)!: make VarChar a domain struct

kadinrabo:
`types.VarChar` was `= proto.Expression_Literal_VarChar`, so the generated message was substrait-go's public API. It becomes substrait-go's own struct with the same field names and the same wire numbers.

BREAKING CHANGE: `types.VarChar` is no longer an alias for `proto.Expression_Literal_VarChar`.

Part of #280.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -276,13 +276,19 @@ func (b CastFailBehavior) String() string {
 type (
 	IntervalYearToMonth  = proto.Expression_Literal_IntervalYearToMonth
 	IntervalDayToSecond  = proto.Expression_Literal_IntervalDayToSecond
-	VarChar              = proto.Expression_Literal_VarChar
 	Decimal              = proto.Expression_Literal_Decimal
 	UserDefinedLiteral   = proto.Expression_Literal_UserDefined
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -276,14 +276,20 @@ func (b CastFailBehavior) String() string {
...
 	Decimal              = proto.Expression_Literal_Decimal
 	UserDefinedLiteral   = proto.Expression_Literal_UserDefined
 	PrecisionTime        = proto.Expression_Literal_PrecisionTime
 	PrecisionTimestamp   = proto.Expression_Literal_PrecisionTimestamp_
 	PrecisionTimestampTz = proto.Expression_Literal_PrecisionTimestampTz
 )
 
+// VarChar is a variable-length character literal: its value and length, mirroring the fields of
+// the Substrait VarChar literal message.
+type VarChar struct {
```
benbellick, line 288: Not for this PR, but I realize I was a bit confused about how things are organized here. I assumed that `types/types.go` was for representing types, and some other file, maybe `literals.go` was for representing literals. However, here was a type representing a literal. We should probably make that distinction a bit clearer in the future.

benbellick (approved):

PR #313, opened 2026-09-01
feat(type)!: make BoundsType a domain enum

kadinrabo:
`types.BoundsType` was `= proto.Expression_WindowFunction_BoundsType`, so the generated enum was substrait-go's public API. It becomes substrait-go's own `int32` with the same constant names and the same wire numbers plus a hand-written `String()`, casting at the proto boundaries, so consumers still compile.

BREAKING CHANGE: `types.BoundsType` is no longer an alias for `proto.Expression_WindowFunction_BoundsType`; its constants are now typed `BoundsType` values. Code that passes one where the other is expected needs a cast.

Part of #280.


benbellick (approved):
💯

PR #317, opened 2026-09-02
feat(type)!: make IntervalDayToSecond a domain type

kadinrabo:
`types.IntervalDayToSecond` was `= proto.Expression_Literal_IntervalDayToSecond`, so the generated protobuf message was substrait-go's public API. It becomes a substrait-go struct with the same fields, and its `precision_mode` oneof becomes a small `IntervalDayToSecondPrecisionMode` interface (Microseconds / Precision arms). Conversion to and from the protobuf message lives in `IntervalDayToSecondToProto` / `IntervalDayToSecondFromProto`, so consumers still compile.

BREAKING CHANGE: `types.IntervalDayToSecond` is no longer an alias for `proto.Expression_Literal_IntervalDayToSecond`; it is a plain struct whose sub-second precision is carried by the `PrecisionMode` oneof interface. Code that used the protobuf message directly must construct the struct.

Part of #280.

### DIFF
--- a/types/interval_day_to_second_test.go
+++ b/types/interval_day_to_second_test.go
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package types_test
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/substrait-io/substrait-go/v9/types"
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+func TestIntervalDayToSecondMatchesDescriptor(t *testing.T) {
+	// Scalar fields are pinned by wire number and kind; the precision_mode oneof is pinned
+	// separately since the domain type collapses its two arms into a single PrecisionMode field.
+	wantScalars := map[protoreflect.Name]struct {
+		number protoreflect.FieldNumber
+		kind   protoreflect.Kind
+	}{
+		"days":       {1, protoreflect.Int32Kind},
+		"seconds":    {2, protoreflect.Int32Kind},
+		"subseconds": {5, protoreflect.Int64Kind},
+	}
+	wantOneof := map[protoreflect.Name]struct {
+		number protoreflect.FieldNumber
+		kind   protoreflect.Kind
+	}{
+		"microseconds": {3, protoreflect.Int32Kind},
+		"precision":    {4, protoreflect.Int32Kind},
+	}
+
+	desc := (&proto.Expression_Literal_IntervalDayToSecond{}).ProtoReflect().Descriptor()
+	fields := desc.Fields()
+	require.Equal(t, len(wantScalars)+len(wantOneof), fields.Len(), "spec interval_day_to_second field set changed")
+
+	// 3 scalars + 1 PrecisionMode field standing in for the 2-arm oneof.
+	require.Equal(t, len(wantScalars)+1, reflect.TypeOf(types.IntervalDayToSecond{}).NumField(),
+		"types.IntervalDayToSecond field count drifted from the spec")
+
+	for i := 0; i < fields.Len(); i++ {
+		f := fields.Get(i)
+		if f.ContainingOneof() != nil {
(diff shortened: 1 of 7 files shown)


types/types.go
```diff
@@ -336,6 +335,82 @@ type IntervalYearToMonth struct {
...
 }
 
+// IntervalDayToSecond is an interval literal spanning days down to sub-seconds, mirroring the
+// fields of the Substrait IntervalDayToSecond literal message. Its sub-second precision is carried
+// by the PrecisionMode oneof.
+type IntervalDayToSecond struct {
+	Days          int32
+	Seconds       int32
+	Subseconds    int64
+	PrecisionMode IntervalDayToSecondPrecisionMode
```
benbellick: I think this is a case where we actually don't want to perfectly model things the way they are in the protobuf. Yes `PrecisionMode` can represent one of two things, but microseconds is deprecated on this version. 

If we leave it like this, it is ambiguous the meaning when `subseconds` is set AND `PrecisionMode=Microseconds{<some val>}`. I think it makes more sense to just only use `PrecisionMode TimePrecision` and handle any awkwardness in the protobuf conversion code. 

I realize this expands the scope of the PR a tiny bit.

benbellick (changes requested):

kadinrabo:
pushed an update dropping the deprecated microseconds arm from the domain type

types/types.go
```diff
@@ -336,6 +335,66 @@ type IntervalYearToMonth struct {
...
+// receiver.
+func (i *IntervalDayToSecond) GetPrecisionProtoVal() int32 {
+	if i == nil {
+		return 0
+	}
+	return i.Precision.ToProtoVal()
+}
+
+// IntervalDayToSecondToProto encodes the domain interval as its protobuf literal message. It always
+// writes the precision precision_mode arm; the deprecated microseconds arm is never emitted.
```
benbellick, line 358: This is good, we will just have to remember to update this comment when we bump the substrait version.

types/types.go
```diff
@@ -336,6 +335,66 @@ type IntervalYearToMonth struct {
...
+		PrecisionMode: &proto.Expression_Literal_IntervalDayToSecond_Precision{
+			Precision: v.Precision.ToProtoVal(),
+		},
+	}
+}
+
+// IntervalDayToSecondFromProto decodes a protobuf interval literal message into the domain type.
+// The deprecated microseconds precision_mode arm, and an absent precision_mode (the legacy
+// encoding), are both normalized to microsecond precision, matching IntervalDayType so the domain
+// type never carries the deprecated representation.
```
benbellick:
```suggestion
// IntervalDayToSecondFromProto decodes a protobuf interval literal message into the domain type.
```
I'm not sure that is worth explaining here.

types/interval_day_to_second_test.go
```diff
@@ -0,0 +1,110 @@
...
+			PrecisionMode: &proto.Expression_Literal_IntervalDayToSecond_Microseconds{Microseconds: 7},
+		}
+		assert.Equal(t,
+			&types.IntervalDayToSecond{Days: 4, Seconds: 5, Subseconds: 7, Precision: types.PrecisionMicroSeconds},
+			types.IntervalDayToSecondFromProto(p))
+	})
+
+	// An absent precision_mode is the legacy encoding and decodes as microsecond precision,
+	// matching IntervalDayType's default.
+	t.Run("absent precision_mode defaults to microseconds", func(t *testing.T) {
```
benbellick: I would actually consider this an error:
https://github.com/substrait-io/substrait/blob/v0.87.0/proto/substrait/algebra.proto#L1099-L1112

types/interval_day_to_second_test.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Round-trip domain->proto->domain always uses the non-deprecated precision arm, so a swapped or
+// dropped field fails here.
+func TestIntervalDayToSecondRoundTrip(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+		in   *types.IntervalDayToSecond
+	}{
+		{
```
benbellick: nit: honestly, I think just one case is enough here though two doesn't hurt.

types/interval_day_to_second_test.go
benbellick: Everything else is a nit but this should be solved before we merge IMO.

benbellick (approved):
Approving but we should address remaining comments, specifically about the incorrect default behavior.

PR #319, opened 2026-09-02
feat(expr)!: make MaskListElement and MaskListSlice domain structs

kadinrabo:
`expr.MaskListElement` and `expr.MaskListSlice` were type definitions over the generated protobuf messages, so those messages were substrait-go's public API. They become substrait-go structs with the same fields, converting to and from the protobuf messages by explicit field construction instead of the previous unchecked pointer casts, so consumers still compile.

BREAKING CHANGE: `expr.MaskListElement` and `expr.MaskListSlice` are no longer defined over the protobuf messages; they are plain structs. Code relying on the old proto-typed layout must use the struct fields.

Part of #280.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of mask-list expressions when loading and saving serialized data.
  - Field selections and range selections now preserve their values more reliably during conversion.
  - Improved support for nested mask expressions and round-trip serialization, reducing the risk of lost or altered selections.
  - Mask-list ranges now retain their start and end positions consistently across serialization cycles.

### DIFF
--- a/expr/mask_list_test.go
+++ b/expr/mask_list_test.go
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package expr_test
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/substrait-io/substrait-go/v9/expr"
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+	pbproto "google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+func TestMaskListElementMatchesDescriptor(t *testing.T) {
+	want := map[protoreflect.Name]struct {
+		number protoreflect.FieldNumber
+		kind   protoreflect.Kind
+	}{
+		"field": {1, protoreflect.Int32Kind},
+	}
+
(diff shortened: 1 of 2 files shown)


benbellick:
@coderabbitai review

expr/mask_list_test.go
```diff
@@ -0,0 +1,104 @@
...
+	assert.EqualValues(t, 2, elem.GetField())
+	assert.EqualValues(t, 2, elem.ToProto().GetItem().GetField())
+
+	slice := &expr.MaskListSlice{Start: 1, End: 3}
+	start, end := slice.GetBounds()
+	assert.EqualValues(t, 1, start)
+	assert.EqualValues(t, 3, end)
+	assert.EqualValues(t, 1, slice.ToProto().GetSlice().GetStart())
+	assert.EqualValues(t, 3, slice.ToProto().GetSlice().GetEnd())
+
```
benbellick:
```suggestion
```
I think those cases are already covered by below?

expr/mask_list_test.go
```diff
@@ -0,0 +1,104 @@
...
+								Type: &proto.Expression_MaskExpression_Select_Struct{
+									Struct: &proto.Expression_MaskExpression_StructSelect{},
+								},
+							},
+						},
+					},
+				},
+			}},
+		},
+	}
```
benbellick, line 88: I wish there was a good way to make this test better but can't think of a clean solution, so this is okay for now

benbellick (approved):

PR #320, opened 2026-09-02
fix(expr): guard nil FunctionOption in WindowFunction.Equals

kadinrabo:
WindowFunction.Equals compares function options by reading `l.Name`/`r.Name` directly, so comparing a WindowFunction whose options slice contains a nil element panics with a nil-pointer dereference.

ScalarFunction.Equals does not have this problem because it compares options via `proto.Equal`, which is nil-safe. This brings WindowFunction into line by treating a nil option element as equal only to another nil, then comparing fields as before.

Added a regression test that panics without the guard.


kadinrabo:
Moving this to my fork — it's an incidental fix unrelated to the de-alias work; not ready for upstream review yet.

PR #321, opened 2026-09-02
feat(type)!: make FunctionOption a domain struct

kadinrabo:
Replace the `proto.FunctionOption` alias with a substrait-go struct plus slice converters at the proto seam, so the core packages no longer expose a protobuf type here.

`ScalarFunction.Equals` previously compared options with `proto.Equal`; it now compares the domain fields directly (name + preferences), preserving the same semantics including nil handling.

BREAKING CHANGE: `types.FunctionOption` is now a substrait-go struct, not an alias of `proto.FunctionOption`.

Part of #280.

### DIFF
--- a/expr/function_option_internal_test.go
+++ b/expr/function_option_internal_test.go
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package expr
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/substrait-io/substrait-go/v9/types"
+)
+
+func opts(pref string) []*types.FunctionOption {
+	return []*types.FunctionOption{{Name: "overflow", Preference: []string{pref}}}
+}
+
+// TestScalarFunctionEqualsOptions checks that ScalarFunction.Equals compares
+// FunctionOptions by name and preference, treating nil elements as equal.
+func TestScalarFunctionEqualsOptions(t *testing.T) {
+	ot := &types.Int64Type{Nullability: types.NullabilityRequired}
+	base := &ScalarFunction{funcRef: 1, outputType: ot, options: opts("ERROR")}
+	same := &ScalarFunction{funcRef: 1, outputType: ot, options: opts("ERROR")}
+	diffPref := &ScalarFunction{funcRef: 1, outputType: ot, options: opts("SILENT")}
+	noOpts := &ScalarFunction{funcRef: 1, outputType: ot}
+
+	assert.True(t, base.Equals(same))
+	assert.False(t, base.Equals(diffPref))
+	assert.False(t, base.Equals(noOpts))
+
+	// nil option elements must compare equal without dereferencing.
+	nilA := &ScalarFunction{funcRef: 1, outputType: ot, options: []*types.FunctionOption{nil}}
+	nilB := &ScalarFunction{funcRef: 1, outputType: ot, options: []*types.FunctionOption{nil}}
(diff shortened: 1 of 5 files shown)


expr/function_option_internal_test.go
```diff
@@ -0,0 +1,46 @@
...
+func TestFunctionGetOption(t *testing.T) {
+	o := opts("ERROR")
+	s := &ScalarFunction{options: o}
+	assert.Equal(t, []string{"ERROR"}, s.GetOption("overflow"))
+	assert.Nil(t, s.GetOption("missing"))
+
+	a := &AggregateFunction{options: o}
+	assert.Equal(t, []string{"ERROR"}, a.GetOption("overflow"))
+	assert.Nil(t, a.GetOption("missing"))
+}
```
benbellick:
```suggestion
```

I don't quite see the value of testing this here. We didn't really change the impl and sclar and aggreagate handling in this case are basically the same

types/function_option_test.go
```diff
@@ -0,0 +1,52 @@
...
+	opts := []*types.FunctionOption{
+		{Name: "rounding", Preference: []string{"TIE_TO_EVEN", "TIE_AWAY_FROM_ZERO"}},
+		{Name: "overflow"},
+	}
+	pb := types.FunctionOptionsToProto(opts)
+	require.Len(t, pb, len(opts))
+	assert.Equal(t, "rounding", pb[0].Name)
+	assert.Equal(t, []string{"TIE_TO_EVEN", "TIE_AWAY_FROM_ZERO"}, pb[0].Preference)
+	assert.Equal(t, opts, types.FunctionOptionsFromProto(pb))
+}
```
benbellick, line 45:
```suggestion
func TestFunctionOptionsRoundTrip(t *testing.T) {
	options := []*types.FunctionOption{
		{Name: "rounding", Preference: []string{"TIE_TO_EVEN", "TIE_AWAY_FROM_ZERO"}},
		{Name: "overflow"},
	}

	assert.Equal(t, options, types.FunctionOptionsFromProto(types.FunctionOptionsToProto(options)))
}
```
Simple is better here, also no need to enforce the `nil` behavior unless we actually think that is an important part of the contract.

expr/function_option_internal_test.go
```diff
@@ -0,0 +1,46 @@
...
+	assert.True(t, base.Equals(same))
+	assert.False(t, base.Equals(diffPref))
+	assert.False(t, base.Equals(noOpts))
+
+	// nil option elements must compare equal without dereferencing.
+	nilA := &ScalarFunction{funcRef: 1, outputType: ot, options: []*types.FunctionOption{nil}}
+	nilB := &ScalarFunction{funcRef: 1, outputType: ot, options: []*types.FunctionOption{nil}}
+	assert.True(t, nilA.Equals(nilB))
+	assert.False(t, nilA.Equals(base))
+}
```
benbellick:
```suggestion
func TestScalarFunctionEqualsOptions(t *testing.T) {
	outputType := &types.Int64Type{Nullability: types.NullabilityRequired}
	left := &ScalarFunction{funcRef: 1, outputType: outputType, options: opts("ERROR")}
	right := &ScalarFunction{funcRef: 1, outputType: outputType, options: opts("ERROR")}

	assert.True(t, left.Equals(right))
}
```
I think a simple test is sufficient here.

benbellick (commented):
I think the code mostly looks good but the tests could be simplified a bit. Thanks!

benbellick (approved):

PR #322, opened 2026-09-02
feat(type)!: make PrecisionTime, PrecisionTimestamp, PrecisionTimestampTz domain structs

kadinrabo:
Replace the `proto.Expression_Literal_PrecisionTime` /
`PrecisionTimestamp` / `PrecisionTimestampTz` aliases with substrait-go structs
holding `Precision`/`Value`, flattening the oneof-wrapper aliases at the
construction and read seams so the core packages no longer expose protobuf types
here.

**BREAKING CHANGE:** `types.PrecisionTime`, `types.PrecisionTimestamp`, and
`types.PrecisionTimestampTz` are now substrait-go structs, not aliases of their
protobuf literal messages / oneof wrappers.

Part of #280.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -307,11 +307,8 @@ func (b CastFailBehavior) String() string {
 }
 
 type (
-	IntervalDayToSecond  = proto.Expression_Literal_IntervalDayToSecond
-	UserDefinedLiteral   = proto.Expression_Literal_UserDefined
-	PrecisionTime        = proto.Expression_Literal_PrecisionTime
-	PrecisionTimestamp   = proto.Expression_Literal_PrecisionTimestamp_
-	PrecisionTimestampTz = proto.Expression_Literal_PrecisionTimestampTz
+	IntervalDayToSecond = proto.Expression_Literal_IntervalDayToSecond
+	UserDefinedLiteral  = proto.Expression_Literal_UserDefined
 )
 
 // VarChar is a variable-length character literal: its value and length, mirroring the fields of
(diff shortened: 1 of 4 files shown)


types/types.go
```diff
@@ -336,6 +333,29 @@ type IntervalYearToMonth struct {
 	Months int32
 }
 
+// PrecisionTime is a time-of-day literal: the number of precision units past
+// midnight, mirroring the fields of the Substrait PrecisionTime literal message.
+type PrecisionTime struct {
+	Precision int32
```
benbellick, line 339: hmm I wonder if we should make this a `TimePrecision` (which is just an alias to an `int32` anyways)?

types/types.go
```diff
@@ -336,6 +333,29 @@ type IntervalYearToMonth struct {
...
+	Value     int64
+}
+
+// PrecisionTimestamp is a timestamp literal in an unspecified time zone: the
+// number of precision units since the UNIX epoch, mirroring the fields of the
+// Substrait PrecisionTimestamp literal message.
+type PrecisionTimestamp struct {
+	Precision int32
```
benbellick, line 347: and this

types/types.go
```diff
@@ -336,6 +333,29 @@ type IntervalYearToMonth struct {
...
+	Value     int64
+}
+
+// PrecisionTimestampTz is a UTC timestamp literal: the number of precision units
+// since the UNIX epoch, mirroring the fields of the Substrait PrecisionTimestamp
+// literal message that backs the precision_timestamp_tz field.
+type PrecisionTimestampTz struct {
+	Precision int32
```
benbellick, line 355: and this

benbellick (approved):
I fixed the conflict with a push but it LGTM!

PR #327, opened 2026-09-08
fix(extensions): return errors for malformed function argument YAML

### LINKED ISSUE #287: Malformed simple-extension function arguments can panic YAML decoding
`FuncParameterList.UnmarshalYAML` uses unchecked type assertions for argument fields. A malformed simple-extension file can therefore panic instead of returning a YAML decoding error.

This reproduces with `substrait-go/v8` v8.1.1:

```go
package main

import (
    "fmt"

    "github.com/goccy/go-yaml"
    "github.com/substrait-io/substrait-go/v8/extensions"
)

func main() {
    var extension extensions.SimpleExtensionFile
    err := yaml.Unmarshal([]byte(`
urn: extension:test:malformed
scalar_functions:
- name: example
  impls:
  - args:
    - name: 7
      value: i64
    return: i64
`), &extension)
    fmt.Println(err)
}
```

The call panics:

```text
panic: interface conversion: interface {} is uint64, not string

github.com/substrait-io/substrait-go/v8/extensions.(*FuncParameterList).UnmarshalYAML(...)
    extensions/simple_extension.go:129
```

The decoder should return an error describing the invalid `name` field instead.

The same issue appears possible for other directly asserted fields, including `description`, `constant`, and `options` values. Replacing the assertions with checked conversions would allow malformed extension YAML to fail validation without crashing the caller.

alexandrefimov:
Malformed function argument fields can panic while loading a simple-extension YAML file. `name: 7` is one example. The regression cases cover 13 panics from unchecked type assertions and two reflection panics caused by null `value` or `type` fields.

Validate argument field types and return errors naming the argument index and field. Parse `value` and `type` strings directly, removing the reflection adapter. Four other cases already returned errors; those errors now carry the same argument context.

Reject arguments with none of `value`, `type`, or `options`, including `{}`, a name-only argument, and a misspelled `valu` key. These previously decoded to a nil parameter and could panic later when building a function's compound name. Valid argument values and defaults keep their existing behavior.

Closes #287

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid YAML function argument fields now return clear, contextual errors instead of causing application crashes.
  - Parsing errors identify the affected argument and field for easier troubleshooting.
  - Malformed nested options and missing argument definitions are handled safely.

- **Tests**
  - Added coverage for invalid argument fields, nested option values, and malformed definitions.
  - Added validation for successfully decoded value, enum, and type arguments, including metadata and parameterized types.


benbellick:
@coderabbit review

PR #339, opened 2026-09-16
feat(expr)!: make UserDefinedLiteral a domain type

kadinrabo:
Part of #280. De-aliases UserDefinedLiteral from the generated protobuf message into a substrait-go type, converting at the proto boundary so consumers still compile. It lives in expr rather than types because its struct value arm is an expr.StructLiteralValue and types cannot import expr.

BREAKING CHANGE: types.UserDefinedLiteral is removed; it is now expr.UserDefinedLiteral, with its value modeled as expr.UserDefinedLiteralValue (UserDefinedValueAny / UserDefinedValueStruct).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed serialization and round-trip handling for pointer-based arbitrary and structured user-defined literal values.
  * Prevented failures when decoding user-defined literals with missing structured values.

* **Refactor**
  * Improved consistency when constructing and converting user-defined literals across expression and literal APIs.
  * Removed reliance on an exposed protobuf-based type alias.

* **Tests**
  * Expanded coverage for literal variants, payload preservation, schema consistency, and missing structured values.


expr/user_defined_literal.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// setUserDefinedVal encodes a domain UserDefinedLiteralValue onto the val oneof of
+// a protobuf UserDefined literal.
+func setUserDefinedVal(ud *proto.Expression_Literal_UserDefined, v UserDefinedLiteralValue) {
+	switch val := v.(type) {
+	case UserDefinedValueAny:
+		ud.Val = &proto.Expression_Literal_UserDefined_Value{Value: val.Value}
+	case UserDefinedValueStruct:
+		ud.Val = &proto.Expression_Literal_UserDefined_Struct{Struct: val.Value.ToProto()}
```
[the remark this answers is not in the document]
benbellick: This was still an issue, I pushed a commit to address it.

expr/user_defined_literal_test.go
```diff
@@ -309,6 +330,38 @@ func TestMixedRepresentationNestedUserDefinedLiteral(t *testing.T) {
 	require.Equal(t, triangle, result)
 }
 
+// TestUserDefinedLiteralNilStructDecodes verifies a UserDefined literal whose struct
+// oneof is selected but nil decodes without dereferencing the absent struct.
+func TestUserDefinedLiteralNilStructDecodes(t *testing.T) {
```
benbellick: I'm going to drop this test. I don't think this is behavior we want to encode. Really what we should be doing is returning an error on `nil` values but we'd have to make API changes for that.

expr/literals.go
```diff
@@ -55,6 +55,9 @@ type (
 
 // Deprecated: use VirtualTableExprFromLiteralProto
 func StructLiteralFromProto(s *proto.Expression_Literal_Struct) StructLiteralValue {
+	if s == nil {
```
benbellick: Also got rid of this, no need to change this behavior in this pr

benbellick (approved):

PR #340, opened 2026-09-16
feat(extensions)!: make AdvancedExtension a domain type

kadinrabo:
Part of #280. De-aliases AdvancedExtension from the generated protobuf message into a substrait-go struct (Optimization / Enhancement), converting at the proto boundary so consumers still compile. Nil-safe getters keep it satisfying the plan.AdvancedExtension interface.

BREAKING CHANGE: extensions.AdvancedExtension is now a substrait-go struct rather than an alias of the protobuf message.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of advanced extensions when reading and writing plans, relations, and expressions.
  * Advanced extension data now converts reliably between serialized and runtime representations.
  * Optimization and enhancement values are preserved accurately, including when values are absent.
  * Advanced extension values now round-trip correctly across supported plan and relation types.

* **Tests**
  * Added coverage for field access, serialization round trips, populated values, and nil-value handling.


extensions/extension_mgr.go
```diff
@@ -17,9 +17,54 @@ import (
...
 )
 
-type AdvancedExtension = extensions.AdvancedExtension
+// AdvancedExtension embeds additional, non-standard information into a
+// serialized Substrait plan. Optimizations may be ignored by a consumer;
+// enhancements alter semantics and cannot be ignored.
+type AdvancedExtension struct {
+	Optimization []*anypb.Any
+	Enhancement  *anypb.Any
+}
```
benbellick, line 36: In terms of nice UX for users of the library, it could be nice to add a "dumb" wrapper for these types like:
```go
// <Light comment explaining what this is>
type Optimization anypb.Any
// <Light comment explaining what this is>
type Enhancement anypb.Any

type AdvancedExtension struct {
	Optimization []*Optimization
	Enhancement  *Enhancement
}
```
We can leave this for a follow up though if you think it adds a lot of complexity though.

extensions/extension_mgr.go
```diff
@@ -17,9 +17,54 @@ import (
...
 	"github.com/substrait-io/substrait-protobuf/go/substraitpb/extensions"
+	"google.golang.org/protobuf/types/known/anypb"
 )
 
-type AdvancedExtension = extensions.AdvancedExtension
+// AdvancedExtension embeds additional, non-standard information into a
+// serialized Substrait plan. Optimizations may be ignored by a consumer;
+// enhancements alter semantics and cannot be ignored.
+type AdvancedExtension struct {
+	Optimization []*anypb.Any
```
benbellick:
```suggestion
	Optimizations []*anypb.Any
```

extensions/extension_mgr.go
```diff
@@ -17,9 +17,54 @@ import (
...
+	Enhancement  *anypb.Any
+}
+
+// GetOptimization returns the optimization payloads, or nil for a nil receiver.
+func (a *AdvancedExtension) GetOptimization() []*anypb.Any {
```
benbellick:
```suggestion
func (a *AdvancedExtension) GetOptimizations() []*anypb.Any {
```

extensions/extension_mgr.go
benbellick: good to emphasize that there are possibly many

extensions/extension_mgr.go
benbellick: If you do decide to leave it as a follow up, let's just make a short issue so we don't forget.

benbellick (approved):

plan/relations.go
```diff
@@ -88,7 +88,7 @@ func (b *baseReadRel) fromProtoReadRel(rel *proto.ReadRel, reg expr.ExtensionReg
 		b.projection = expr.MaskExpressionFromProto(rel.Projection)
 	}
 
-	b.advExtension = rel.AdvancedExtension
+	b.advExtension = extensions.AdvancedExtensionFromProto(rel.AdvancedExtension)
 	return nil
 }
```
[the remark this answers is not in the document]
benbellick: I'm okay on skipping this one.

PR #341, opened 2026-09-16
feat(plan)!: make Hint, Stats, and RuntimeConstraint domain types

kadinrabo:
Part of #280. De-aliases Hint, Stats, and RuntimeConstraint from the generated protobuf messages into substrait-go structs (including the Hint's SavedComputation/LoadedComputation entries and a ComputationType domain enum), converting at the proto boundary so consumers still compile.

BREAKING CHANGE: plan.Hint, plan.Stats, and plan.RuntimeConstraint are now substrait-go structs rather than aliases of the protobuf messages.

## Summary by CodeRabbit

* **New Features**
  * Added support for plan hints, statistics, runtime constraints, and computation details.
  * Added saving and loading of nested computation information.
  * Added readable labels for supported and unknown computation types.
  * Improved conversion between plan information and its serialized form.
  * Added validation to ensure plan metadata remains consistent during serialization and loading.


plan/common.go
```diff
@@ -117,12 +118,221 @@ func walkExpr(e expr.Expression, fn func(expr.Expression)) {
...
-type (
-	Hint              = proto.RelCommon_Hint
-	Stats             = proto.RelCommon_Hint_Stats
-	RuntimeConstraint = proto.RelCommon_Hint_RuntimeConstraint
+// Stats are the statistics related to a Hint (physical properties of records).
+type Stats struct {
+	RowCount          float64
+	RecordSize        float64
+	AdvancedExtension *extensions.AdvancedExtension
+}
```
benbellick, line 126: nit: would be nice to add comments on these fields.

plan/common_test.go
benbellick, line 1: For consistency, I think all of the other domain types we added those descriptor tests to make sure we catch new fields that are missing. Probably good to add that here as well for all the new types?

benbellick (approved):

PR #342, opened 2026-09-16
fix(plan): derive physical join output schemas by join type

### LINKED ISSUE #330: plan: HashJoinRel and MergeJoinRel ignore the join type when deriving their output schema
`JoinRel.directOutputSchema()` (`plan/relations.go:861`) switches on `j.joinType` and applies the rule per type: the opposite side widened to nullable for left and right, every field nullable for outer, one side for semi and anti. `HashJoinRel.directOutputSchema()` (`plan/relations.go:2111`) and `MergeJoinRel.directOutputSchema()` (`plan/relations.go:2228`) never read the join type at all. Each is a single return:

```go
func (hr *HashJoinRel) directOutputSchema() types.RecordType {
	return hr.left.RecordType().Concat(hr.right.RecordType())
}
```

`physical_relations.md` gives HashJoin and MergeJoin the Direct Output Order "Same as the [Join](logical_relations.md#join-operation) operator", so concatenating is right for inner and wrong wherever the join type changes anything.

## Observed

Left input `(i64, i64?)`, right input `(i64?, i64)`, join type `LEFT`, so the right side should come back nullable throughout.

| relation | derived |
| --- | --- |
| `JoinRel` | `[i64 i64? i64? i64?]`, which is correct |
| `HashJoinRel` | `[i64 i64? i64? i64]` |
| `MergeJoinRel` | `[i64 i64? i64? i64]` |

The same three with join type `INNER` all return `[i64 i64? i64? i64]`, and that is the right answer there, which is why inner alone does not show this.

Reproduced on main at `cb2d6e648bc0ee1549e354dbfe705d9f8a1ce2b1`.

<details>
<summary>To reproduce</summary>

There is no builder for the physical joins, so this goes through the protobuf. Save as `main.go` in an empty directory beside a `go.mod` requiring `github.com/substrait-io/substrait-go/v9` at that commit, and run `go run .`.

```go
package main

import (
	"fmt"

	"github.com/substrait-io/substrait-go/v9/extensions"
	"github.com/substrait-io/substrait-go/v9/plan"
	spb "github.com/substrait-io/substrait-protobuf/go/substraitpb"
)

func i64(n spb.Type_Nullability) *spb.Type {
	return &spb.Type{Kind: &spb.Type_I64_{I64: &spb.Type_I64{Nullability: n}}}
}

func read(table string, types ...*spb.Type) *spb.Rel {
	names := make([]string, len(types))
	for i := range types {
		names[i] = fmt.Sprintf("c%d", i)
	}
	return &spb.Rel{RelType: &spb.Rel_Read{Read: &spb.ReadRel{
		Common:     &spb.RelCommon{EmitKind: &spb.RelCommon_Direct_{Direct: &spb.RelCommon_Direct{}}},
		BaseSchema: &spb.NamedStruct{Names: names, Struct: &spb.Type_Struct{Types: types, Nullability: spb.Type_NULLABILITY_REQUIRED}},
		ReadType:   &spb.ReadRel_NamedTable_{NamedTable: &spb.ReadRel_NamedTable{Names: []string{table}}},
	}}}
}

func main() {
	R, N := spb.Type_NULLABILITY_REQUIRED, spb.Type_NULLABILITY_NULLABLE
	left := read("l", i64(R), i64(N))
	right := read("r", i64(N), i64(R))
	common := &spb.RelCommon{EmitKind: &spb.RelCommon_Direct_{Direct: &spb.RelCommon_Direct{}}}
	fieldRef := func(i int32) *spb.Expression_FieldReference {
		return &spb.Expression_FieldReference{
			ReferenceType: &spb.Expression_FieldReference_DirectReference{DirectReference: &spb.Expression_ReferenceSegment{
				ReferenceType: &spb.Expression_ReferenceSegment_StructField_{StructField: &spb.Expression_ReferenceSegment_StructField{Field: i}}}},
			RootType: &spb.Expression_FieldReference_RootReference_{RootReference: &spb.Expression_FieldReference_RootReference{}},
		}
	}
	key := &spb.ComparisonJoinKey{Left: fieldRef(0), Right: fieldRef(0),
		Comparison: &spb.ComparisonJoinKey_ComparisonType{InnerType: &spb.ComparisonJoinKey_ComparisonType_Simple{
			Simple: spb.ComparisonJoinKey_SIMPLE_COMPARISON_TYPE_EQ}}}
	alwaysTrue := &spb.Expression{RexType: &spb.Expression_Literal_{Literal: &spb.Expression_Literal{
		LiteralType: &spb.Expression_Literal_Boolean{Boolean: true}}}}

	cases := []struct {
		name string
		rel  *spb.Rel
	}{
		{"JoinRel      ", &spb.Rel{RelType: &spb.Rel_Join{Join: &spb.JoinRel{
			Common: common, Left: left, Right: right, Expression: alwaysTrue, Type: spb.JoinRel_JOIN_TYPE_LEFT}}}},
		{"HashJoinRel  ", &spb.Rel{RelType: &spb.Rel_HashJoin{HashJoin: &spb.HashJoinRel{
			Common: common, Left: left, Right: right, Keys: []*spb.ComparisonJoinKey{key}, Type: spb.HashJoinRel_JOIN_TYPE_LEFT}}}},
		{"MergeJoinRel ", &spb.Rel{RelType: &spb.Rel_MergeJoin{MergeJoin: &spb.MergeJoinRel{
			Common: common, Left: left, Right: right, Keys: []*spb.ComparisonJoinKey{key}, Type: spb.MergeJoinRel_JOIN_TYPE_LEFT}}}},
	}
	for _, c := range cases {
		p := &spb.Plan{
			Version:   &spb.Version{MinorNumber: 102},
			Relations: []*spb.PlanRel{{RelType: &spb.PlanRel_Root{Root: &spb.RelRoot{Input: c.rel, Names: []string{"a", "b", "c", "d"}}}}},
		}
		pl, err := plan.FromProto(p, extensions.GetDefaultCollectionWithNoError())
		if err != nil {
			fmt.Printf("%s %v\n", c.name, err)
			continue
		}
		out := []string{}
		for _, ft := range pl.GetRoots()[0].RecordType().Struct.Types {
			out = append(out, ft.String())
		}
		fmt.Printf("%s %v\n", c.name, out)
	}
}
```

It prints:

```
JoinRel       [i64 i64? i64? i64?]
HashJoinRel   [i64 i64? i64? i64]
MergeJoinRel  [i64 i64? i64? i64]
```

</details>

alexandrefimov:
Derive HashJoinRel and MergeJoinRel output schemas by join type, including nullability and field selection, before applying emit.

Bind post-join filters to the direct output before emit, following the [clarified semantics](https://redirect.github.com/substrait-io/substrait/pull/1044).

Tests cover all 12 join types, emit and post-filters. Full tests, build and lint pass. Corpus checks improve 11 schema answers with no other changes.

Fixes #330.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved hash and merge join output schemas across semi, anti, outer, single, and mark joins.
  - Preserved correct nullability for unmatched results, nested fields, and mark indicators.
  - Fixed post-join filters to resolve column references against the join’s direct output and consistently reject unavailable fields.

- **Tests**
  - Added coverage for join output types, nullability, direct outputs, and post-join filtering across join variants.
  - Added safeguards ensuring combined record results do not alter source records.


plan/relations.go
```diff
@@ -1885,8 +1885,52 @@ const (
 	HashMergeRightSemi
 	HashMergeLeftAnti
 	HashMergeRightAnti
+	HashMergeLeftSingle
+	HashMergeRightSingle
+	HashMergeLeftMark
+	HashMergeRightMark
 )
 
+// hashMergeJoinOutputSchema applies the physical join type before emit mapping.
```
benbellick: Thanks for your work here! A few things I would suggest. One, I think this code would be a bit more readable if we just returned the appropriate schema on the spot in every case, rather than sometimes using these `nullableLeft` and `nullableRight` assignments:
```go
switch joinType {
case <>:
	return <>
case <>:
	return <>
// etc.
} 
```
Doing this nicely will probably require some nicer helper methods. Maybe things like a helper function called `nullableRecordType` which we can use to make the fields nullable. 


One other fix that I would suggest here. I think that our implementation of `Concat` is not very good 😢 It would be better not to mutate the underlying slice IMO and return something new. 

This is exactly what this PR would partially address, but I would actually rather we make `Concat` do what we really want, which is something like:
```go
   func (r RecordType) Concat(other RecordType) RecordType {
      return RecordType{
         types: slices.Concat(r.Types(), other.Types()),
      }
   }
```
rather than the current behavior of mutating the underlying slice
https://github.com/substrait-io/substrait-go/blob/a09cc3c10be8a9136bc408ba6b8e58399d4aa9fa/types/types.go#L2078-L2080
what do you think?

plan/physical_join_schema_test.go
```diff
@@ -0,0 +1,188 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package plan
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
benbellick, line 8: This testing is all quite complicated and a bit difficult to follow. I would recommend thinking about how to simplify these tests a bit to optimize for readability. They should really focus on the core contribution of the PR here, which is fixing the schema derivation for joins. So, for example, I don't think we need to add tests that test both emit kinds.

plan/relations.go
```diff
@@ -1885,8 +1885,52 @@ const (
...
+// HashMergeJoinType and JoinType use different values for semi, anti and single joins.
+func hashMergeJoinOutputSchema(left, right types.RecordType, joinType HashMergeJoinType) types.RecordType {
```
benbellick: I also think this function may be a bit more natural on the `HashMergeJoinType`, i.e.
```go
func (t HashMergeJoinType) outputSchema(left, right ...) {
    ...
}
```

benbellick (changes requested):
Thanks for the work! Just a few comments

plan/relations.go
alexandrefimov: Both done. `Concat` was overwriting an earlier result from the same receiver, not only mutating.

plan/physical_join_schema_test.go
alexandrefimov: Simplified. The `emit=false/true` loops are gone; emit stays only where it is the point.

plan/relations.go
alexandrefimov: Done.

plan/relations.go
```diff
@@ -1885,8 +1885,43 @@ const (
...
 	HashMergeRightAnti
+	HashMergeLeftSingle
+	HashMergeRightSingle
+	HashMergeLeftMark
+	HashMergeRightMark
 )
 
+// outputSchema applies the physical join type before emit mapping.
+// HashMergeJoinType and JoinType use different values for semi, anti and single joins.
+func (t HashMergeJoinType) outputSchema(left, right types.RecordType) types.RecordType {
```
benbellick, line 1896: Nice! I think this is much better :)

plan/relations.go
```diff
@@ -1885,8 +1885,43 @@ const (
...
+		return nullableRecordType(left).Concat(nullableRecordType(right))
+	case HashMergeLeft, HashMergeLeftSingle:
+		return left.Concat(nullableRecordType(right))
+	case HashMergeRight, HashMergeRightSingle:
+		return nullableRecordType(left).Concat(right)
+	case HashMergeLeftMark:
+		return left.Concat(*types.NewRecordTypeFromTypes([]types.Type{&types.BooleanType{Nullability: types.NullabilityNullable}}))
+	case HashMergeRightMark:
+		return right.Concat(*types.NewRecordTypeFromTypes([]types.Type{&types.BooleanType{Nullability: types.NullabilityNullable}}))
+	default:
```
kadinrabo: This is a good cleanup, deriving the schema per join type reads much better than the old flat concat, and the 12-case table covers the space well.

On the [`default` branch](https://github.com/substrait-io/substrait-go/blob/3c059ba15242198a472096404ef4b334b2462ac5/plan/relations.go#L1912): `joinType` is cast straight from the proto with no validation, so `HashMergeUnspecified` (the zero value) falls here and silently gets an inner schema, and any join type added to the enum later would too. The logical side uses an explicit [`case JoinTypeInner`](https://github.com/substrait-io/substrait-go/blob/3c059ba15242198a472096404ef4b334b2462ac5/plan/relations.go#L864) rather than a catch-all. Worth matching that here so an unset or unknown type surfaces instead of passing as inner?

kadinrabo, line 1914: Would it be worth handling `HashMergeInner` explicitly like the [logical side does](https://github.com/substrait-io/substrait-go/blob/3c059ba15242198a472096404ef4b334b2462ac5/plan/relations.go#L864), so `default` is left for the unexpected cases?

alexandrefimov: Good point — done in 467b60c.

plan/relations.go
```diff
@@ -1885,8 +1885,45 @@ const (
...
+		return left.Concat(right)
+	}
+}
+
+func nullableRecordType(record types.RecordType) types.RecordType {
+	fields := make([]types.Type, record.FieldCount())
+	for i, field := range record.Types() {
+		fields[i] = field.WithNullability(types.NullabilityNullable)
```
benbellick, line 1922: I noticed an existing bug from this change (not due to you) which is that `WithNullability` can mutate one of the types when it isn't supposed to. I have a fix for that here: https://github.com/substrait-io/substrait-go/pull/349. @alexandrefimov would you mind reviewing that PR please? 

Or you can open up the PR yourself and I would be happy to review and get it in. I think we should do it before this one. Thanks :)

plan/physical_join_schema_test.go
```diff
@@ -0,0 +1,147 @@
...
+	mark := &types.BooleanType{Nullability: types.NullabilityNullable}
+	left := &NamedTableReadRel{names: []string{"left"}, baseReadRel: baseReadRel{
+		baseSchema: types.NamedStruct{Names: []string{"l0", "l1"}, Struct: types.StructType{
+			Nullability: types.NullabilityRequired, Types: []types.Type{i64, strNull},
+		}},
+	}}
+	right := &NamedTableReadRel{names: []string{"right"}, baseReadRel: baseReadRel{
+		baseSchema: types.NamedStruct{Names: []string{"r0", "r1"}, Struct: types.StructType{
+			Nullability: types.NullabilityRequired, Types: []types.Type{i64Null, str},
+		}},
```
benbellick: Can we use `NamedScan` builders for these?

benbellick (commented):
Sorry didn't have time to finish the review but wanted to dump my comments so far. Thanks!

plan/physical_join_schema_test.go
alexandrefimov: Done in b99fa4b.

PR #344, opened 2026-09-16
fix(plan): handle absent RelCommon without panicking

### LINKED ISSUE #328: plan: RelFromProto panics on a join, cross, sort, aggregate or set relation without RelCommon
`plan.RelFromProto` panics with a nil pointer dereference when a `JoinRel`, `CrossRel`, `SortRel`, `AggregateRel` or `SetRel` has no `common`. With an empty `RelCommon` the same relations are accepted, and [`algebra.proto`](https://github.com/substrait-io/substrait/blob/v0.103.0/proto/substrait/algebra.proto#L17-L21) says in the comment on `RelCommon.emit_kind` that an absent `RelCommon` is treated as Direct.

DataFusion writes sorts, aggregates and set operations without `common`. With the `datafusion` 54.0.0 Python package, the plans it produces for `SELECT a FROM t ORDER BY a`, `SELECT a FROM t UNION ALL SELECT b FROM u` and `SELECT a FROM t GROUP BY a` each panic in `plan.FromProto`.

The panic is in [`fromProtoCommon`](https://github.com/substrait-io/substrait-go/blob/cb2d6e648bc0ee1549e354dbfe705d9f8a1ce2b1/plan/common.go#L133-L135), which reads `c.Hint` and `c.AdvancedExtension` directly. Fourteen branches of `RelFromProto` call it. The filter, fetch, project and write branches check `common` for nil first, and the other ten do not: the five relations above, `HashJoinRel`, `MergeJoinRel` and the three extension relations. When the two fields are read through `c.GetHint()` and `c.GetAdvancedExtension()` instead, the test below passes, the rest of `./plan` still passes, and substrait-go derives `NSTRUCT<a: i32?>` for each of the three DataFusion plans.

Reproduced on main at `cb2d6e648bc0`. With this test in `plan/`, the `absent` subtest panics and the `empty` one passes. A join, cross, sort or aggregate relation in place of the set relation behaves the same way.

```go
package plan_test

import (
	"testing"

	"github.com/stretchr/testify/assert"
	"github.com/substrait-io/substrait-go/v9/expr"
	"github.com/substrait-io/substrait-go/v9/extensions"
	"github.com/substrait-io/substrait-go/v9/plan"
	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
)

func TestSetRelWithoutCommon(t *testing.T) {
	read := func() *proto.Rel {
		return &proto.Rel{RelType: &proto.Rel_Read{Read: &proto.ReadRel{
			BaseSchema: &proto.NamedStruct{Names: []string{"x"}, Struct: &proto.Type_Struct{
				Nullability: proto.Type_NULLABILITY_REQUIRED,
				Types: []*proto.Type{{Kind: &proto.Type_I32_{I32: &proto.Type_I32{
					Nullability: proto.Type_NULLABILITY_REQUIRED}}}}}},
			ReadType: &proto.ReadRel_NamedTable_{NamedTable: &proto.ReadRel_NamedTable{Names: []string{"t"}}},
		}}}
	}
	reg := expr.NewEmptyExtensionRegistry(extensions.GetDefaultCollectionWithNoError())

	for name, c := range map[string]*proto.RelCommon{"empty": {}, "absent": nil} {
		rel := &proto.Rel{RelType: &proto.Rel_Set{Set: &proto.SetRel{
			Common: c, Inputs: []*proto.Rel{read(), read()}, Op: proto.SetRel_SET_OP_UNION_ALL}}}
		t.Run(name, func(t *testing.T) {
			assert.NotPanics(t, func() {
				_, err := plan.RelFromProto(rel, reg)
				assert.NoError(t, err)
			})
		})
	}
}
```

I can send the getter change with this test as a PR, if that would help.

alexandrefimov:
Use nil-safe getters when decoding RelCommon so an absent common is treated as Direct instead of causing a panic.

A focused SetRel test checks the output schema when common is absent. Full tests pass.

Fixes #328.

## Summary by CodeRabbit

- **Tests**
  - Updated coverage for deserializing set relations when common metadata is absent.
  - Confirmed that deserialization completes successfully and preserves the input record type.
  - Simplified the test scenario to focus on missing common metadata without covering empty metadata or panic behavior.


plan/rel_common_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/substrait-io/substrait-go/v9/expr"
+	"github.com/substrait-io/substrait-go/v9/types"
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+	extensionspb "github.com/substrait-io/substrait-protobuf/go/substraitpb/extensions"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+func TestRelCommonAbsentIsDirect(t *testing.T) {
```
benbellick: This is a lot of complex testing for a simple code change. Is it possible to have a simpler and more readable test to capture the fixed behavior? Thanks!

alexandrefimov: Done

plan/rel_common_test.go
benbellick, line 1: I think the interpretation of a missing `RelCommon` was technically added after the version of substrait this repo is currently pinned to.
https://github.com/substrait-io/substrait/blob/a902858cd0bc28c5017765e71e6e3c32b01076d6/proto/substrait/algebra.proto#L19-L21

However, because it is a compatible change and expresses what we want to move to, I think its reasonable to accept it now.

plan/rel_common_test.go
```diff
@@ -0,0 +1,35 @@
...
+				Op:     proto.SetRel_SET_OP_UNION_ALL,
+			}}}
+			var rel Rel
+			var err error
+			require.NotPanics(t, func() { rel, err = RelFromProto(wire, joinTestRegistry()) })
+			require.NoError(t, err)
+			assert.Equal(t, input.RecordType(), rel.RecordType())
+		})
+	}
+}
```
benbellick:
```suggestion
   func TestRelFromProtoWithoutCommon(t *testing.T) {
      input := createJoinInput("input")
      wire := &proto.Rel{RelType: &proto.Rel_Set{Set: &proto.SetRel{
         Inputs: []*proto.Rel{input.ToProto(), input.ToProto()},
         Op:     proto.SetRel_SET_OP_UNION_ALL,
      }}}

      rel, err := RelFromProto(wire, joinTestRegistry())
      require.NoError(t, err)
      assert.Equal(t, input.RecordType(), rel.RecordType())
   }
```
This is a massive improvement but I think we can simplify even further. LMK if you agree. Thanks!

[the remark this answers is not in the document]
benbellick: Disagree here, they all fundamentally exercise the same path 🤷

benbellick (approved):

plan/rel_common_test.go
alexandrefimov: Done

plan/rel_common_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/stretchr/testify/require"
+	proto "github.com/substrait-io/substrait-protobuf/go/substraitpb"
+)
+
+func TestRelFromProtoWithoutCommon(t *testing.T) {
+	input := createJoinInput("input")
+	wire := &proto.Rel{RelType: &proto.Rel_Set{Set: &proto.SetRel{
+		Inputs: []*proto.Rel{input.ToProto(), input.ToProto()},
+		Op:     proto.SetRel_SET_OP_UNION_ALL,
+	}}}
```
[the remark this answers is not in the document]
benbellick: eh i disagree 🙂

PR #348, opened 2026-09-22
fix(extensions): leave arguments unknown when a compound name does not parse

### LINKED ISSUE #345: parseFuncName panics on a compound name the library itself emits
`extensions.NewScalarFuncVariant` panics when the argument types in a compound name do not parse. That name comes from a plan: this constructor is the path taken for a function the registry does not hold, so a plan crashes the library instead of being rejected.

```go
id := extensions.FunctionID{
	URN:  "extension:io.substrait:functions_arithmetic_decimal",
	Name: "no_such_function:dec_dec",
}
extensions.NewScalarFuncVariant(id)   // panic: error parsing input 'dec'
```

`dec` is not an arbitrary string. `(*DecimalType).ShortString()` returns `"dec"`, which is what this library writes into a compound name; `parser.ParseType("dec")` rejects it, and `ParseType("decimal")` rejects it too, wanting parameters. A decimal argument does not survive the round trip through a compound name whenever the lookup misses. So this shows on any plan naming a decimal function the registry does not hold, and not on `add:dec_dec`, which resolves.

[`parseFuncName`](https://github.com/substrait-io/substrait-go/blob/v9.0.0-alpha.0/extensions/variants.go#L273-L289) calls `parser.ParseType` on each part of the suffix with `panic(err)` on failure. `resolveVariant` already returns `ErrNotFound` for a compound name it cannot find, so only the on-the-fly constructors reach it. Separate from #287 and #327, which are about `UnmarshalYAML`.

Happy to send the fix. Returning an error from `parseFuncName` changes its signature and the six constructors that call it, so say which shape you would prefer.

alexandrefimov:
`parseFuncName` no longer panics when the suffix of a compound name does not parse, such as the `dec` this library writes for a decimal. The variant gets no arguments, the same as a name with no suffix; the return type on this path comes from the plan's `output_type`.

I kept the constructor signatures. If you would rather have them return the parse error, I can change that (a breaking change to six public functions).

Fixes #345.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of function variants with unsupported compound-name suffixes.
  * Such variants now return the base function name without parsed arguments instead of causing an error.
  * Valid compound names continue to parse their arguments correctly.

* **Tests**
  * Added coverage for scalar, aggregate, and window function variants with both unsupported and valid compound names.


extensions/variants.go
```diff
@@ -279,7 +284,7 @@ func parseFuncName(compoundName string) (name string, args FuncParameterList) {
 	for _, argStr := range splitArgs {
 		parsed, err := parser.ParseType(argStr)
 		if err != nil {
-			panic(err)
```
benbellick, line 282: Yes, I agree this is a problem, but I don’t think the solution is to silently swallow the error. However (...and maybe this is an overly aggressive suggestion😬) I think the better approach may be to remove `parseFuncName` and its callers altogether.

Its callers are the six public constructors for on-the-fly scalar, aggregate, and window variants. Three of those are also used when we read a plan containing a function that isn’t defined in the extension collection.

My reasons are:

1. We’re using the wrong parser in the first place. The components of a function signature are short type names, not general type expressions. `dec` and `list` are valid short names, but `parser.ParseType` can’t parse them as complete types.
2. What value is there in allowing people to construct these variants in the first place? I think functions should be defined through extension YAML. Otherwise, we’re treating a partially specified function like a fully defined variant. We may have a concrete output type from the plan or caller, but we don’t have the function’s return-type expression, nullability behavior, or full argument definitions.
3. The one case where I can see a reason to construct something on the fly is a function in a plan we consume whose definition we haven’t loaded. Personally, I think we should reject those and only accept functions defined in the extension collection. But if we *must* support them, I think the API for doing so should at least be private and should explicitly mark the function as partially defined, rather than treating it like any other variant.

This would be a breaking API change. What do you think?

alexandrefimov: Reworked in 4f4bfc5: dropping the arguments changed the variant's identity, so a plan naming `no_such:dec_dec` came back with a second declaration, `no_such:`. The variant now keeps the name it was built from, with a round-trip test.

The constructors run only when the URN resolves but the function is missing from the collection; an unknown URN is already rejected. Removing them would end reading a plan that uses a newer function of an extension we do have, such as `brand_new_fn:i64` under `functions_arithmetic`, typed today from the plan's own `output_type`. I can do the removal instead if you prefer.

